PR #98, opened 2019-12-05
Deploy on AWS

alexduf:
## Why are you doing this?

To host the app on AWS

## Changes

- Package the app for riff-raff
- Build it and package it via team city
- Deploy the app via riff-raff (except the auto scaling group)
- force the port to 9000
- add a simple healthcheck for the load balancer to pick

### DIFF
--- a/artifact.json
+++ b/artifact.json
@@ -1,0 +1,16 @@
+{
+  "projectName": "Mobile::mobile-apps-rendering",
+  "vcsURL": "https://github.com/guardian/apps-rendering",
+  "actions": [
+    {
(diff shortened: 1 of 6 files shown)


alexduf:
We now have a fully set of machines behind a load balancer. The load balancer is only visible from MAPI's VPC, and I tested the connectivity from MAPI's instances.

```
curl https://<domain-name-of-the-private-load-balancer>/healthcheck
Ok
```

Deployments are also working fine, so I think that covers all the points raised.
Ready for a new review (probably next week!)

PR #99, opened 2019-12-05
Create snyk action

webb04:
## Why are you doing this?
Add GitHub action using Snyk to detect vulnerabilities before merging into master.

It will fail for any vulnerability detected but we do have the option to only fail on severe vulnerabilities. I would suggest we see how we go with this first.

I've deleted the GitHub integration in Snyk.io so the last 2 checks shouldn't run in future pull requests.

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -1,0 +1,22 @@
+name: Snyk
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  validate:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [12.x]
+
+    steps:
+    - uses: actions/checkout@master
+    - name: Run Snyk to check for vulnerabilities
+      uses: snyk/actions/node@master
+      env:
+        SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -6,7 +6,7 @@ on:
       - master
 
 jobs:
-  validate:
+  snyk:
 
     runs-on: ubuntu-latest
 


.github/workflows/snyk.yml
```diff
@@ -0,0 +1,22 @@
+name: Snyk
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  validate:
```
JamieB-gu: Since we already have a validate job, could this be named something different?

.github/workflows/snyk.yml
```diff
@@ -0,0 +1,22 @@
...
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [12.x]
+
+    steps:
+    - uses: actions/checkout@master
```
JamieB-gu: Is this using the [v2 beta](https://github.com/actions/checkout) of the checkout action? Should we prefer v1 until v2 is out of beta?

JamieB-gu:
> more details about how that all works?

@webb04 am I correct in thinking this is derived from Snyk's recommendation for their GH Action [here](https://github.com/snyk/actions)?

webb04:
This will run every 24 hours https://app.snyk.io/org/guardian-mobile/projects

webb04:
> I don't see anything contention in this PR, do you mind going in a little bit more details about how that all works?

Yep, using the template available here https://github.com/snyk/actions and pulling in their GitHub action. Added the Snyk token as an encrypted secret in GitHub. 

It should fail builds if any vulnerability is detected.

JamieB-gu:
> This will run every 24 hours

Is the plan to add `apps-rendering` to that as part of this ticket?

jacobwinch:
> 1. As far as I understand it, this will run whenever a PR is opened against master? Could this cause confusion? For example, a vulnerability is discovered in a dependency that exists in `master`. A PR for `branch-1` that does not add any dependencies (or adds an unrelated dependency) is opened, and Snyk reports this vulnerability against this PR. However, in this case, am I correct in thinking that vulnerability has nothing to do with `branch-1`?

This is a really good point. We could add a steps to checkout `master`, run snyk against that and _then_ checkout `branch-1` and re-run the checks. This would give us more clarity, but could also slow things down a bit...

webb04:
Actually a GitHub action might be a little unnecessary as Snyk is already running on this branch https://github.com/guardian/apps-rendering/pull/100

JamieB-gu:
> Snyk is already running on this branch #100

How is that configured?

webb04:
> > Snyk is already running on this branch #100
> 
> How is that configured?

I think just enabling it in the snyk.io UI

webb04:
Merging this now, we can always tweak it after and/or add the daily check.

PR #100, opened 2019-12-06
Opinion articles

webb04:
## Why are you doing this?
Create opinion components to handle specific opinion styles.

## Changes

- Moved `articleBody`, `articleSeries` and `articleStandfirst` into shared
- Created opinion components
- New cutout component
- Reusable headline styles

## Screenshots

| Before | After |
| --- | --- |
|  |  |


src/components/opinion/opinionArticle.tsx
```diff
@@ -0,0 +1,112 @@
+import React from 'react';
+
+import HeaderImage from '../shared/headerImage';
+import ArticleSeries from '../news/articleSeries';
+import OpinionHeadline from '../opinion/opinionHeadline';
+import ArticleStandfirst from '../news/articleStandfirst';
+import OpinionByline from './opinionByline';
+import ArticleBody from '../shared/articleBody';
```
JamieB-gu: Worth using absolute paths for these? `components/shared/...`, `components/news/...` etc.

src/components/opinion/opinionArticle.tsx
```diff
@@ -0,0 +1,112 @@
+import React from 'react';
+
+import HeaderImage from '../shared/headerImage';
+import ArticleSeries from '../news/articleSeries';
```
JamieB-gu: What's the difference between `news` and `shared`?

src/components/opinion/opinionArticle.tsx
```diff
@@ -0,0 +1,112 @@
...
+    background: ${palette.neutral.darkMode};
+`;
+
+const BorderStyles = css`
+    background: ${palette.opinion.faded};
+
+    ${from.wide} {
+        width: ${breakpoints.wide}px;
+        margin: 0 auto;
+    }
```
JamieB-gu: Nice!

src/components/opinion/opinionCutout.tsx
```diff
@@ -0,0 +1,46 @@
...
+}: OpinionCutoutProps): JSX.Element | null => {
+    const [contributor] = contributors;
+
+    if (isSingleContributor(contributors) && contributor.bylineLargeImageUrl) {
+        const imgSrc = transformUrl(imageSalt, contributor.bylineLargeImageUrl, imageWidth*3);
+        return (
+            <div css={[className, OpinionCutoutStyles]}>
+                <img css={ImageStyles} src={imgSrc} alt={contributor.webTitle}/>
+            </div>
+        );
```
JamieB-gu: Is there any way we can use some of the functions in `image` or `asset` to make this a `srcset`?

src/components/opinion/opinionHeadline.tsx
```diff
@@ -0,0 +1,52 @@
...
+    pillarStyles: PillarStyles;
+}
+
+const OpinionHeadline = ({
+    byline,
+    headline
+}: OpinionHeadlineProps): JSX.Element =>
+    <div css={[HeadlineStyles, HeadlineDarkStyles]}>
+        <h1>{headline}</h1>
+        <address dangerouslySetInnerHTML={{__html: byline}}></address>
```
JamieB-gu: Is there a way we can do this without introducing `dangerouslySetInnerHTML`?

src/components/opinion/opinionArticle.tsx
webb04: I can move some more `news` components to be `shared` actually. 
I was thinking `localhost:8080` is news articles and `http://localhost:8080/commentisfree/2019/dec/09/wins-tories-future-boris-johnson-victory-conservatism` to be Opinion.


So the components used in `news` and `opinion` articles should live in `shared`?

src/components/opinion/opinionCutout.tsx
webb04: hmm I don't think we get the Asset data type. Also I don't think the image changes size on different screen widths.

src/components/opinion/opinionArticle.tsx
JamieB-gu: Yeah I think if you're going to explicitly specify a `shared` directory then probably everything that's shared should live in there.

Alternatively you could flatten the directory structure and:

a) Have everything shared live directly under `components`, with subdirectories for stuff that isn't shared:

```sh
components/
  headerImage.tsx
  articleSeries.tsx
  opinion/
    opinionArticle.tsx
  ...
```
b) Have everything live directly under `components`:

```sh
components/
  headerImage.tsx
  articleSeries.tsx
  opinionArticle.tsx
```

src/components/opinion/opinionCutout.tsx
JamieB-gu: This is the author avatar? Yeah I guess that would stay the same "css size", but I was thinking more about [resolution switching](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#Resolution_switching_Same_size_different_resolutions).

src/components/opinion/opinionArticle.tsx
webb04: I like the `a` structure, I'm worried that we would have too many shared components at that top level though... which is also a good thing I guess

JamieB-gu: Do you foresee problems with having many components at the top level?

src/components/opinion/opinionHeadline.tsx
webb04: We would need something like `renderBlocks` but for individual html fields. We've still got a few `dangerouslySetInnerHTML`s in the codebase so maybe that's good to do in another ticket?

src/components/opinion/opinionArticle.tsx
webb04: Might be harder for new devs or us to find components? Will try moving more into the shared folder now but can refactor later if needed.

JamieB-gu: Yeah sounds good.

On the point about finding components I think you could make the same argument about subdirectories. A path to find a component that seems obvious to one dev may not be obvious to another.

webb04: Yeah true, we might need to think about where the tests live too because they seem to be bloating folders too at the moment.

src/components/news/article.tsx
```diff
@@ -1,11 +1,11 @@
 import React from 'react';
 
 import HeaderImage from '../shared/headerImage';
-import ArticleSeries from './articleSeries';
+import ArticleSeries from '../shared/articleSeries';
```
JamieB-gu:
```suggestion
import ArticleSeries from 'components/shared/articleSeries';
```

src/components/news/articleHeadline.tsx
```diff
@@ -1,16 +1,14 @@
 import React from 'react';
-import ArticleRating from './articleRating';
-import { basePx, sidePadding, headlineFont, darkModeCss } from 'styles';
+import ArticleRating from '../shared/articleRating';
```
JamieB-gu:
```suggestion
import ArticleRating from 'components/shared/articleRating';
```

PR #102, opened 2019-12-06
replace articleTemplate.html with react component

webb04:
## Changes

- Replace `articleTemplate.html` with React component
- No need to read in file on every request
- Contains logic to include twitter script or not depending if there's a tweet on the page
- No more html in the project, 100% TypeScript/JavaScript

### DIFF
--- a/src/components/shared/articleContainer.tsx
+++ b/src/components/shared/articleContainer.tsx
@@ -1,102 +1,109 @@
-<!DOCTYPE html>
-<html lang="en-US">
-<head>
-    <title>Article</title>
-    <meta charset="UTF-8" />
-    <meta id="twitter-theme" name="twitter:widgets:theme" content="light" />
-    <meta name="viewport" content="initial-scale=1, maximum-scale=1" />
-    <!-- This to be a shared common css file with fonts. Fonts should probably still be bundled with apps -->
-    <style>
-      @font-face {
+import React from 'react';
+import { css } from '@emotion/core';
+
+const ArticleContainerStyles = css`
+    @font-face {
         font-family: "Guardian Egyptian Web";
         src: url('/public/fonts/egyptiantext-regular.otf');
-      }
+    }
 
-      @font-face {
+    @font-face {
         font-family: "Guardian Egyptian Web";
         font-weight: 500;
         src: url("/public/fonts/egyptiantext-medium.otf");
-      }
+    }
 
-      @font-face {
+    @font-face {
         font-family: "Guardian Egyptian Web";
         font-weight: bold;
         src: url("/public/fonts/egyptiantext-medium.otf");
-      }
+    }
 
-      @font-face {
+    @font-face {
         font-family: "Guardian Text Sans Web";
         src: url('/public/fonts/GuardianSansWeb-Regular.ttf');
-      }
+    }
 
-      @font-face {
+    @font-face {
         font-family: "Guardian Headline";
         font-weight: 200;
         src: url("/public/fonts/GHGuardianHeadline-Light.ttf");
-      }
+    }
 
-      @font-face {
+    @font-face {
         font-family: "Guardian Headline";
         font-weight: 400;
         src: url("/public/fonts/GHGuardianHeadline-Regular.ttf");
-      }
+    }
 
-      @font-face {
+    @font-face {
         font-family: "Guardian Headline";
         font-weight: 500;
         src: url("/public/fonts/GHGuardianHeadline-Medium.ttf");
-      }
+    }
 
-      @font-face {
+    @font-face {
         font-family: "Guardian Headline";
         font-weight: 700;
         src: url("/public/fonts/GHGuardianHeadline-Bold.ttf");
-      }
+    }
 
-      @font-face {
+    @font-face {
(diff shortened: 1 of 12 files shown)


src/components/shared/articleContainer.tsx
```diff
@@ -1,102 +1,109 @@
...
-  <div id="root"></div>
-  <script src="/assets/client.js"></script>
-  <script src="https://platform.twitter.com/widgets.js"></script>
-</body>
-</html>
\ No newline at end of file
+    }
+`;
+
+function ArticleContainer(includesTweets: boolean): JSX.Element {
```
alexduf: Should the logic that decides whether to include tweets or not be part of the component rather than outside?

alexduf: (I came to that conclusion after reading this excellent blog post: https://www.theguardian.com/info/2019/dec/08/migrating-the-guardian-website-to-react )

src/server/server.ts
```diff
@@ -97,14 +91,15 @@ app.get('/*', async (req, res) => {
...
     const articleId = req.params[0] || defaultId;
 
-    const template = await readTemplate();
     const key = await getConfigValue<string>("capi.key");
     const imageSalt = await getConfigValue<string>('apis.img.salt');
     const resp = await fetch(capiEndpoint(articleId, key), {});
     const capi = await resp.text();
+    const parsedCapi = parseCapi(capi);
+    const articleContainer = ArticleContainer(includesTweets(parsedCapi))
+    const template = renderToString(articleContainer)
```
alexduf: So if I understand this correctly the template is rendered to a string, then we inject more components, should the article container be allowed to have children instead?

webb04: I can refactor to pass in child components, that why we would only use `renderToString` once and not twice.

webb04: Although I'll probably wait until this is merged in before refactoring https://github.com/guardian/apps-rendering/pull/101 @JamieB-gu

src/components/shared/articleContainer.tsx
webb04: Good point, it's only used in that one place so makes sense to be moved inside

src/components/shared/articleContainer.tsx
```diff
@@ -0,0 +1,121 @@
...
+    .element-audio iframe {
+        width: 100%;
+    }
+`;
+
+function ArticleContainer(parsedCapi: Result<string, Capi>): JSX.Element {
+    const includesTweets = (parsedCapi: Result<string, Capi>): boolean => parsedCapi
+        .either(
+            () => false,
+            data => !!data.response.content.blocks.body
```
JamieB-gu: As discussed offline, think you can drop the `!!` here because `some` always returns a `boolean`?

```suggestion
            data => data.response.content.blocks.body
```

src/components/shared/articleContainer.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+    video,
+    .element-atom iframe,
+    .element-audio iframe {
+        width: 100%;
+    }
+`;
+
+function ArticleContainer(parsedCapi: Result<string, Capi>): JSX.Element {
+    const includesTweets = (parsedCapi: Result<string, Capi>): boolean => parsedCapi
```
JamieB-gu: Is it possible for this function to live outside the `ArticleContainer` scope? I don't think it relies on anything that necessitates it being a closure, and that would avoid the local argument `parsedCapi` shadowing the enclosing one?

src/components/shared/articleContainer.tsx
```diff
@@ -0,0 +1,121 @@
...
+            .some(Boolean)
+        );
+
+    const twitterScript = includesTweets(parsedCapi)
+        ? <script src="https://platform.twitter.com/widgets.js"></script>
+        : null
+
+    return <html lang="en-US" css={ArticleContainerStyles}>
+        <head>
+            <title>Article</title>
```
JamieB-gu: Presumably we'll want to pass this through from the CAPI response at some point as well?

src/components/shared/articleContainer.tsx
```diff
@@ -0,0 +1,121 @@
...
+            data => !!data.response.content.blocks.body
+            .flatMap(block => block.elements.some(element => element.type === ElementType.TWEET))
+            .some(Boolean)
+        );
+
+    const twitterScript = includesTweets(parsedCapi)
+        ? <script src="https://platform.twitter.com/widgets.js"></script>
+        : null
+
+    return <html lang="en-US" css={ArticleContainerStyles}>
```
JamieB-gu: What does this mean for `DOCTYPE`?

webb04: That's a good question maybe we'll need `response.write()` or some string concatenation.

Didn't find a way to use in React...

JamieB-gu: Something like this?

```ts
res.write('<!DOCTYPE html>');
res.write(renderToString(ArticleContainer));
res.end();
```

webb04: Yeah that might be the best option!

PR #105, opened 2019-12-10
Storybook

webb04:
## Why are you doing this?
Add storybook to project. We can test components in isolation here and use it as documentation.

### Initial impressions
+ After some webpack wrangling it seems quite easy to add components. 
+ Acts as good documentation on how components look
- The way our components accept complex capi types will slow this down and our UI testing. 
- Some components can miss global styles (fonts in the articleHeadline case). This has shown us that some components are not completely isolated.

## Changes

- `npm run storybook`
- Added `articleHeadline` component as a story
- Added storybook instructions to README

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,0 +1,3 @@
+import { configure } from '@storybook/react';
+
+configure(require.context('../src', true, /\.stories\.tsx$/), module);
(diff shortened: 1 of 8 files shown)


webb04:
Merging this now. Will make more tickets to add more components, editable props in the UI and add a GitHub action to deploy storybook to GitHub pages.

PR #106, opened 2019-12-11
Client-side Scripts

JamieB-gu:
## Why are you doing this?

In order for us to support different client-side scripts on different pages. This builds on #102 by migrating more functionality into the root component. It also narrows the scope of the `server.ts` to handling express and CAPI requests.

## Changes

- Renamed ArticleContainer to Page
- Migrated some functionality from server to Page
- Renamed `client.ts` to `article.ts`
- Created `liveblog.ts` for client-side script
- Added new webpack entry points


src/components/shared/page.tsx
```diff
@@ -0,0 +1,113 @@
...
+    figure.element-embed {
+        margin: 2em 0
+    }
+
+    .js-email-sub__iframe {
+        width: 100%;
+    }
+
+    .js-email-sub__iframe + figcaption {
+        margin-top: -8px;
```
webb04: Not related to this PR but now it's a React component we can use `basePx(-1)`

src/components/shared/page.tsx
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    video,
+    .element-atom iframe,
+    .element-audio iframe {
+        width: 100%;
+    }
+`;
+
+function ArticleBody(props: { imageSalt: string, capi: Content }): React.ReactElement {
```
webb04: Ah was literally rewriting this myself :)

webb04: Turns out there's a few different types of articles with a capi type of 'article', including `immersives`

JamieB-gu: Yeah I think we may have to write a data type for our own definition of article "type". Then we can map the MAPI response onto it:

```ts
const enum Article {
  STANDARD,
  OPINION,
  IMMERSIVE,
  LIVEBLOG,
}

const parseArticle(content: Content): Article {
  ...
}
```

PR #111, opened 2019-12-19
Native connection for thrift communication

webb04:
- Pulls in new npm package with the thrift stubs.
- This branch is being used to implement `insertAdverts` on [iOS](https://github.com/guardian/ios-live/pull/4863)
- Reports position of adverts to native layer
- Uses a mutationObserver to report ad positions if they change after a DOM change


src/client/article.ts
```diff
@@ -1,8 +1,46 @@
...
+
+function poller(interval: number, adSlotPositions: AdSlot[]) {
+    const newAdSlotPositions = getAdSlots();
+    console.log(adSlotPositions)
+    console.log(newAdSlotPositions)
+    console.log("-----------")
+    if (JSON.stringify(adSlotPositions) !== JSON.stringify(newAdSlotPositions)) {
+        // nativeClient.updateAdverts(newAdSlotPositions);
+    }
+    setTimeout(poller.bind(null, interval + 50, newAdSlotPositions), interval);
```
alexduf: Do we currently move the ad slots like that? Is there any way to listen to an event rather than poll for the position of the ad?

JamieB-gu: We had a discussion about this offline. In theory our code should be the instigator of any DOM manipulations, so yes it may be possible to fire an event whenever we make a change that we know will cause a reflow.

webb04: Yep this is how the current templates work. Another option would be to listen to DOM changes like this https://github.com/guardian/apps-rendering/pull/111/commits/3846d3d0c85cbb390c214b7c76e4e626d35910a3

But there's other events that can cause height changes such as images loading in which wouldn't count as a DOM manipulation event.

Maybe fixing image heights and using that is the best way? https://theguardian.atlassian.net/browse/MSS-1350

JamieB-gu: I think ideally we'd want the image slots to be there and of a given height in the initial render, so that when they actually load in there's no jump in page layout (as @webb04 describes for header images).

Interactives may be more difficult since they can dynamically adjust their height using code that isn't instigated by us. In this case we may have to fall back to observing mutations for each iframe that represents an interactive (thankfully this will only apply to a subset of pages).

If using React we may be able to use the [lifecycle methods](https://reactjs.org/docs/react-component.html#updating) to automatically fire the event on any re-renders.

webb04: Maybe we could change an attribute every time we change the interactive's height, that should trigger an event in the MutationObserver..

Also just came across this: https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver

JamieB-gu: I *think* you may be able to use `MutationObserver` in conjunction with `iframe`s via the [`target` argument](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/observe#Syntax) of `.observe` 🤔

Unfortunately `ResizeObserver` appears to be an Editor's Draft, and not supported by Safari.

src/client/article.ts
```diff
@@ -1,8 +1,49 @@
...
+            width: slotPosition.width,
+            height: slotPosition.height
+        })
+    });
+}
+
 setup();
+let adSlots = getAdSlots();
+
+// TODO: this can be removed after adding a min-height to images
```
alexduf: Can that be removed now?

webb04: I think it can now 👍

PR #114, opened 2019-12-20
Create components rather than set html

webb04:
## Why are you doing this?
Replace all uses of `dangerouslySetInnerHtml` and reuse the jsdom code inside of `renderBlocks`.

We can reuse this `componentFromHtml` function whenever we receive individual html fields from capi.


JamieB-gu:
🚀

PR #116, opened 2019-12-20
Re-write Blocks Rendering

JamieB-gu:
## Why are you doing this?

I was starting to hit the limitations of the current approach in `renderBlocks`. The parsing of content from CAPI was too interleaved with the rendering layer. This was only causing minor problems server-side, but client-side we need to maintain state independent of the view and deal with updates to that state.

This splits out parsing and rendering so that they happen independent of one another. Both pieces of functionality are located in the new `block` module, written around the `Block` type. This may change later.

Currently this does not touch the liveblogs, as this PR is already pretty large. Changes there will happen in a future PR.

I've moved some styles around; I *think* everything looks OK, but please shout if you spot any problems.

## Changes

- Created new `block.ts` file to handle parsing and rendering of blocks
- Updated `page.tsx` to call the new parser and renderer
- Updated article and opinion components to use changes
- Updated ArticleBody to accept children over rendering directly
- Added a `partition` helper to Result
- Moved some global styles to specific components
- Updated handling of bullets to sidestep the need for string interpolation of HTML.

### DIFF
--- a/src/block.ts
+++ b/src/block.ts
@@ -38,7 +38,7 @@ type Block = {
     linkText: string;
 } | {
     kind: ElementType.TWEET;
-    content: HTMLCollection;
+    content: NodeList;
 };
 
 type DocParser = (html: string) => DocumentFragment;
@@ -46,11 +46,11 @@ type DocParser = (html: string) => DocumentFragment;
 
 // ----- Parser ----- //
 
-const tweetContent = (tweetId: string, doc: DocumentFragment): Result<string, HTMLCollection> => {
+const tweetContent = (tweetId: string, doc: DocumentFragment): Result<string, NodeList> => {
     const blockquote = doc.querySelector('blockquote');
 
     if (blockquote !== null) {
-        return new Ok(blockquote.children);                    
+        return new Ok(blockquote.childNodes);
     }
 
     return new Err(`There was no blockquote element in the tweet with id: ${tweetId}`);
@@ -181,7 +181,7 @@ const Text = (props: { doc: DocumentFragment; pillar: Pillar }): ReactElement =>
     styledH(
         Fragment,
         null,
-        Array.from(props.doc.children).map((node, key) =>
+        Array.from(props.doc.childNodes).map((node, key) =>
             h(Fragment, { key }, textElement(props.pillar)(node))
         ),
     );
@@ -335,7 +335,7 @@ const Interactive = (props: { url: string }): ReactElement =>
         h('iframe', { src: props.url, height: 500 }, null)
     );
 
-const Tweet = (props: { content: HTMLCollection; pillar: Pillar }): ReactElement =>
+const Tweet = (props: { content: NodeList; pillar: Pillar }): ReactElement =>
     h('blockquote', null, ...Array.from(props.content).map(textElement(props.pillar)));
 
 const render = (salt: string) => (pillar: Pillar) => (block: Block, key: number): ReactNode => {
--- a/src/block.ts
+++ b/src/block.ts
@@ -1,0 +1,374 @@
+// ----- Imports ----- //
+
+import { ReactNode, createElement as h, Fragment } from 'react';
+import { css, jsx as styledH, SerializedStyles } from '@emotion/core';
+import { from, until } from '@guardian/src-foundations/mq';
+import { palette } from '@guardian/src-foundations';
+
+import { BlockElement, ElementType } from 'capiThriftModels';
+import { Result, Ok, Err } from 'types/result';
+import { Option, fromNullable } from 'types/option';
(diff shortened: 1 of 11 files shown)


src/block.ts
```diff
@@ -0,0 +1,380 @@
...
+`;
+
+const Anchor = (props: { href: string; text: string; pillar: Pillar }): ReactElement =>
+    styledH(
+        'a',
+        { css: anchorStyles(getPillarStyles(props.pillar).kicker), href: props.href },
+        props.text,
+    );
+
+const bulletStyles = (colour: string): SerializedStyles => css`
```
webb04: Won't this be needed outside of `blocks` to? e.g. standfirsts

JamieB-gu: Probably. The location of some of this stuff will likely change. It's just in the `blocks` file for now because it's only used for that at the moment.

src/block.ts
```diff
@@ -0,0 +1,381 @@
...
+        border-bottom: 1px solid ${palette.neutral[60]};
+        a {
+            &::before {
+                color: ${palette.neutral[60]};
+            }
+        }
+    `}
+`;
+
+const RichLink = (props: { url: string; linkText: string; pillar: Pillar }): ReactElement =>
```
webb04: This file isn't too large right now but if we have to handle more types in the future maybe we should pull them out into `blocks/components` or somewhere?

src/components/shared/articleBody.tsx
```diff
@@ -36,16 +33,8 @@ const ArticleBodyStyles = (pillarStyles: PillarStyles): SerializedStyles => css`
...
-    ${until.wide} {
-        figure:not(.interactive) {
-            margin-left: ${basePx(-1)};
-            margin-right: ${basePx(-1)};
-        }
-    }
-
     ${adStyles}
     ${sidePadding}
-    ${commonArticleStyles(pillarStyles)}
```
webb04: Guessing once liveblogs are migrated this function can be deleted?

src/block.ts
webb04: Might make it easier to use them as storybook components too.

src/block.ts
```diff
@@ -0,0 +1,381 @@
...
+        props.text.replace(/•/, ''),
+    );
+
+const textElement = (pillar: Pillar) => (node: Node): ReactNode => {
+    switch (node.nodeName) {
+        case 'P':
+            return h(Paragraph, null, ...Array.from(node.childNodes).map(textElement(pillar)));
+        case '#text':
+            const text = node.textContent;
+            return text?.includes('•') ? h(Bullet, { pillar, text }) : text;
```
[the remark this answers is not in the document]
JamieB-gu: At the moment, yes. I've scoped this to blocks for now to avoid this PR getting too big.

src/components/shared/articleBody.tsx
JamieB-gu: The `commonArticleStyles` function? Yes I think so.

src/block.ts
JamieB-gu: Yeah that might make sense. I'm trying to write everything in one file and then split it into smaller files when I discover from the code what those smaller files should be.

src/block.ts
```diff
@@ -0,0 +1,381 @@
...
+        default:
+            return null;
+    }
+}
+
+const Text = (props: { doc: DocumentFragment; pillar: Pillar }): ReactElement =>
+    styledH(
+        Fragment,
+        null,
+        Array.from(props.doc.children).map((node, key) =>
```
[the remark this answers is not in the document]
JamieB-gu: Yeah good idea! Updated.

JamieB-gu:
The previous problem with ad-rendering has now been fixed as of 917ccde.

webb04:
Missing `<h2>` and interactives have more padding here http://localhost:8080/world/2020/jan/03/visual-guide-airstrike-that-killed-qassam-suleimani-us-iran?a

webb04:
Byline links have merged into [one:](http://localhost:8080/world/2020/jan/03/visual-guide-airstrike-that-killed-qassam-suleimani-us-iran?a) Screen Shot 2020-01-03 at 15 57 54

src/components/immersive/immersiveArticle.tsx
```diff
@@ -30,6 +31,11 @@ const HeaderStyles = css`
         font-size: 2.6rem;
         line-height: 3.2rem;
         font-weight: 200;
+        margin-bottom: 8px;
```
webb04: basePx?

webb04:
Lgtm, future PRs can migrate `commonArticleStyles` to individual components. And we can make stories of all the individual components.

PR #117, opened 2019-12-20
Handle optional bylineHtml

webb04:
Fix an error in master.

BylineHtml can be optional, we should handle this case throughout the app.


src/contentTransformations.ts
```diff
@@ -1,4 +1,4 @@
 export const transform = (html: string): string => html
```
JamieB-gu: This `string` type isn't being conformed to at the call-site.

PR #119, opened 2019-12-31
missing byline text

webb04:
## Changes

- Add the remaining text, this was getting cut off if the lastChild node was a text element..
- Removed the unnecessary transform in standfirst (as it now happens inside `componentFromHtml`)

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/news/articleStandfirst.tsx
+++ b/src/components/news/articleStandfirst.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import { sidePadding, bulletStyles, headlineFont, darkModeCss, linkStyle } from 'styles';
-import { transform } from '../../contentTransformations';
 import { css, SerializedStyles } from '@emotion/core';
 import { palette } from '@guardian/src-foundations';
 import { PillarStyles } from 'pillar';
--- a/src/components/news/articleStandfirst.tsx
+++ b/src/components/news/articleStandfirst.tsx
@@ -59,7 +59,7 @@ const ArticleStandfirst = ({
             StandfirstDarkStyles(pillarStyles)
         ]}
     >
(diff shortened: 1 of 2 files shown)


JamieB-gu:
> Add the remaining text, this was getting cut off if the lastChild node was a text element..

I'm not sure I understand, could you elaborate on why this was happening?

webb04:
There's a bug when the html is `<a href="/profile">Author name</a> Economics correspondent`

JSDom.fragment(html).children only shows the a tag
JSDom.fragment(html).textContent shows "Author name Economics correspondent"

JamieB-gu:
Ah, interesting. I think this is happening because [`ParentNode.children`](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/children) returns a collection of [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element), which is a subset of [Node](https://developer.mozilla.org/en-US/docs/Web/API/Node) that **does not** include `Text` nodes.

The [childNodes](https://developer.mozilla.org/en-US/docs/Web/API/Node/childNodes) property would return us a `NodeList` that **does** include `Text` nodes.

JamieB-gu:
I think [this line](https://github.com/guardian/apps-rendering/blob/master/src/renderBlocks.ts#L74) could probably be migrated to `.childNodes`, since `textElement` is actually expecting the more generic `Node`.

webb04:
> I think [this line](https://github.com/guardian/apps-rendering/blob/master/src/renderBlocks.ts#L74) could probably be migrated to `.childNodes`, since `textElement` is actually expecting the more generic `Node`.

Yep that works 👏

PR #123, opened 2020-01-03
New fonts

webb04:
## Why are you doing this?
Use the fonts we're currently using in mobile-apps-article-templates

## Screenshots

| Before | After |
| --- | --- |
|  |  |


paperboyo:
Hiya. Unfortunately, these are either deprecated (agate, egyptiandisplay) or outdated (titlepiece). [These are](https://github.com/guardian/editions/tree/master/projects/Mallard/src/assets/fonts) the font files that should be used.

I hope none of the templates use deprecated fonts, too. Even current templates. But, regardless, new platform should get rid of them.

src/components/shared/page.tsx
```diff
@@ -22,15 +22,28 @@ import { insertAdPlaceholders } from 'ads';
...
-    ${fontFace("Guardian Headline", new Some(200), "/public/fonts/GHGuardianHeadline-Light.ttf")}
-    ${fontFace("Guardian Headline", new Some(400), "/public/fonts/GHGuardianHeadline-Regular.ttf")}
-    ${fontFace("Guardian Headline", new Some(500), "/public/fonts/GHGuardianHeadline-Medium.ttf")}
-    ${fontFace("Guardian Headline", new Some(700), "/public/fonts/GHGuardianHeadline-Bold.ttf")}
-    ${fontFace("Guardian Icons", new None, "/public/fonts/icons.otf")}
+    ${fontFace("Guardian Egyptian Web", new None, new None, "/public/fonts/GuardianTextEgyptian-Reg.ttf")}
+    ${fontFace("Guardian Egyptian Web", new None, new Some("italic"), "/public/fonts/GuardianTextEgyptian-RegItalic.ttf")}
+    ${fontFace("Guardian Egyptian Web", new Some(500), new None, "/public/fonts/GuardianTextEgyptian-Bold.ttf")}
+    ${fontFace("Guardian Egyptian Web", new Some("bold"), new None, "/public/fonts/GuardianTextEgyptian-Bold.ttf")}
+    ${fontFace("Guardian Egyptian Web", new Some("bold"), new Some("italic"), "/public/fonts/GuardianTextEgyptian-BoldItalic.ttf")}
```
JamieB-gu: Is this not `Guardian Text Egyptian Web`?

src/components/shared/page.tsx
```diff
@@ -22,15 +22,28 @@ import { insertAdPlaceholders } from 'ads';
...
+
+    ${fontFace("Guardian Text Sans Web", new Some(200), new None, "/public/fonts/GuardianSansWeb-Light.ttf")}
+    ${fontFace("Guardian Text Sans Web", new Some(300), new None, "/public/fonts/GuardianSansWeb-Regular.ttf")}
+    ${fontFace("Guardian Text Sans Web", new Some(600), new None, "/public/fonts/GuardianSansWeb-Semibold.ttf")}
```
JamieB-gu: Is the `Guardian Sans Web` or `Guardian Text Sans Web`? The filenames suggest it's the former 🤔

paperboyo:
Brill, thanks!

> if we have `woff2` versions of these available

We do not. These `ttf`s are specially crafted fro apps (eg. they have Postscript names intact for iOS). Also, my understanding is `woff2` do not have wide native support.

Now, we *do* have web versions of these fonts. They are used by eg. dotcom and [dotcom-rendering](https://github.com/guardian/dotcom-rendering/tree/master/src/static/fonts), but they are no good for native apps. Whether they would be good enough for all, or only some, webviews and whether it would make sense to use them in only these some webviews but use `ttf`s in others – I don’t know. The only benefit of `woff2`s is filesize.

src/components/shared/page.tsx
paperboyo: Aren’t `Guardian Egyptian Web` headline fonts (formerly Egyptian Headline, now Guardian Headline)? (easy way to check would be their use in headlines). If yes – they should be mapped to `GH Guardian Headline` family. If they are indeed Guardian Text Egyptians, though, they were/are just misnamed.

If that’s confusing, just grab me in person!

src/components/shared/page.tsx
paperboyo: As Egyptians (serifs) have headline (formerly Egyptian Headline, now Guardian Headline) and text (Egyptian Text) flavours, so do sans serifs:
Guardian Text Sans is text flavour of headline Guardian Sans.

So this mapping isn’t correct: it maps Text to Headline.

Again, grab me if needed.

src/components/shared/page.tsx
JamieB-gu: I think it's the second case, i.e. they're `Guardian Text Egyptian` and just misnamed, because the filenames are e.g. `GuardianTextEgyptian-Bold.ttf`.

JamieB-gu:
> my understanding is `woff2` do not have wide native support

> use them in only these some webviews but use `ttf`s in others

Ah, so they're in this format because it's the one that's supported by both the native layers and the web.

All the *articles* in the app should be able to support `woff`, `woff2`, `otf` etc, as they're running in webviews and thus have browser-level support. Other content, such as fronts, nav etc. will only be able to support `.ttf` because they're native...

I guess the question is whether we want:

a) one set of font files (`.ttf`) for the native layers and another set of font files (`woff`/`woff2`) for the webviews, or
b) `.ttf` for both native and the webviews, and just accept that they're not optimal for the web content

paperboyo:
> I guess the question

Yep. And the answer is likely b) since the only benefit of woff2 is filesize and we would need to bundle both anywayz…

webb04:
I'm slowly understanding more about these fonts... I've used the same fonts as the editions for apart from titlepiece, icons and GuardianTextEgyptian-Black

paperboyo:
Titlepiece needs to be replaced with new one (they are very different). Wasn't there one in Editions repo (sorry on a phone)?

Will check Text Egyptian Black (although, I haven't seen it used ANYWHERE, so maybe better kill it?) 

Icons are the same.

.storybook/preview-head.html
```diff
@@ -4,16 +4,31 @@
...
+    @font-Face{font-family:"Guardian Headline"; font-weight:600; font-weight: "italic"; src:url( "./public/fonts/GHGuardianHeadline-SemiboldItalic.ttf")}
+    @font-Face{font-family:"Guardian Headline"; font-weight:700; src:url("/public/fonts/GHGuardianHeadline-Bold.ttf")}
+    @font-Face{font-family:"Guardian Headline"; font-weight:700; font-weight: "italic"; src:url("./public/fonts/GHGuardianHeadline-BoldItalic.ttf")}
+
+    @font-Face{font-family:"Guardian Icons"; src:url("/public/fonts/icons.otf")}
+
+    body{margin:0;font-family:'Guardian Text Egyptian Web';font-size:1.6em;line-height:1.5em;overflow-x:hidden;}
     figure{margin:1em 0;}
     figure.element-embed{margin:2em 0;}
     .js-email-sub__iframe{width:100%;}
```
JamieB-gu: As @paperboyo mentioned offline, could the 500 weight fonts be referred to as "medium", and the 700 weights be referred to as "bold"? Would that cause any problems?

Also, a number of the `font-face` declarations in this file refer to `font-weight:"italic"`... do you perhaps mean `font-style:"italic"`?

src/components/shared/page.tsx
```diff
@@ -23,22 +23,36 @@ import { Article, fromCapi } from 'article';
...
+    ${fontFace("Guardian Headline", new Some(400), new None, "/public/fonts/GHGuardianHeadline-Regular.ttf")}
+    ${fontFace("Guardian Headline", new Some(400), new Some("italic"), "/public/fonts/GHGuardianHeadline-RegularItalic.ttf")}
+    ${fontFace("Guardian Headline", new Some(500), new None,  "/public/fonts/GHGuardianHeadline-Medium.ttf")}
+    ${fontFace("Guardian Headline", new Some(500), new Some("italic"),  "/public/fonts/GHGuardianHeadline-MediumItalic.ttf")}
+    ${fontFace("Guardian Headline", new Some(600), new None,  "/public/fonts/GHGuardianHeadline-Semibold.ttf")}
+    ${fontFace("Guardian Headline", new Some(600), new Some("italic"),  "/public/fonts/GHGuardianHeadline-SemiboldItalic.ttf")}
+    ${fontFace("Guardian Headline", new Some(700), new None, "/public/fonts/GHGuardianHeadline-Bold.ttf")}
+    ${fontFace("Guardian Headline", new Some(700), new Some("italic"), "/public/fonts/GHGuardianHeadline-BoldItalic.ttf")}
+
+    ${fontFace("Guardian Icons", new None, new None, "/public/fonts/icons.otf")}
```
JamieB-gu: Same comment about the font weights as above, i.e. 500 = medium, 700 = bold?

.storybook/preview-head.html
paperboyo: I’m as familiar with CSS as American Dad fighting windows blinds, but my understanding (or the lack of it) is:
1. Correct numbers for weights are like that (forget the font name, it applies to weights from all of them):
image
So they should be in the code *unless* some (old?) content may ask for `500` meaning Bold, but since bold will be `700`, the call we actually end up returning Regular. If that’s the case, we may want to decide to keep wrong numbers for the sake of bakwards compatibility.
2. `font-weight:"italic"` indeed sounds bonkers 😆

src/components/shared/page.tsx
```diff
@@ -23,22 +23,37 @@ import { Article, fromCapi } from 'article';
...
+    ${fontFace("Guardian Text Egyptian Web", new Some(400), new Some("italic"), "/public/fonts/GuardianTextEgyptian-RegItalic.ttf")}
+    ${fontFace("Guardian Text Egyptian Web", new Some(700), new None, "/public/fonts/GuardianTextEgyptian-Bold.ttf")}
+    ${fontFace("Guardian Text Egyptian Web", new Some(700), new Some("italic"), "/public/fonts/GuardianTextEgyptian-BoldItalic.ttf")}
+    ${fontFace("Guardian Text Egyptian Web", new Some("bold"), new None, "/public/fonts/GuardianTextEgyptian-Bold.ttf")}
+    ${fontFace("Guardian Text Egyptian Web", new Some("bold"), new Some("italic"), "/public/fonts/GuardianTextEgyptian-BoldItalic.ttf")}
+
+    ${fontFace("Guardian Text Sans Web", new Some(400), new None, "/public/fonts/GuardianTextSans-Regular.ttf")}
+    ${fontFace("Guardian Text Sans Web", new Some(400), new Some("italic"), "/public/fonts/GuardianTextSans-RegularItalic.ttf")}
+    ${fontFace("Guardian Text Sans Web", new Some(600), new None, "/public/fonts/GuardianTextSans-Bold.ttf")}
+    ${fontFace("Guardian Text Sans Web", new Some(600), new Some("italic"), "/public/fonts/GuardianTextSans-BoldItalic.ttf")}
```
JamieB-gu: Should these be 700?

paperboyo: Yep.

PR #127, opened 2020-01-06
Set height on images

webb04:
## Why are you doing this?
Set a min height on images (with a background colour) to avoid content jumping when the images load.

This is not the same ratio as before. e.g. removing these styles will show a smaller image... maybe this calc needs changing or we pass another height param to the fastly image resizer?

### DIFF
--- a/src/block.ts
+++ b/src/block.ts
@@ -26,10 +26,8 @@ type Block = {
     displayCredit: boolean;
     credit: string;
     file: string;
-    typeData?: {
-        width: number;
-        height: number;
-    };
+    width: number;
+    height: number;
 } | {
     kind: ElementType.PULLQUOTE;
     quote: string;
@@ -72,15 +70,15 @@ const parser = (docParser: DocParser) => (block: BlockElement): Result<string, B
             const masterAsset = block.assets.find(asset => asset.typeData.isMaster);
             const { alt, caption, displayCredit, credit } = block.imageTypeData;
             const imageBlock: Option<Result<string, Block>> = fromNullable(masterAsset)
-                .map(asset => asset.file)
-                .map(file => new Ok({
+                .map(asset => new Ok({
                     kind: ElementType.IMAGE,
                     alt,
                     caption,
                     displayCredit,
                     credit,
-                    file,
-                    typeData: masterAsset?.typeData
+                    file: asset.file,
+                    width: asset.typeData.width,
+                    height: asset.typeData.height,
                 }));
 
             return imageBlock.withDefault(new Err('I couldn\'t find a master asset'));
@@ -219,10 +217,8 @@ interface ImageProps {
     caption: string;
     displayCredit: boolean;
     credit: string;
-    typeData?: {
-        width: number;
-        height: number;
-    };
+    width: number;
+    height: number;
 }
 
 const imageStyles = css`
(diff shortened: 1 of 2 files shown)


webb04:
May need to use cover and supply the height and widths to Fastly

src/block.ts
```diff
@@ -25,6 +26,10 @@ type Block = {
     displayCredit: boolean;
     credit: string;
     file: string;
+    typeData?: {
+        width: number;
+        height: number;
+    };
```
JamieB-gu: Does this need to be nested? Could we simply do:

```ts
{
  ...
  width: number;
  height: number;
}
```

src/block.ts
```diff
@@ -75,6 +80,7 @@ const parser = (docParser: DocParser) => (block: BlockElement): Result<string, B
                     displayCredit,
                     credit,
                     file,
+                    typeData: masterAsset?.typeData
```
JamieB-gu: `masterAsset` is available as a non-optional in the `asset` argument here, so I think you can do:

```ts
const imageBlock: Option<Result<string, Block>> = fromNullable(masterAsset)
  .map(asset => new Ok({
    kind: ElementType.IMAGE,
    alt,
    caption,
    displayCredit,
    credit,
    file: asset.file,
    width: asset.typeData.width,
    height: asset.typeData.height,
}));
```

src/block.ts
```diff
@@ -213,6 +219,10 @@ interface ImageProps {
     caption: string;
     displayCredit: boolean;
     credit: string;
+    typeData?: {
+        width: number;
+        height: number;
+    };
```
JamieB-gu: Above comment concerning nested properties.

src/block.ts
```diff
@@ -245,16 +255,19 @@ const imageStyles = css`
     }
 `;
 
-const Image = ({ url, alt, salt, caption, displayCredit, credit }: ImageProps): ReactElement =>
-    styledH('figure', { css: imageStyles },
+const Image = (imageProps: ImageProps): ReactElement => {
+    const { url, alt, salt, caption, displayCredit, credit, typeData } = imageProps;
```
JamieB-gu: Minor point, but the [React convention](https://reactjs.org/docs/components-and-props.html) is to name the parameter `props`.

src/components/blocks/image.ts
```diff
@@ -16,20 +20,38 @@ interface Image {
 
 // ----- Functions ----- //
 
+const imageRatioStyles = (sizes = "100vw", width: number, height: number): SerializedStyles => css`
+        --size: ${sizes};
+        height: calc(var(--size) * ${height / width});
```
JamieB-gu: Out of curiosity, what is the CSS variable used for here?

src/components/blocks/image.ts
```diff
@@ -16,20 +20,38 @@ interface Image {
...
+                sizes,
+                srcSet: AssetUtils.toSrcset(salt, assets).withDefault(''),
+                alt,
+                src: AssetUtils.toUrl(salt, assets[0]),
+            });
+    }
 
 const immersiveImageElement = element('calc(80vh * 5/3)');
-const imageElement = element('100%');
+const imageElement = element('100vw');
```
JamieB-gu: Does `100%` no longer work here?

src/components/blocks/image.ts
webb04: It was a way to use a `calc` inside a `calc`. Might be able to refactor it some way

src/components/blocks/image.ts
webb04: It would collapse the height of the placeholder to 0

src/components/blocks/image.ts
JamieB-gu: 🤔 where is there a `calc` inside a `calc?

webb04: `sizes` can be a calc :)

PR #128, opened 2020-01-07
Catch undefined element

webb04:
## Why are you doing this?
Error trying to access html on an undefined element.
http://localhost:8080/football/2020/jan/06/joy-of-fa-cup-osaze-urhoghide-antidote-critics?a

### DIFF
--- a/src/block.ts
+++ b/src/block.ts
@@ -98,7 +98,7 @@ const parser = (docParser: DocParser) => (block: BlockElement): Result<string, B
             return new Ok({ kind: ElementType.RICH_LINK, url, linkText });
 
         case ElementType.TWEET:
-            return tweetContent(block.tweetTypeData.id, docParser(block.textTypeData?.html))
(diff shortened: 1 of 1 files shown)


src/block.ts
```diff
@@ -98,7 +98,7 @@ const parser = (docParser: DocParser) => (block: BlockElement): Result<string, B
             return new Ok({ kind: ElementType.RICH_LINK, url, linkText });
 
         case ElementType.TWEET:
-            return tweetContent(block.tweetTypeData.id, docParser(block.textTypeData.html))
+            return tweetContent(block.tweetTypeData.id, docParser(block.textTypeData?.html))
```
JamieB-gu: Ah, I've actually fixed this on a separate branch. That change won't actually render the tweet, I think what you want is?

```suggestion
            return tweetContent(block.tweetTypeData.id, docParser(block.tweetTypeData.html))
```

PR #129, opened 2020-01-07
Article Type - Encapsulate Pillar and Layout

JamieB-gu:
## Why are you doing this?

I think all articles are described by two key pieces of information. First is the `Pillar`, which we have an existing type for. Second is a less well-defined concept that I've called "layout". This is basically a description of the various ways articles vary in design based on a number of factors:

- Is it an opinion piece?
- Is it an analysis?
- Does it have the feature tone?
- Is it a liveblog?
- Does it have the 'immersive' `displayHint`?

etc.

I've attempted to define a sum type called `Layout` to capture these differences, which can be derived from the fields available in the CAPI response.

A given article will fall in the set created by taking the product of `Pillar` and `Layout`, as any article layout can theoretically fall under any of the Pillars. Therefore I've created the `Article` product type to describe this:

```ts
type Article = {
    layout: Layout;
    pillar: Pillar;
};
```

FYI @gtrufitt @oliverlloyd 

## Changes

- Added `Article` product type, encapsulates `Pillar` and `Layout`
- All articles can now use this product type
- `Layout` is derived from a range of CAPI fields
- Added further CAPI helpers to check if an article is a 'review' or an 'immersive'

### DIFF
--- a/src/article.ts
+++ b/src/article.ts
@@ -1,0 +1,79 @@
+// ----- Imports ----- //
+
+import { Pillar, pillarFromString } from 'pillar';
+import { Content } from 'capiThriftModels';
+import { isFeature, isAnalysis, isImmersive, isReview } from 'capi';
+
+
+// ----- Types ----- //
+
+const enum Layout {
+    Default,
+    Immersive,
+    Feature,
+    Review,
+    Analysis,
+    Opinion,
+    Liveblog,
+    Gallery,
+    Interactive,
+    Picture,
+    Video,
+    Audio,
+}
+
+type Article = {
+    layout: Layout;
+    pillar: Pillar;
+};
+
+
+// ----- Functions ----- //
+
+function layoutFromCapi(content: Content): Layout {
+    switch (content.type) {
+        case 'article':
+            if (pillarFromString(content.pillarId) === Pillar.opinion) {
+                return Layout.Opinion;
+            } else if (isImmersive(content)) {
+                return Layout.Immersive;
+            } else if (isFeature(content)) {
+                return Layout.Feature;
+            } else if (isReview(content)) {
+                return Layout.Review;
+            } else if (isAnalysis(content)) {
+                return Layout.Analysis;
+            }
+            return Layout.Default;
+        case 'liveblog':
+            return Layout.Liveblog;
+        case 'gallery':
+            return Layout.Gallery;
+        case 'interactive':
+            return Layout.Interactive;
+        case 'picture':
+            return Layout.Picture;
+        case 'video':
+            return Layout.Video;
+        case 'audio':
+            return Layout.Audio;
+        default:
+            return Layout.Default;
+    }    
+}
+
+function fromCapi(content: Content): Article {
+    return {
+        layout: layoutFromCapi(content),
+        pillar: pillarFromString(content.pillarId),
+    };
+}
+
+
+// ----- Exports ----- //
+
+export {
+    Layout,
+    Article,
+    fromCapi,
+};
--- a/src/article.ts
+++ b/src/article.ts
@@ -8,7 +8,7 @@ import { isFeature, isAnalysis, isImmersive, isReview } from 'capi';
 // ----- Types ----- //
 
 const enum Layout {
-    Default,
+    Standard,
     Immersive,
     Feature,
     Review,
@@ -44,7 +44,7 @@ function layoutFromCapi(content: Content): Layout {
             } else if (isAnalysis(content)) {
                 return Layout.Analysis;
             }
-            return Layout.Default;
+            return Layout.Standard;
         case 'liveblog':
             return Layout.Liveblog;
         case 'gallery':
@@ -58,7 +58,7 @@ function layoutFromCapi(content: Content): Layout {
         case 'audio':
             return Layout.Audio;
         default:
-            return Layout.Default;
+            return Layout.Standard;
     }    
 }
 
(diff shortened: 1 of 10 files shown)


oliverlloyd:
For DCR we take a similar approach. We use different Layout files to contain the layout logic, such as:

`[ShowcaseLayout](https://github.com/guardian/dotcom-rendering/blob/0a61bb1b971fa037fe310dd3703fe70094fe42bf/src/web/layouts/ShowcaseLayout.tsx)`
`[StandardLayout](https://github.com/guardian/dotcom-rendering/blob/0a61bb1b971fa037fe310dd3703fe70094fe42bf/src/web/layouts/StandardLayout.tsx)`
and soon:
`ImmersiveLayout`
`OpinionLayout`

We have a [single component](https://github.com/guardian/dotcom-rendering/blob/308b35e88399c2e3efa9393d37606ea424a2ef2d/src/web/layouts/DecideLayout.tsx) where we make the branch for these.

We initially planned to have a `FeatureLayout`, etc, but in the end these decisions ended up being made further down the stack. For example [ArticleHeadline uses a switch](https://github.com/guardian/dotcom-rendering/blob/0a61bb1b971fa037fe310dd3703fe70094fe42bf/src/web/components/ArticleHeadline.tsx#L186) to render differently based on `designType` and `pillar`. This could still change though as we scale up.

oliverlloyd:
By the way, I'm so stealing your [decision logic](https://github.com/guardian/apps-rendering/pull/129/files#diff-b18b5d92d3f0a7671e5d35b6756e2c8eR121) around how to choose article type, sorry!

JamieB-gu:
> We have a [single component](https://github.com/guardian/dotcom-rendering/blob/308b35e88399c2e3efa9393d37606ea424a2ef2d/src/web/layouts/DecideLayout.tsx) where we make the branch for these.

Ah, interesting parallel evolution there. I *think* we do something similar in [this file](https://github.com/guardian/apps-rendering/blob/master/src/components/shared/page.tsx#L86).

> I'm so stealing your decision logic

😂

JamieB-gu:
> being made further down the stack

We do something like that at the moment for [feature headlines](https://github.com/guardian/apps-rendering/blob/master/src/components/news/articleHeadline.tsx#L41). In fact looking at your headline code I think the approach I'm taking in this PR might end up looking quite like that file.

src/article.ts
```diff
@@ -0,0 +1,79 @@
...
+    Default,
+    Immersive,
+    Feature,
+    Review,
+    Analysis,
+    Opinion,
+    Liveblog,
+    Gallery,
+    Interactive,
+    Picture,
```
webb04: Is this `PhotoEssay`?

src/components/shared/page.tsx
```diff
@@ -80,26 +82,27 @@ function getArticleSubtype(capi: Content): FunctionComponent<ArticleProps> {
         return ImmersiveArticle;
     }
   
-    return Article;
+    return ArticleComponent;
```
webb04: If Default is the term we will use shouldn't that be used here too? e.g. `DefaultArticle` to be consistent with `OpinionArticle`, `LiveblogArticle` ...

All of the `article` components live inside a `news` folder, should that now be a default folder?

src/components/news/article.tsx
```diff
@@ -55,13 +57,12 @@ const HeaderImageStyles = css`
...
-const Article = ({ capi, imageSalt, children }: ArticleProps): JSX.Element => {
+const Article = ({ capi, imageSalt, article, children }: ArticleProps): JSX.Element => {
 
-    const { fields, tags, webPublicationDate, pillarId } = capi;
+    const { fields, tags, webPublicationDate } = capi;
     const series = articleSeries(capi);
-    const feature = isFeature(capi) || 'starRating' in fields;
-    const pillar = pillarFromString(pillarId);
-    const pillarStyles = getPillarStyles(pillar);
+    const feature = article.layout === Layout.Feature || article.layout === Layout.Review;
```
webb04: Maybe another PR as it's not related but `feature` could be changed to `boldHeadline` as that's a bit more accurate of what it does?

src/components/news/article.tsx
```diff
@@ -90,7 +91,7 @@ const Article = ({ capi, imageSalt, children }: ArticleProps): JSX.Element => {
                             className={articleWidthStyles}
                         />
                     </div>
-                    <Keyline pillar={pillar} type={'article'}/>
+                    <Keyline article={article}/>
```
webb04: Good job removing the string!

src/article.ts
JamieB-gu: This came from the [`ContentType` enum](https://github.com/guardian/apps-rendering/blob/master/src/capiThriftModels.ts#L26). Assuming it's the same thing, I don't mind too much what it's called. If it's *not* the same thing then we may need a `PhotoEssay` layout.

src/components/shared/page.tsx
JamieB-gu: I think so. An alternative name could also be `Standard`, but that could be somewhat ambiguous...

src/components/news/article.tsx
JamieB-gu: When we're describing an article of that type what term would we use?

src/article.ts
webb04: It sounds like it's the same thing so makes sense to keep the capi type.

src/components/shared/page.tsx
webb04: I thought about `Standard` too and it looks like dcr are using that term

src/components/news/article.tsx
webb04: Oh no I'm referring to the `feature` prop which is basically just a boolean to show a bold headline.

JamieB-gu: As discussed offline, I think I'm going to drop this prop in a follow-up.

PR #130, opened 2020-01-07
Comment counts

webb04:
## Why are you doing this?
To reach feature parity of the current templates, we'll need comment counts.
I can't see the counts in the CAPI response so it might be fetched from the discussion service in MAPI. Once we receive MAPI data we can pass these values in. 

## Changes

- Standard article comment count
- Opinion comment count
- Liveblog comment count

## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |
|  |  |


src/components/opinion/opinionArticle.tsx
```diff
@@ -85,12 +103,23 @@ function OpinionArticle({ capi, imageSalt, article, children }: OpinionArticlePr
...
-                        contributors={contributors}
-                        className={articleWidthStyles}
-                    />
+
+                    <section css={[articleWidthStyles, topBorder]}>
+                        <OpinionByline
+                            pillarStyles={pillarStyles}
+                            publicationDate={webPublicationDate}
+                            contributors={contributors}
+                            className={css``}
```
JamieB-gu: ?

src/components/shared/commentCount.tsx
```diff
@@ -0,0 +1,43 @@
...
+        line-height: 1.8rem;
+        border: none;
+        color: ${colour};
+    }
+`
+
+export const CommentCount = ({ count, colour, className }: CommentCountProps): JSX.Element => {
+    return <div css={[CommentCountStyles(colour), className]}>
+        <span></span>
+        <button>{count}</button>
```
JamieB-gu: Does this drop you to the bottom of the page where the comments are, as on dotcom? If so, should this be an anchor?

src/components/liveblog/liveblogByline.tsx
```diff
@@ -45,15 +44,33 @@ const LiveblogBylineStyles = ({ liveblogBackground }: PillarStyles): SerializedS
...
+`
+
 interface LiveblogBylineProps {
     byline?: string;
     pillarStyles: PillarStyles;
     publicationDate: string;
     contributors: Contributor[];
     article: Article;
     imageSalt: string;
+    commentable: boolean;
```
JamieB-gu: If this will in future also need the count, could we combine both pieces of information into a single type? Where a value of `Some(<number_of_comments>)` would both display the comment component *and* pass it the number of comments, and a value of `None` wouldn't display the comment component at all.

```suggestion
    commentCount: Option<number>;
```

src/components/shared/commentCount.tsx
webb04: I'm a bit undecided here, we could have a `<button onClick={nativeClient.viewComments()}></button>` or a link like we have now...

src/components/liveblog/liveblogByline.tsx
webb04: 👍 Maybe one for when we get the MAPI data?

PR #132, opened 2020-01-08
Fix react key warnings

webb04:
## Why are you doing this?
Fixes three different React key warnings for this article: http://localhost:8080/lifeandstyle/2020/jan/07/can-sun-simulator-improve-winter-mood-lethargy?a

### DIFF
--- a/src/ads.ts
+++ b/src/ads.ts
@@ -9,7 +9,7 @@ function insertAdPlaceholders(reactNodes: ReactNode[]): ReactNode[] {
 
     const className = numParas < 15 ? 'ad-placeholder short' : 'ad-placeholder';
 
-    const ad = h('aside', { className },
+    const ad = (key: number) => h('aside', { className, key },
         h('div', { className: 'ad-labels' },
             h('h1', null, 'Advertisement'),
             h('button', { className: 'ad-hide' },
@@ -19,7 +19,7 @@ function insertAdPlaceholders(reactNodes: ReactNode[]): ReactNode[] {
     );
 
     const insertAd = (para: number, nodes: ReactNode[]): ReactNode[] =>
-        adIndices.includes(para) ? [...nodes, ad] : nodes;
+        adIndices.includes(para) ? [...nodes, ad(++para)] : nodes;
 
     return flattenedNodes
         .reduce(([paraNum, prevNodes], node) => {
--- a/src/ads.ts
+++ b/src/ads.ts
@@ -1,4 +1,4 @@
-import { createElement as h, ReactNode } from 'react';
+import { createElement as h, ReactNode, ReactElement } from 'react';
 
 function insertAdPlaceholders(reactNodes: ReactNode[]): ReactNode[] {
     const adIndices = [3, 9];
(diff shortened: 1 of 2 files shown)


src/ads.ts
```diff
@@ -19,7 +19,7 @@ function insertAdPlaceholders(reactNodes: ReactNode[]): ReactNode[] {
     );
 
     const insertAd = (para: number, nodes: ReactNode[]): ReactNode[] =>
-        adIndices.includes(para) ? [...nodes, ad] : nodes;
+        adIndices.includes(para) ? [...nodes, ad(++para)] : nodes;
```
alexduf: I think `++para` is confusing here. It mutates a parameter before passing it as an argument to the `ad` function, but we do nothing with the newly modified value of `para`.

Mutating parameters is to be avoided as in most case won't do much (for small types such as `number`), but sometimes will affect the actual value (if the function receives an object for instance). It makes it hard from the caller point of view to understand what's happening.

I think simply passing `ad(para + 1)` will achieve the same effect without the confusion

src/renderBlocks.ts
```diff
@@ -179,6 +179,8 @@ function render(bodyElements: BlockElement[], imageSalt: string, ads = true): Re
 
 function componentFromHtml(html: string): ReactNode[] {
     const fragment = JSDOM.fragment(transform(html))
+    Array.from(fragment.children)
+        .forEach((element, index) => element.setAttribute('key', `${index}`))
```
alexduf: I've been trying to find another way to achieve the same thing, but I can't come up with something without having to fall into hashing the content of a block to identify it...

Somehow it feels wrong to simply assign a number with each component. Any thoughts on that?

src/ads.ts
JamieB-gu: I didn't know `++foo` was a thing you could do in JavaScript! TIL...

@webb04 what's the reason for incrementing `para` here?

src/renderBlocks.ts
JamieB-gu: Is it the concept of using a number in general that feels wrong, or is it manipulating the DOM to add an attribute?

I think applying an index to *React components* is a fairly straightforward method when working with arrays, because you typically already have it if you're mapping/reducing, Assigning a randomly generated value (e.g. using uuid) is another technique, although I'm not keen on that because it breaks referential transparency.

This is solved in the new renderer [like this](https://github.com/guardian/apps-rendering/blob/f8b2c7033138363510f7255ad4a6798a749f2239/src/block.ts#L189), which is possible because the code is structured differently.

JamieB-gu: Having done some research I'm pretty sure I'm wrong about this... You're right @alexduf, the [react docs](https://reactjs.org/docs/lists-and-keys.html#keys) explain that using index as the key is a problem if the list is likely to change.

In practice, I don't think we'd see any problems from this because it's a single render. However, on the client this may result in bugs due to possible re-renders (particularly on the liveblog). As we're sharing code between the client and server, perhaps it would be best to eliminate this pattern entirely. @webb04 @alexduf what do you think?

src/ads.ts
webb04: At the time I wanted that ad to have a different key to the paragraphs, but I realise it would have still had a duplicated key.

I'll add something to generate IDs using `uuid` or something.

JamieB-gu: Opinions may vary on how much of an issue this is, but using UUID would introduce side effects into all the functions that make use of keys somewhere down their call stack...

src/renderBlocks.ts
JamieB-gu: I think we can decide that this is a relevant issue only on the client.

Since the server-side code and client-side code are built separately, perhaps we can disable this warning for the server build only?

webb04: Doesn't seem an easy way to turn off warnings, unless we use production React which will remove a lot more warnings.

webb04: `componentFromHtml` should never make it's way to the client.

alexduf: I guess it makes sense then

PR #137, opened 2020-01-10
Expand Article Type

JamieB-gu:
## Why are you doing this?

This solves a few problems:

- The code to parse the data coming in from CAPI is separate from the view code.
- The parsing code lives in one place, `article.ts`, so any changes (for example if the data structure is modified or expanded) should result in minimal code changes.
- Usage of `JSDOM` no longer breaks storybook/client-side code because a) it happens higher up the stack at the parsing stage, not inside React components, and b) its usage has been abstracted into a generic "document parser", which can be replaced by `DOMParser` on the client if needs be.

FYI @oliverlloyd @gtrufitt you may be interested in the `article.ts`, `page.tsx` and `renderer.ts` files.

## Changes

- Added fields to `Article` to include all data needed to render article
- Moved CAPI parsing code from `block.ts` into `article.ts`
- Split renderer code out from `block.ts` into own file
- Deleted `block.ts` file
- Migrated standard article to use `Article` type instead of CAPI
- Some refactoring of standard standfirst

### DIFF
--- a/src/ads.test.ts
+++ b/src/ads.test.ts
@@ -1,10 +1,11 @@
 import { insertAdPlaceholders } from './ads';
 import { ReactNode } from 'react';
-import { Block, renderAll } from 'block';
+import { renderAll } from 'renderer';
 import { ElementType } from 'capiThriftModels';
 import { JSDOM } from 'jsdom';
 import { Pillar } from 'pillar';
 import { compose } from 'lib';
+import { Block } from 'article';
 
 const textBlock = (nodes: string[]): Block =>
     ({
--- a/src/article.ts
+++ b/src/article.ts
@@ -1,8 +1,10 @@
 // ----- Imports ----- //
 
 import { Pillar, pillarFromString } from 'pillar';
-import { Content } from 'capiThriftModels';
-import { isFeature, isAnalysis, isImmersive, isReview } from 'capi';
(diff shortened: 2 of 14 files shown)


gtrufitt:
This is interesting actually (cc/ @oliverlloyd). Our 'element types' don't directly map to CAPI element types ([for example we breakout 'subheading' and 'blockquote'](https://github.com/guardian/frontend/blob/30c68288fc50ae6495e5134bcdf2a48586ed6098/common/app/model/dotcomrendering/pageElements/PageElement.scala#L171)) so using the Thrift Definitions directly wouldn't be 100% applicable in our case?

JamieB-gu:
@gtrufitt @oliverlloyd we've actually diverged from using the Thrift types directly in our UI code, as we're trying to isolate knowledge of the CAPI response to our parsing code. So elements are defined in terms of our own types [here](https://github.com/guardian/apps-rendering/blob/master/src/article.ts#L81).

Our thinking is that the CAPI types are optimised for representing the CAPI API, but not necessarily for the data structures that we're using in our own rendering code. We now do something like this:

```ts
import { Content } from 'capiTypes';

type Article = ...

declare function parse(content: Content): Article;
declare function render(article: Article): ReactNode;

const reactNode = render(parse(content));

if (server) {
  return ReactDOM.renderToString(reactNode);
} else if (client) {
  ReactDOM.render(reactNode, pageElement);
}
```

PR #138, opened 2020-01-10
Tidy Standard Article

JamieB-gu:
## Why are you doing this?

I've been moving a lot of code around recently, mostly in the `Standard` article, so a follow-up refactor seemed a good idea. Also, our component files seem to have evolved to follow a common structure, so this attempts to standardise that structure somewhat.

## Changes

- Organised imports
- Applied common file structure
- Simplified names

### DIFF
--- a/src/components/standard/article.tsx
+++ b/src/components/standard/article.tsx
@@ -1,37 +1,31 @@
 // ----- Imports ----- //
 
 import React, { ReactNode } from 'react';
-
-import HeaderImage from 'components/shared/headerImage';
-import ArticleSeries from 'components/shared/articleSeries';
-import ArticleHeadline from 'components/standard/headline';
-import ArticleStandfirst from 'components/standard/standfirst';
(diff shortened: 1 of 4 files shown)


src/components/standard/article.tsx
```diff
@@ -1,37 +1,31 @@
...
-import ArticleBody from 'components/shared/articleBody';
+import Body from 'components/shared/articleBody';
 import Tags from 'components/shared/tags';
 import { darkModeCss, articleWidthStyles } from 'styles';
-import { palette } from '@guardian/src-foundations';
-import { from, breakpoints } from '@guardian/src-foundations/mq';
-import { css } from '@emotion/core';
 import { Keyline } from 'components/shared/keyline';
 import { getPillarStyles } from 'pillar';
 import { Article } from 'article';
```
webb04: Another possible future cleanup - consistently use `export default`?

JamieB-gu: For types?

PR #142, opened 2020-01-13
Post endpoint

webb04:
## Why are you doing this?

Post endpoint to accept article data and return the rendered page.
Currently accepts JSON but future work can use Thrift.

### DIFF
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -165,7 +165,7 @@ app.get('/*', async (req, res) => {
   }
 });
 
-app.post('/article', async (req, res) => {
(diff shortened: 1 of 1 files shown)


src/server/server.ts
```diff
@@ -141,8 +163,11 @@ app.get('/*', async (req, res) => {
   } else {
     queryMapi(req, res);
   }
-
 });
 
+app.post('/article', async (req, res) => {
```
JamieB-gu:
```suggestion
app.post('/article', serveArticlePost);
```

PR #143, opened 2020-01-13
Language: en-US -> en

JamieB-gu:
## Why are you doing this?

I think this makes more sense as the default for now. However, as we have several variations of English content (thinking about US and AU here), perhaps we want to do one of the following as an alternative:

1. Use `en` to specify English generically, without tying it to a specific region.
2. Editionalise this attribute, so it specifies `en-GB`, `en-US`, `en-AU` etc. for different editions.

As far as I can tell, dotcom seem to do option 1. Is this correct @gtrufitt @shtukas @nicl ?

## Changes

- Updated the html `lang` attribute to use `GB` instead of `US`.

### DIFF
--- a/src/components/shared/page.tsx
+++ b/src/components/shared/page.tsx
@@ -130,7 +130,7 @@ function Page({ content, imageSalt }: Props): JSX.Element {
         : null
 
     return (
-        <html lang="en-US" css={PageStyles}>
+        <html lang="en-GB" css={PageStyles}>
             <head>
                 <title>{content.id}</title>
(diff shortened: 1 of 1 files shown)


JamieB-gu:
Validation failing due to problems in `master`, don't think it's related to this PR.

webb04:
We've been using `en-US` for the last 6 years in the app... 
`en` sounds like a good idea until we can add editions to the mapi endpoint

Screen Shot 2020-01-13 at 11 11 46

PR #145, opened 2020-01-13
Refactor Opinion To Match Standard

JamieB-gu:
## Why are you doing this?

This brings the changes made to the 'Standard' article over to Opinion. It also makes `Author` a standalone shared component. It's mostly renaming and re-ordering.

## Changes

- Updated file structure
- Updated naming conventions
- Extracted Author into shared component
- Removed need to pass CAPI object to Opinion

### DIFF
--- a/src/components/shared/author.tsx
+++ b/src/components/shared/author.tsx
@@ -1,0 +1,27 @@
+// ----- Imports ----- //
+
+import { createElement as h } from 'react';
+
+import { renderText } from 'renderer';
+import { Pillar } from 'pillar';
+import { Option } from 'types/option';
+
+
+// ----- Component ----- //
+
+interface Props {
+    byline: Option<DocumentFragment>;
+    pillar: Pillar;
+}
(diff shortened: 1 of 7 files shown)


src/components/shared/author.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+interface Props {
+    byline: Option<DocumentFragment>;
+    pillar: Pillar;
+}
+
+const Author = (props: Props): JSX.Element | null =>
+    props.byline.map<JSX.Element | null>((bylineHtml: DocumentFragment) =>
+        // This is not an iterator, ESLint is confused
+        // eslint-disable-next-line react/jsx-key
```
webb04: Should we rename `map`? It will mean we can remove all the rule skips and might avoid some confusion for new developers (my first impression is that it's iterating)

JamieB-gu: `fmap`?

PR #146, opened 2020-01-15
Deserialise thrift and return html

webb04:
## Why are you doing this?
Post endpoint receives a `Content` object in thrift bytes and returns the rendered html.

This PR has expanded into a number of things (new endpoint, deserialising thrift from mapi, adding new types, handling differences of new types such as optionals).

I'll leave this PR as reference and open smaller ones based on subtasks.


alexduf:
Should this one be closed then?

PR #148, opened 2020-01-16
New mapi/capi types

webb04:
MSS-1382.

Autogenerated types from capi. The current generated types don't have codecs or read/write functions for deserialisation. They also don't reflect the capi models as some fields should be optional.

Future PRs will swap the existing types for these new ones.

Second commit adds generation instructions to the README.md. Although this will change when the types change (we'll probably need our own thrift definitions).


alexduf:
Someone's about to blow up the LoC stats on that repo :)

PR #151, opened 2020-01-16
replace imports

webb04:
Pausing this PR as it's likely to conflict with @JamieB-gu's work.


src/article.ts
```diff
@@ -125,55 +127,68 @@ const tweetContent = (tweetId: string, doc: DocumentFragment): Result<string, No
...
 
-const parseImage = (element: BlockElement): Option<Image> => {
-    const masterAsset = element.assets.find(asset => asset.typeData.isMaster);
-    const { alt, caption, displayCredit, credit } = element.imageTypeData;
+const parseImage = (element: IBlockElement): Option<Image> => {
+    const masterAsset = element.assets.find(asset => asset?.typeData?.isMaster);
+    const { alt = "", caption = "", displayCredit = false, credit = "" } = element.imageTypeData || {};
 
-    return fromNullable(masterAsset).map(asset => ({
+    return new Some({
```
alexduf: I think you're losing the optionality here. It should be closer to the original file and return None if `masterAsset` is null.

src/article.ts
```diff
@@ -182,46 +197,58 @@ const parseElement =
...
-        commentable: content.fields.commentable,
+        commentable: content?.fields?.commentable ?? false,
         tags: content.tags,
     })
 
-const articleFieldsWithBody = (docParser: DocParser, content: Content): ArticleFieldsWithBody =>
-    ({
+const articleFieldsWithBody = (docParser: DocParser, content: Content): ArticleFieldsWithBody => {
+    const body = content?.blocks?.body;
+    const elements = body?.length ? body[0]?.elements : undefined;
```
alexduf: That's a bit hard to read. Any chance something like that would work? `body?[0].elements`

src/article.ts
```diff
@@ -182,46 +197,58 @@ const parseElement =
...
+    return ({
         ...articleFields(docParser, content),
-        body: parseElements(docParser)(content.blocks.body[0].elements),
+        body: parseElements(docParser)(elements),
     });
+}
 
 const fromCapi = (docParser: DocParser) => (content: Content): Article => {
-    switch (content.type) {
+    switch (content.type.toString()) {
```
alexduf: it feels like we should be able to use the type in the switch case? Is that not an enum?

webb04: This is the problem where thrift is an enum with numbers, but the CAPI JSON returns strings :/

alexduf: So toString returns 0, 1, 2 etc?

webb04: Actually it doesn't, but `content.type === ContentType.ARTICLE` doesn't work...

alexduf:
I think it looks good now, the only thing I'm worried about is how there's a lot of default values that wouldn't make sense if the code were to ever hit these cases.
We probably need to handle the optionality if these fields a bit better.

Any thoughts @JamieB-gu

src/renderer.ts
```diff
@@ -78,6 +77,7 @@ const Bullet = (props: { pillar: Pillar; text: string }): ReactElement =>
     h(Fragment, null,
         styledH('span', { css: bulletStyles(getPillarStyles(props.pillar).kicker) }, '•'),
         props.text.replace(/•/, ''),
+        h('br', null, null),
```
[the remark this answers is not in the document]
webb04: Bullets would be inline without this.

src/article.ts
webb04: I thought it was at first but not quite. It seems like this should be possible but it's not ``` content.type === ContentType.ARTICLE ```

src/article.ts
```diff
@@ -182,46 +196,58 @@ const parseElement =
...
-        publishDate: content.webPublicationDate,
+        pillar: pillarFromString(content?.pillarId),
+        headline: content?.fields?.headline ?? "",
+        standfirst: docParser(content?.fields?.standfirst ?? ""),
+        byline: content?.fields?.byline ?? "",
+        bylineHtml: fromNullable(content?.fields?.bylineHtml).map(docParser),
+        // eslint-disable-next-line
+        // @ts-ignore
+        // CAPI is sending us a string, even though the thrift definition is CapiDateTime
+        publishDate: content?.webPublicationDate,
```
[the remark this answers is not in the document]
webb04: Here is the culprit https://github.com/guardian/content-api-models/blob/master/models/src/main/thrift/content/v1.thrift#L1473.
I don't think we can cast this?

src/article.ts
JamieB-gu: I don't know if there's a nice way to do an optional lookup on array indices... I tried to hunt around for one but didn't have much luck.

Sadly, I think this will compile fine but error at runtime:

```ts
type Foo = {
    bar: {
        baz: number;
    }
}

const foos: Foo[] = [];

foos[0].bar.baz;
```

src/renderer.ts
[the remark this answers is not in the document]
JamieB-gu: Can we solve that by adding a `display: block` to the `bulletStyles`?

src/article.ts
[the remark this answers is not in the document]
JamieB-gu: I think we could do something like this:

```ts
typeof content?.webPublicationDate === 'string' ? content.webPublicationDate : ''
```

Although since we don't have a sensible fallback for this I think maybe we should make it an `Option` too:

```ts
typeof content?.webPublicationDate === 'string' ? new Some(content.webPublicationDate) : new None()
```

src/article.ts
twrichards: could you change the enum values to be strings? https://www.typescriptlang.org/docs/handbook/enums.html#string-enums

webb04: These types were generated from this thrift file https://github.com/guardian/content-api-models/blob/master/models/src/main/thrift/content/v1.thrift#L25.

We didn't want to change any of the generated files but it might make sense to make these strings?

src/renderer.ts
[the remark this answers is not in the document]
webb04: Tried that but then we'd have:

•
Some text
•
Some text

src/capi.ts
```diff
@@ -108,39 +110,47 @@ interface Contributor {
     bylineLargeImageUrl?: string;
 }
 
-const tagsOfType = (type_: string) => (tags: Tag[]): Tag[] =>
-    tags.filter((tag: Tag) => tag.type === type_);
+const tagsOfType = (type_: string) => (tags: ITag[]): ITag[] =>
+    tags.filter((tag: ITag) => tag.type.toString() === type_);
```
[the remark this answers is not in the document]
webb04: hmm same as the problems before I think, unfortunately this doesn't work:
 
``` const tagsOfType = (type_: TagType) => (tags: Tag[]): Tag[] =>
    tags.filter((tag: Tag) => tag.type === type_);```

```const articleContributors = (content: Content): Tag[] =>
    tagsOfType(TagType.CONTRIBUTOR)(content.tags); ```

File: src/article.ts (hunk 2 above)
Comment (JamieB-gu) [reply src/article.ts]: Perhaps the solution is a simple helper function:

```ts
const getIndex = (idx: number) => <A>(arr: A[]): Option<A> => {
    const value = arr[idx];
    return value !== undefined ? new Some(value) : new None();
  }
```

src/renderer.ts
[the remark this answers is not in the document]
JamieB-gu: Ah yeah, because those styles are on the `span`. In that case does the bullet actually make more semantic sense as a `<p>` tag?

```ts
const Bullet = (props: { pillar: Pillar; text: string }): ReactElement =>
    h('p', null,
        styledH('span', { css: bulletStyles(getPillarStyles(props.pillar).kicker) }, '•'),
        props.text.replace(/•/, ''),
    );
```

src/capi.ts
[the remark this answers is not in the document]
JamieB-gu: Ok, likely a similar problem regarding treating the types in the JSON response as the same as the types in the Thrift response. Will fix in future when we can parse Thrift, as discussed offline.

src/article.ts
alexduf: Then I guess it's not worth the effort. It's probably fine as it is

src/article.ts
```diff
@@ -125,10 +127,15 @@ const tweetContent = (tweetId: string, doc: DocumentFragment): Result<string, No
 }
 
 const parseImage = (element: BlockElement): Option<Image> => {
-    const masterAsset = element.assets.find(asset => asset.typeData.isMaster);
-    const { alt, caption, displayCredit, credit } = element.imageTypeData;
+    const masterAsset = element.assets.find(asset => asset?.typeData?.isMaster);
+    const { alt = "", caption = "", displayCredit = false, credit = "" } = element.imageTypeData ?? {};
 
-    return fromNullable(masterAsset).map(asset => ({
+    if (!masterAsset?.file || !masterAsset.typeData?.width || !masterAsset.typeData.height) {
```
JamieB-gu:
```suggestion
    if (!masterAsset?.file || !masterAsset.typeData?.width || !masterAsset.typeData?.height) {
```

src/article.ts
```diff
@@ -181,17 +205,36 @@ const parseElement =
...
+
+const capiDateTimeToDate = (date: CapiDateTime | undefined | string): Option<Date> => {
+    // Thrift definitions define some dates as CapiDateTime but CAPI returns strings
+    try {
+        if (date && typeof date === 'string') {
+            return new Some(new Date(date));
+        }
+        return new None();
+    } catch(e) {
+        console.error(`Unable to convert date from CAPI: ${e}`);
```
JamieB-gu: If we have an error message, should this return `Result<string, Date>`?

src/article.ts
```diff
@@ -200,31 +243,34 @@ const parseBlocks = (docParser: DocParser) => (blocks: Block[]): LiveBlock[] =>
...
-        standfirst: docParser(content.fields.standfirst),
-        byline: content.fields.byline,
-        bylineHtml: fromNullable(content.fields.bylineHtml).map(docParser),
-        publishDate: content.webPublicationDate,
+        pillar: pillarFromString(content?.pillarId),
+        headline: content?.fields?.headline ?? "",
+        standfirst: fromNullable(content?.fields?.standfirst).map(docParser),
+        byline: content?.fields?.byline ?? "",
+        bylineHtml: fromNullable(content?.fields?.bylineHtml).map(docParser),
+        publishDate: typeof content?.webPublicationDate === 'string' ? new Some(content.webPublicationDate) : new None(),
```
JamieB-gu: Would you be able to reuse `capiDateTimeToDate` here?

src/article.ts
```diff
@@ -235,7 +281,7 @@ const fromCapi = (docParser: DocParser) => (content: Content): Article => {
             } else if (isFeature(content)) {
                 return { layout: Layout.Feature, ...articleFieldsWithBody(docParser, content) };
 
-            } else if (isReview(content)) {
+            } else if (isReview(content) && fields?.starRating) {
```
JamieB-gu: As this is more or less what the `isReview` function is doing anyway, and I believe this is the only place `isReview` is used, perhaps we can drop it entirely and just do this:

```suggestion
            } else if ('starRating' in fields) {
```

webb04: isReview does have a check to see if the rating is `0 | 1 |2 | 3 | 4 |5` which if it isn't, falling back to standard articles might be what we want

src/article.ts
webb04: Will that get caught later?

PR #154, opened 2020-01-17
liveblog style fixes

webb04:
## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |


webb04:
> Thank you for tidying up the mess I made 🙂

Think we both added regressions because there was underlines too! The other tickets are connected to the new rendering (seems like some child components aren't rendering like pullquotes). I'll stay away from those as I imagine you're working on them too!

PR #155, opened 2020-01-17
Remove Deprecated Blocks Rendering System

JamieB-gu:
## Why are you doing this?

All the articles have been migrated over to the new parsing/rendering combination system. This means we can delete `renderBlocks` and some other related code (`image.ts`).

This PR also removes a number of uses of the CAPI types sprinkled across the codebase, and introduces the `ElementKind` type.

## Changes

- Deleted `renderBlocks`
- Deleted `image.ts`
- Split out the ImageElement type
- Tweaked the Image component and added separate Figure
- Added BodyImage
- Updated header images to use new ImageElement
- Added `ElementKind` to replace use of the CAPI `ElementType`


src/article.ts
```diff
@@ -58,31 +58,42 @@ type Article
...
     displayCredit: boolean;
     credit: string;
     file: string;
     width: number;
     height: number;
-} | {
-    kind: ElementType.PULLQUOTE;
+}
+
+type BodyElement = {
```
alexduf: While we're at it, should we split each `BodyElement` in separate types and have `type BodyElement= Image | Text | Interactive` etc?

webb04:
| without pixels | with pixels |
| --- | --- |
|  |  |

src/article.ts
JamieB-gu: That seems sensible, it may let us narrow the types of functions further down the call stack.

PR #157, opened 2020-01-20
remove fetch from mapi

webb04:
We'll be receiving a POST request rather than fetching ourselves? https://github.com/guardian/apps-rendering/pull/156

### DIFF
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -124,24 +124,6 @@ async function serveArticle(req: Request, res: ExpressResponse): Promise<void> {
   }
 }
 
-async function queryMapi(req: Request, res: ExpressResponse): Promise<void> {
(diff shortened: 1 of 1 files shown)


alexduf:
Should that remain an option for local development though? something that fetches the thrift and passes it to the other function that takes thrift to generate HTML?

webb04:
That's still possible in `serveArticle` to fetch JSON directly from CAPI.

I thought we would only accept POSTs (https://github.com/guardian/apps-rendering/pull/156) from mapi rather than apps-rendering requesting data?

alexduf:
We just had a chat, it's more clear. The CAPI fetch is still available, what's removed is the MAPI fetch.

JamieB-gu:
Does this mean we've lost the functionality added in #108?

PR #162, opened 2020-01-22
Render unordered lists

webb04:
## Why are you doing this?
UL elements weren't being rendered.

## Changes
- Moved more bullet styles over to `renderer.ts`

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -50,6 +50,31 @@ const Anchor = (props: { href: string; text: string; pillar: Pillar }): ReactEle
         props.text,
     );
 
+const listStyles = (colour: string): SerializedStyles => css`
+    list-style: none;
+    padding-left: 0;
+`
+
+const listItemStyles: SerializedStyles = css`
+    padding-left: 2rem;
+    line-height: 2.2rem;
+
+    &::before {
+        display: inline-block;
+        content: '';
+        border-radius: 0.5rem;
+        height: 1rem;
+        width: 1rem;
+        margin-right: 1rem;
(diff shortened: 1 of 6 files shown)


src/renderer.ts
```diff
@@ -50,6 +50,32 @@ const Anchor = (props: { href: string; text: string; pillar: Pillar }): ReactEle
...
+    &::before {
+        display: inline-block;
+        content: '';
+        border-radius: 0.5rem;
+        height: 1rem;
+        width: 1rem;
+        margin-right: 1rem;
+        background-color: ${palette.neutral[86]};
+        margin-left: -2rem;
+    }
```
JamieB-gu: I think if you use `list-style-position: inside` in `listStyles` then you will be able to drop the `padding-left: 2rem` and `margin-left: -2rem` in `listItemStyles`.

src/renderer.ts
```diff
@@ -50,6 +50,32 @@ const Anchor = (props: { href: string; text: string; pillar: Pillar }): ReactEle
...
+
+    > p:first-of-type {
+        display: inline;
+    }
```
JamieB-gu: Out of curiosity, what does this do?

webb04: `p` tags will be block and clear to the next line

src/renderer.ts
webb04: Ah that looked like an improvement but it's not working with out custom bullet points made from a pseudo ::Before

PR #166, opened 2020-01-23
Rename functor map to fmap

webb04:
## Why are you doing this?
Rename the functor `map` function to `fmap`. This lets us:
1. Easily distinguish between types, standard JavaScript arrays and `Option`, `Result` etc
2. Remove ESLint warnings and rule suppression comments throughout the codebase

Would be interested in your opinions on this... we're using functor `map` a lot more with the new CAPI types.


JamieB-gu:
FYI @regiskuckaertz @adamnfish

PR #169, opened 2020-01-24
Standfirst link styles and render elements inside a tags

webb04:
## Changes

- Remove old specific standfirst link styles (no underline opacity)
- Make standfirsts bold (current dotcom styles)
- Render elements inside of a tags

## Screenshots

| Before | After |
| --- | --- |
|  |  |


src/renderer.ts
```diff
@@ -150,7 +150,7 @@ const textElement = (pillar: Pillar) => (node: Node, key: number): ReactNode =>
         case 'SPAN':
             return text;
         case 'A':
-            return h(Anchor, { href: getHref(node).withDefault(''), text, pillar, key });
+            return h(Anchor, { href: getHref(node).withDefault(''), text, pillar, key }, Array.from(node.childNodes).map(textElement(pillar)));
```
alexduf: I think at this point it might be nice to extract `Array.from(node.childNodes).map(textElement(pillar))` into its own function!

JamieB-gu: Annoyingly this results in a circular dependency. This function relies on `textElement`, and `textElement` relies on this function. That's why the build's failing: use before definition 😬

alexduf: But wouldn't that be solve by using `function myFunction()` rather than `const myFunction` ?

JamieB-gu: I don't think so 🤔, I believe [this rule](https://eslint.org/docs/rules/no-use-before-define) applies to `function`s as well.

PR #170, opened 2020-01-27
Update src-foundations To 0.12.3

JamieB-gu:
## Why are you doing this?

To keep us up to date with the latest design system thinking.

**Note**: There will be a follow-up PR to migrate to functional colour names.

## Changes

- Bumped `src-foundations` to 0.12.3.
- Changed palette usage to handle breaking changes.
- Imported palette as individual module where possible to take advantage of tree shaking 🌳

### DIFF
--- a/src/components/immersive/byline.tsx
+++ b/src/components/immersive/byline.tsx
@@ -2,7 +2,7 @@
 
 import React from 'react';
 import { css, SerializedStyles } from '@emotion/core';
-import { palette } from '@guardian/src-foundations';
+import { neutral, background } from '@guardian/src-foundations/palette';
 
 import Follow from 'components/shared/follow';
 import { sidePadding, textSans, darkModeCss, basePx } from 'styles';
@@ -27,14 +27,14 @@ const Styles = ({ kicker }: PillarStyles): SerializedStyles => css`
 
         time {
             font-size: 1.4rem;
-            color: ${palette.neutral[46]};
+            color: ${neutral[46]};
         }
     }
 `;
 
 const DarkStyles = ({ inverted }: PillarStyles): SerializedStyles => darkModeCss`
-    background: ${palette.neutral.darkMode};
-    color: ${palette.neutral[86]};
+    background: ${background.inverse};
+    color: ${neutral[86]};
 
     .author {
         .follow, a {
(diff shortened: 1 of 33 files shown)


src/components/immersive/byline.tsx
```diff
@@ -27,22 +27,22 @@ const Styles = ({ kicker }: PillarStyles): SerializedStyles => css`
 
         time {
             font-size: 1.4rem;
-            color: ${palette.neutral[46]};
+            color: ${neutral[46]};
```
SiAdcock: You could also use `text.secondary`, which is a bit more semantic

src/components/immersive/byline.tsx
```diff
@@ -27,22 +27,22 @@ const Styles = ({ kicker }: PillarStyles): SerializedStyles => css`
...
+    color: ${neutral[86]};
 
     .author {
         .follow, a {
             color: ${inverted};
         }
 
         time {
-            color: ${palette.neutral[60]};
+            color: ${neutral[60]};
```
SiAdcock: I'm going to investigate adding the following tokens:

- neutral[86] -> text.inversePrimary`
- neutral[60] -> text.inverseSecondary`

Would this make sense, from your perspective?

src/components/liveblog/block.tsx
```diff
@@ -80,7 +80,7 @@ interface TitleProps {
 const Title = ({ title, highlighted }: TitleProps): JSX.Element | null => {
     const TitleStyles = css`
         padding: 0.5rem 0.125rem;
-        background-color: ${palette.yellow.main};
+        background-color: ${brandAlt[400]};
```
SiAdcock: I'd go with `brandAltBackground.primary` here

SiAdcock:
These changes look great! I've made a couple of suggestions regarding functional colours you could use, but I appreciate you want to keep changes in this PR to a minimum! Also it would be remarkable if the design system provided an ESLint plugin that made sure you're using the most semantic names for everything, but this is a way off yet 😄

src/components/immersive/byline.tsx
JamieB-gu: Ah, yes, I've actually made a note to talk to you about those two colours. If you're planning to create a semantic "dark mode" name for them that would be great!

src/components/immersive/byline.tsx
JamieB-gu: 👍 I probably should have mentioned in the description that I'm planning to do a second PR and migrate to functional colour names. I was hoping to talk to you first about the dark mode colours that you flagged below.

PR #173, opened 2020-01-31
Implement thrift apis

webb04:
## Why are you doing this?

Implement APIs needed for basic articles.

- Is user premium (show/hide ad slots) - *Called on page load*
- Follow topics - *Tapping follow link*
- Unfollow topics - *Tapping unfollow link*
- Is user following topic - *Called on page load*
- Launch slideshow on image tap - *Called on image tap*

We could hydrate some of these components client side and use react, but I think it's better to just add some eventListeners to the elements for the time being.


alexanderedge:
Thanks, I'll give this a test

src/client/article.ts
```diff
@@ -25,20 +26,106 @@ function getAdSlots(): AdSlot[] {
...
+        }
+    })
+}
+
+function launchSlideshow(src: string | null): void {
+    const images = Array.from(document.querySelectorAll('.launch-slideshow'));
+    const imagesWithCaptions: Image[] = images.flatMap((image: Element) => {
+        const url = image.getAttribute('src');
+        const caption =  image.getAttribute('caption') ?? undefined;
+        const credit = image.getAttribute('credit') ?? undefined;
```
JamieB-gu: Could these become [data attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*)?

PR #174, opened 2020-01-31
Mss 1413 more logs

alexduf:
## Why are you doing this?
Because it wasn't working in CODE

## Changes

- use `i` instead of `I` in the SSM configuration (Not in this PR, fixed in the SSM console)
- missing `/` at the end of the SSM path to authorise the loading of keys
- use `EC2MetadataCredentials` instead of `ECSCredentials`. `ECSCredentials` is for lambdas
- better error handling
- access log

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -125,7 +125,17 @@ async function serveArticle(req: Request, res: ExpressResponse): Promise<void> {
 // ----- App ----- //
 
 const app = express();
-app.use(bodyParser.raw({limit: '50mb'}))
+app.use(bodyParser.raw({limit: '50mb'}));
+
+app.all('*', (request, response, next) => {
+  const start = Date.now();
+  response.once('finish', () => {
+    const duration = Date.now() - start;
+    console.log(`HTTP ${request.method} ${request.path} returned ${response.statusCode} in ${duration}ms`)
+  });
+
+  next();
+});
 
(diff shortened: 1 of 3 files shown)


src/server/server.ts
```diff
@@ -125,7 +125,17 @@ async function serveArticle(req: Request, res: ExpressResponse): Promise<void> {
...
+app.use(bodyParser.raw({limit: '50mb'}));
+
+app.all('*', (request, response, next) => {
+  const start = Date.now();
+  response.once('finish', () => {
+    const duration = Date.now() - start;
+    console.log(`HTTP ${request.method} ${request.path} returned ${response.statusCode} in ${duration}ms`)
+  });
+
+  next();
```
webb04: This is interesting!

JamieB-gu:
FYI, there's an interesting part of the express docs covering [production best practices](https://expressjs.com/en/advanced/best-practice-performance.html). It includes [a section](https://expressjs.com/en/advanced/best-practice-performance.html#do-logging-correctly) recommending against using `console.log` in production.

alexduf:
Yes, we have a ticket to address that MSS-1170

We'll have to introduce a logging framework that is asynchronous and formats the logs correctly to be shipped to kinesis

PR #178, opened 2020-02-04
Receive thrift function calls from iOS and Android

webb04:
## Why are you doing this?
Some tasks will require the native layer to pass a message to the webview and will expect an action and/or response.

An example of this is inserting epics into the html.


alexanderedge:
Max raised a question related to this which I posted into the mobile-apps-thrift repo https://github.com/guardian/mobile-apps-thrift/issues/12.

src/native/thrift/nativeConnection.ts
```diff
@@ -123,14 +128,45 @@ export class NativeConnection<Context = void> extends ThriftConnection {
     }
 }
 
-function receiveiOSMessage(buffer: Buffer): void {
-    const transport = new BufferedTransport(buffer);
-    const protocol = new CompactProtocol(transport);
-    const processor = new Webview.Processor(new WebviewHandler);
-    processor.process(protocol, protocol, null);
+class ThriftServer<T> {
```
davidfurey: Should this be a different file?

src/native/thrift/nativeConnection.ts
```diff
@@ -123,14 +128,45 @@ export class NativeConnection<Context = void> extends ThriftConnection {
...
+    }
+
+    constructor(processor: ThriftProcessor<null, T>) {
+        this.processor = processor
+        window.receiveNativeMessage = this.receive;
+    }
 }
 
-window.receiveiOSMessage = receiveiOSMessage;
+new ThriftServer(new Webview.Processor(new WebviewHandler));
```
davidfurey: This should probably live somewhere else

src/native/thrift/nativeConnection.ts
```diff
@@ -21,15 +22,19 @@ import { WebviewHandler } from 'native/webviewApi';
 declare global {
     interface Window {
         nativeConnections: { [id: string]: NativeConnection };
-        AndroidWebViewMessage?: (nativeMessage: NativeMessage) => {};
+        AndroidWebViewRequest?: (nativeMessage: NativeMessage) => {};
```
davidfurey: What do you think about changing these names?
`AndroidWebViewRequest?`
`AndroidNativeResponse?`

`iOSWebViewRequest`
`iOSNativeResponse`

`receiveNativeRequest`

src/native/thrift/nativeConnection.ts
```diff
@@ -94,7 +89,17 @@ export class NativeConnection<Context = void> extends ThriftConnection {
         const message = this.outBuffer.shift();
         if (message) {
             console.log("Sending next message")
```
davidfurey: Should we remove this or change it to `console.debug`?

src/native/thrift/nativeConnection.ts
```diff
@@ -9,22 +9,32 @@ import {
...
-                iOSWebViewMessage: {
+                iOSWebViewRequest: {
+                    postMessage: (nativeMessage: NativeMessage) => {};
+                };
+                iOSWebViewResponse: {
                     postMessage: (nativeMessage: NativeMessage) => {};
                 };
             };
         };
+        receiveNativeMessage: (nativeMessage: NativeMessage) => void;
```
webb04: Should `NativeMessage` be renamed to `Message` as it's used both ways now. Maybe the `connectionId` should be optional or we have two separate types?

davidfurey: I think the connection id should always be required. Probably makes sense to rename to `Message` or maybe `ThriftMessage`?

src/native/thrift/nativeConnection.ts
webb04: We shouldn't send them to Prod but maybe we look at adding https://stackoverflow.com/questions/41040266/remove-console-logs-with-webpack-uglify

webb04:
Closing this PR as it makes sense to add the changes directly onto the new viewController.

This PR implemented a two way thrift communication:

native calling webview
webview calling native

It can be used as a reference with this [iOS](https://github.com/guardian/ios-live/pull/4940).

PR #180, opened 2020-02-05
code coverage

webb04:
## Why are you doing this?

It's part of our health OKR to have 75% test coverage on our core parsing logic.

-----

To run: `npm run test`

I've added `renderer.ts` but we can include more files with parsing logic.

The threshold is currently 15% so that the builds don't fail but we should increase this as we add tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,24 @@
     "storybook": "start-storybook -s ./",
     "deploy-storybook": "storybook-to-ghpages && ./.storybook/pushFonts.sh"
   },
+  "jest": {
+    "collectCoverage": true,
+    "transform": {
+      "^.+\\.ts$": "ts-jest"
+    },
+    "coverageThreshold": {
+      "global": {
+          "branches": 75,
+          "functions": 75,
+          "lines": 75,
+          "statements": 75
+      }
+  },
+  "coverageReporters": ["json", "lcov", "text", "clover"],
+    "collectCoverageFrom": [
+      "<rootDir>/src/renderer.ts"
+    ]
+  },
   "author": "",
   "license": "ISC",
   "dependencies": {
@@ -68,6 +86,7 @@
     "eslint": "^6.8.0",
     "eslint-plugin-react": "^7.18.0",
     "jest": "^24.9.0",
+    "ts-jest": "^25.2.0",
     "ts-loader": "^6.2.1",
     "webpack": "^4.41.5",
     "webpack-cli": "^3.3.10",
--- a/package.json
+++ b/package.json
@@ -9,8 +9,8 @@
   ],
   "scripts": {
     "upload": "node-riffraff-artifact",
-    "test:build": "webpack --config webpack.config.js --config-name tests",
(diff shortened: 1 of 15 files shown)


src/renderer.test.ts
```diff
@@ -0,0 +1,7 @@
+import { renderAll } from './renderer';
+
+describe('', () => {
+    test('', () => {
+        expect(renderAll("")).toBeDefined();
```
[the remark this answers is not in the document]
webb04: Yeah this test should be changed but I expected that to be recognised as covered code, but it's not... it's still reporting as 0% coverage and the `renderAll` line is displayed as `Uncovered`

package.json
```diff
@@ -20,6 +21,34 @@
...
+    "collectCoverage": true,
+    "transform": {
+      "^.+\\.tsx?$": "ts-jest"
+    },
+    "coverageThreshold": {
+      "global": {
+        "branches": 15,
+        "functions": 15,
+        "lines": 15,
+        "statements": 15
```
JamieB-gu: 15% of each?

package.json
```diff
@@ -20,6 +21,34 @@
...
+      }
+    },
+    "coverageReporters": [
+      "json",
+      "lcov",
+      "text",
+      "clover",
```
JamieB-gu: What's clover?

src/components/shared/keyline.test.tsx
```diff
@@ -6,24 +6,28 @@ import { Layout } from 'article';
 
 configure({ adapter: new Adapter() });
 
+const styles = (component: any): string => component.props().css
+    .map(({ styles }: any) => styles)
```
JamieB-gu: Do we have a stricter type for these?

src/renderer.test.ts
```diff
@@ -0,0 +1,8 @@
+import { makeCaption } from './renderer';
+
+describe('renderer returns expected content', () => {
+    test('formats captions', () => {
```
JamieB-gu: Should we be testing functions that are private to a module?

tsconfig.json
```diff
@@ -6,7 +6,7 @@
         // Import non-ES modules as default imports.
         "esModuleInterop": true,
         // Don't transpile jsx (handled by babel w/ emotion)
-        "jsx": "preserve",
+        "jsx": "react",
```
JamieB-gu: What does this mean?

package.json
webb04: Should be 75%+ eventually but I set it lower so the builds don't fail

tsconfig.json
webb04: This is to get it to work with `ts-jest`. Client and server configs still use `preserve`

package.json
webb04: https://istanbul.js.org/docs/advanced/alternative-reporters/

webb04: Removed some of them and kept `text` and `html` which you can see by running `open coverage/index.html`

PR #182, opened 2020-02-11
Fix Storybook and Add Standard Stories

JamieB-gu:
## Why are you doing this?

The headline stories were failing to build because the type of `Article` had changed (I think this happened in the transition from JSON to Thrift for CAPI types, possibly in in #168). This fixes that problem by correctly mocking the new article. It also fixes a problem where storybook was 404ing for some fonts due to an incorrect path.

In addition it adds some stories for the standard byline, and includes a new client-side parser derived from [`DOMParser`](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser), which we may use in the future for pages like liveblogs.

## Changes

- Fix `Article` mocking for headline stories
- Fix font loading for storybook
- Added Byline stories
- Added client-side parser

### DIFF
--- a/src/client/parser.ts
+++ b/src/client/parser.ts
@@ -1,0 +1,27 @@
+// ----- Imports ----- //
+
+import { Result, Err, Ok } from 'types/result';
+
+
+// ----- Functions ----- //
+
+const parse = (domParser: DOMParser) => (s: string): Result<string, DocumentFragment> => {
+
+    try {
+        const frag = new DocumentFragment();
+        const docNodes = domParser.parseFromString(s, 'text/html').body.childNodes;
+
+        Array.from(docNodes).forEach(node => frag.appendChild(node));
+        return new Ok(frag);
+    } catch (e) {
+        return new Err(`I wasn't able to parse the string into a DocumentFragment because: ${e}`);
+    }
+
+}
+
+
+// ----- Exports ----- //
+
+export {
+    parse,
+}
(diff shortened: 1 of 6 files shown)


src/client/parser.ts
```diff
@@ -0,0 +1,27 @@
+// ----- Imports ----- //
+
+import { Result, Err, Ok } from 'types/result';
+
+
+// ----- Functions ----- //
+
+const parse = (domParser: DOMParser) => (s: string): Result<string, DocumentFragment> => {
```
webb04: This is only needed to parse the html we get from capi right?

If that's the case, hopefully we won't need this client side... as all the liveblog blocks are text we control?

JamieB-gu: We do control them, but I believe we're currently planning to send them over the wire as a string of HTML (the one that CAPI provides)? If that's the case, we still need to parse them into a data structure in order to pull out what we need for rendering.

Alternatively we could parse them on the server, and send the result to the client as structured data for rendering.

webb04: Yeah I think doing that server side would be nicer. But this only applies to rehydrated content? so the `bylineHtml` in this example probably won't need rehydrating?

JamieB-gu: The byline won't, no.

For liveblog blocks I guess we may re-hydrate sections of the page to allow for changes to blocks, key events etc.

PR #183, opened 2020-02-11
bundlesize check

webb04:
Fail the builds if the client side javascript bundles become too large.


webb04:
> I suspect you've pushed the `dist/*` files by mistakes? Should that be added to the gitignore?
> 
> If they are added voluntarily, why?

They were added to check the bundle size in GitHub actions, but we should be able to generate the `dist` files in the CI. I'll give that another go

JamieB-gu:
I like this idea, but do we need to rely on a library to do it?

```js
const fs = require('fs');

const sizeLimit = 100000;
const bundle = 'GHGuardianHeadline-Bold.ttf';

if (fs.statSync(bundle).size > sizeLimit) {
    process.exit(1);
} else {
    process.exit(0);
}
```

webb04:
> I like this idea, but do we need to rely on a library to do it?
> 
> ```js
> const fs = require('fs');
> 
> const sizeLimit = 100000;
> const bundle = './dist/assets/article.js';
> 
> if (fs.statSync(bundle).size > sizeLimit) {
>     process.exit(1);
> } else {
>     process.exit(0);
> }
> ```

That's interesting... I think it does some extra things like gzipping, although we could do this ourselves too. The library had a nice reporter feature to GitHub but I was getting 500s when trying to use it.

webb04:
So after doing this it looks like `danger.js` might be better.... https://github.com/facebook/react/pull/14717

PR #184, opened 2020-02-12
link prefetch header

webb04:
## Why are you doing this?
Added a new `Link` http header to show what scripts and css files the apps need to prefetch. Hopefully the caching layers they use or the webview will do this automatically like browsers do. If not, the http header can still be used to manually fetch the resources so that they're available offline.

### Example:
App requests a standard article and receives the header: 
`Link: <assets/article.js>; rel=prefetch`
We should then assume this JavaScript file is available when the user opens the article (atleast from a home front) and is available offline

- We need to forward these headers in mapi to the apps
- This will change when we serve assets from a more suitable place like s3 (naming of files with hashes and the location)

### DIFF
--- a/src/components/shared/page.tsx
+++ b/src/components/shared/page.tsx
@@ -82,7 +82,7 @@ interface BodyProps {
     capi: Content;
 }
 
-interface elementWithScript {
+interface ElementWithScript {
     element: JSX.Element;
     script: string;
 }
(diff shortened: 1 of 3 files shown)


src/components/shared/page.tsx
```diff
@@ -82,13 +82,18 @@ interface BodyProps {
     capi: Content;
 }
 
+interface ElementWithScript {
```
alexduf: Should this be `ElementWithResources`, and have an array instead of one script?
That opens the door to prefetch more than one thing

src/components/shared/page.tsx
```diff
@@ -82,13 +82,18 @@ interface BodyProps {
...
+    element: JSX.Element;
```
JamieB-gu: `React.ReactElement`?

webb04: Will give that a try, it may of been a legacy type used.

PR #185, opened 2020-02-12
reduce rich link font size

webb04:
## Why are you doing this?
The Rich link font is too large, use a smaller header and reduce the link font size.

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -315,8 +315,9 @@ const richLinkStyles = css`
     background: ${neutral[97]};
     padding: ${basePx(1)};
 
-    h4 {
+    h1 {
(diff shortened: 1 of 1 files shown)


src/renderer.ts
```diff
@@ -315,7 +315,7 @@ const richLinkStyles = css`
     background: ${neutral[97]};
     padding: ${basePx(1)};
 
-    h1 {
+    h4 {
```
JamieB-gu: Why an `h4`? Are there `h1`, `h2` and `h3`s that sit above this in the hierarchy?

webb04: Ah yeah should be a h1, was using it for the styles rather than the semantics

PR #186, opened 2020-02-13
get local times on client

webb04:
## Why are you doing this?
Re-render dates client side to change to user's timezone.

## Changes

- Remove moment library
- Added client side code to detect times in the DOM and re-calculate them using `new Date()`. This should reflect their timezone

## More considerations
- We'll need the JavaScript available to avoid a weird change of date (this might be confusing)
- Maybe we should add the user's timezone after the date. A user recently questioned us about this and dotcom show it.

### DIFF
--- a/src/client/article.ts
+++ b/src/client/article.ts
@@ -129,9 +129,13 @@ function slideshow(): void {
 function formatDates(): void {
     Array.from(document.querySelectorAll('time[data-date]'))
         .forEach(time => {
-            const timestamp = time.getAttribute('data-date');
-            if (timestamp) {
-                time.textContent = formatDate(new Date(timestamp))
+            try {
+                const timestamp = time.getAttribute('data-date');
+                if (timestamp) {
+                    time.textContent = formatDate(new Date(timestamp))
+                }
+            } catch (e) {
+                console.error(e);
             }
         })
 }
--- a/src/client/article.ts
+++ b/src/client/article.ts
@@ -5,6 +5,7 @@ import { nativeClient } from 'native/nativeApi';
 import { AdSlot } from 'mobile-apps-thrift-typescript/AdSlot'
 import { Topic } from 'mobile-apps-thrift-typescript/Topic';
 import { Image } from 'mobile-apps-thrift-typescript/Image';
+import { formatDate } from 'date';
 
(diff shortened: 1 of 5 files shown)


src/client/article.ts
```diff
@@ -125,7 +126,18 @@ function slideshow(): void {
         }));
 }
 
+function formatDates(): void {
+    Array.from(document.querySelectorAll('time[data-date]'))
+        .forEach(time => {
+            const timestamp = time.getAttribute('data-date');
+            if (timestamp) {
+                time.textContent = formatDate(new Date(timestamp))
```
alexduf: you're quite likely to hit an exception when trying to parse a date from a string, I'd make sure this is handled gracefully, as one mis-formatted date would stop the whole page from rendering.

src/date.ts
```diff
@@ -65,5 +63,15 @@ export function makeRelativeDate(date: Date): string | null {
 }
 
 export function formatDate(date: Date): string {
-    return moment(date).format('HH:mm dddd, D MMMM YYYY');
+    const options = {
+        weekday: "long", year: "numeric", month: "long",
+        day: "numeric", hour: "numeric", minute: "numeric"
+    }
+    const dateParts = new Date(date)
+        .toLocaleTimeString("en-us", options)
```
alexduf: This doesn't quite feel right. Are you removing moment to save on the bundle size?
If so, passing an american locale, to then replace comas and split on spaces feels very fiddly!

Any chance javascript comes with the equivalent of this? https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html

webb04: Yes, I don't think there's any native equivalent in browsers.

Just found this, so maybe it's worth using a smaller library https://github.com/you-dont-need/You-Dont-Need-Momentjs

webb04:
I found a useful way to test this: 
`TZ='US/Pacific' open -na "Google Chrome" --args "--user-data-dir=$HOME/chrome-profile"`

src/date.ts
webb04: Updated to use `dayjs` which has the same `moment` api but is only 2KB https://github.com/iamkun/dayjs

PR #187, opened 2020-02-13
update pm2 log output directory

webb04:
Specify correct log output directory so that stream can read logs and push them to ELK.

https://pm2.keymetrics.io/docs/usage/log-management/

### DIFF
--- a/config/cloudformation.yaml
+++ b/config/cloudformation.yaml
@@ -309,8 +309,13 @@ Resources:
           chown -R ${App}:mapi /var/log/${App}
           export Stack=${Stack}
           export Stage=${Stage}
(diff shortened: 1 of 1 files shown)


config/cloudformation.yaml
```diff
@@ -309,7 +309,7 @@ Resources:
           chown -R ${App}:mapi /var/log/${App}
           export Stack=${Stack}
           export Stage=${Stage}
-          /usr/local/node/pm2 start --uid ${App} --gid mapi /opt/${App}/server.js
+          /usr/local/node/pm2 start --uid ${App} --gid mapi --log /var/log/ /opt/${App}/server.js
```
alexduf: Almost, I think it should be 
```suggestion
          /usr/local/node/pm2 start --uid ${App} --gid mapi --log /var/log/${App} /opt/${App}/server.js
```

PR #189, opened 2020-02-14
Format and Item Types

JamieB-gu:
## Why are you doing this?

Part of the motivation for this work is to standardise how we define the design of content. The concept of a [`DesignType`](https://github.com/guardian/content-api-scala-client/blob/9e249bcef47cc048da483b3453c10dd7d2e9565d/client/src/main/scala/com.gu.contentapi.client/utils/DesignType.scala) has existed for a while, and is used in frontend. We're using a [`Layout`](https://github.com/guardian/apps-rendering/blob/66f93cf1832e7cb1d6f26d8af4579edcb653216b/src/article.ts#L14) type in apps-rendering. These both attempt to solve a similar problem, but are incomplete and are not compatible with one another.

Therefore dotcom-rendering and apps-rendering are attempting to co-ordinate their approaches by introducing a new type (terminology may change):

```ts
interface Format {
    pillar: Pillar;
    design: Design;
    display: Display;
}
```

The idea is that this type *should* cover all possible rendering formats that a piece of content can have. Some related work has [recently been merged](https://github.com/guardian/dotcom-rendering/pull/1160) on dotcom-rendering with this goal in mind.

This PR adopts some of the techniques for identifying content that are used by [frontend](https://github.com/guardian/frontend/blob/e25e3f1bad2750c5fd45933a6b4e6ff34deffdd8/common/app/model/meta.scala#L580) and the [CAPI Scala Client](https://github.com/guardian/content-api-scala-client/blob/9e249bcef47cc048da483b3453c10dd7d2e9565d/client/src/main/scala/com.gu.contentapi.client/utils/CapiModelEnrichment.scala), such as identifying `Design` based on tags, and showcase based on the main media `role`.

With these changes I have also renamed the `Article` type to `Item` because

1. the name "article" was becoming somewhat overloaded (for example, it clashed with `Design.Article`), and
2. this is better correlated with MAPI's `mobile-item` service.

**Note:** I haven't made many changes to components and the general file structure to bring them in line with the new `Format` type (hence, for example, why we still have the term `opinion` in a few places where we really mean `Comment`). Updating this will require further thought and will likely come about in future PRs.

## Changes

- Renamed the `Article` type to `Item`
- Created new `Format` type, comprised of `Design`, `Pillar` and `Display`
- Brings apps-rendering more in line with frontend and dotcom-rendering
- Updated some naming across codebase to use `item` over `article`
- Split `Layout` into `Design` and `Display`
- Derived `Design` from tags
- Based decision for "showcase" on main media roles
- Turned off max line length for URLs
- Added `elements` to CAPI request (needed for showcase embeds)

### DIFF
--- a/src/capi.ts
+++ b/src/capi.ts
@@ -109,6 +109,7 @@ const capiEndpoint = (articleId: string, key: string): string => {
       'show-fields': 'all',
       'show-tags': 'all',
       'show-blocks': 'all',
+      'show-elements': 'all',
     })
   
     return `https://content.guardianapis.com/${articleId}?${params.toString()}`;
--- a/config/.eslintrc.js
+++ b/config/.eslintrc.js
@@ -17,7 +17,12 @@ module.exports =  {
         "@typescript-eslint/no-unused-vars": "off",
         'max-len': [
             "error",
-            { code: 100, ignoreStrings: true, ignoreTemplateLiterals: true },
+            {
+                code: 100,
+                ignoreStrings: true,
+                ignoreTemplateLiterals: true,
+                ignoreUrls: true,
+            },
         ],
         '@typescript-eslint/explicit-function-return-type': 2,
         'eqeqeq': 2,
--- a/src/ads.test.ts
+++ b/src/ads.test.ts
@@ -4,7 +4,7 @@ import { renderAll } from 'renderer';
 import { JSDOM } from 'jsdom';
 import { Pillar } from 'pillar';
 import { compose } from 'lib';
-import { ElementKind, BodyElement } from 'article';
+import { ElementKind, BodyElement } from 'item';
 
 const textElement = (nodes: string[]): BodyElement =>
     ({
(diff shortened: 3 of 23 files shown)


JamieB-gu:
FYI, the build is failing because some URLs in comments are too long for the linter's max line length. I think I might disable that rule for comments.

src/capi.ts
```diff
@@ -109,6 +109,7 @@ const capiEndpoint = (articleId: string, key: string): string => {
       'show-fields': 'all',
       'show-tags': 'all',
       'show-blocks': 'all',
+      'show-elements': 'all',
```
webb04: Do we need all or just one field?

JamieB-gu: Sadly `elements` is [an array](https://github.com/guardian/apps-rendering/pull/189/files#diff-ebb6d8b623f26b27144c95b4ab728dddR305) 😔

PR #192, opened 2020-02-17
captions can be html

webb04:
## Why are you doing this?

http://localhost:8080/food/2020/jan/31/bong-bongs-manila-kanteen-london-e2-grace-dent-restaurant-review

It's possible to receive html in image captions. Instead of always returning a string, we should parse the content and render the correct tags.

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -4,7 +4,6 @@ import { ReactNode, createElement as h, ReactElement } from 'react';
 import { css, jsx as styledH, SerializedStyles } from '@emotion/core';
 import { from, until } from '@guardian/src-foundations/mq';
 import { neutral } from '@guardian/src-foundations/palette';
-import { JSDOM } from 'jsdom';
 
 import { Option, fromNullable, Some, None } from 'types/option';
 import { srcset, src } from 'image';
@@ -175,12 +174,6 @@ const textElement = (pillar: Pillar) => (node: Node, key: number): ReactNode => 
 const text = (doc: DocumentFragment, pillar: Pillar): ReactNode[] =>
     Array.from(doc.childNodes).map(textElement(pillar));
 
-const makeCaption = (props: FigureElement): ReactNode[] => {
-    const { caption, credit, displayCredit, pillar } = props;
-    const fullCaption = displayCredit ? `${caption} ${credit}` : caption;
-    return text(JSDOM.fragment(fullCaption), pillar);
-}
-
 interface ImageProps {
     url: string;
     alt: string;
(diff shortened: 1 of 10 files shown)


src/renderer.ts
```diff
@@ -4,6 +4,7 @@ import { ReactNode, createElement as h, ReactElement } from 'react';
 import { css, jsx as styledH, SerializedStyles } from '@emotion/core';
 import { from, until } from '@guardian/src-foundations/mq';
 import { neutral } from '@guardian/src-foundations/palette';
+import { JSDOM } from 'jsdom';
```
JamieB-gu: Unfortunately, importing `JSDOM` here will cause the rendering code to crash if called on the client (because `JSDOM` only exists server-side). The most straightforward way to solve this problem is to pass down the parser we're currently using (client-side or server-side) as an injected dependency. We're doing this [here](https://github.com/guardian/apps-rendering/blob/d71214098c4697e8de794c8506e4e376b493ca04/src/item.ts#L192) in the parsing code.

On the topic of parsing, I believe by the time we get to the rendering code all HTML should have been parsed out into a data structure (`DocumentFragment`) 🤔. We can then extract information from it using the DOM APIs. Therefore, I think the best place to introduce this is probably in [`parseImage`](https://github.com/guardian/apps-rendering/blob/d71214098c4697e8de794c8506e4e376b493ca04/src/item.ts#L170) within `item.ts`:

```ts
const parseImage = (docParser: DocParser) => (element: BlockElement): Option<Image> => {
  // Some code
  return new Some({
    // Some properties
    caption: docParser(caption),
  });
}
```

while updating the corresponding [`Image` type](https://github.com/guardian/apps-rendering/blob/d71214098c4697e8de794c8506e4e376b493ca04/src/item.ts#L74) to reflect that fact that `caption`s can be more than just a string:

```ts
type Image = {
    kind: ElementKind.Image;
    alt: string;
    caption: DocumentFragment;
    displayCredit: boolean;
    credit: string;
    file: string;
    width: number;
    height: number;
}
```

webb04:
Moved the parsing higher up. Had to pass down a document fragement and a string because of the thrift communication to the native layers where they display the images.

JamieB-gu:
> the thrift communication to the native layers where they display the images

I'm not sure I understand, would you be able to elaborate?

webb04:
> > the thrift communication to the native layers where they display the images
> 
> I'm not sure I understand, would you be able to elaborate?

We need to pass the caption to iOS/Android (html string). I couldn't find a way to go from DocumentFragment to html string.. so I've passed both a string and a DocumentFragment

webb04:
[nvm us](url)

JamieB-gu:
> We need to pass the caption to iOS/Android (html string)

Ah, I see, this is from #173. Ok, makes sense, thanks.

PR #196, opened 2020-02-18
Production logs

alexduf:
## Why are you doing this?

- To get logs in our ELK stack

## Changes

- Set environment variables at the time PM2 starts
- Productionise the js bundle
- Create a logger with three implementation: clientDev, clientProd and server
- Swap the implementation based on which bundle we're building
- Use the new logger in all reasonable places instead of console.log

image

### DIFF
--- a/src/logger/index.ts
+++ b/src/logger/index.ts
@@ -1,0 +1,5 @@
+import {logger as consoleLogger} from "logger/clientDev";
+import {ILogger} from "./ILogger";
+
+// this is only here to please IntelliJ
+export const logger: ILogger = consoleLogger;
--- a/src/logger/index.ts
+++ b/src/logger/index.ts
@@ -1,5 +1,5 @@
 import {logger as consoleLogger} from "logger/clientDev";
-import {ILogger} from "./ILogger";
+import {LoggerFunctions} from "./LoggerFunctions";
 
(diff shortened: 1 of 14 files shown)


src/logger/index.ts
```diff
@@ -0,0 +1,5 @@
+import {logger as consoleLogger} from "logger/clientDev";
```
webb04: should this be 
```suggest
import {logger as consoleLogger} from "./clientDev";
```

src/logger/index.ts
```diff
@@ -0,0 +1,5 @@
...
+import {LoggerFunctions} from "./LoggerFunctions";
+
+// this is only here to please IntelliJ
```
webb04: Can confirm, VScode is also pleased with this line

webpack.config.js
```diff
@@ -26,54 +26,38 @@ class LaunchServerPlugin {
...
+    if (env && env.watch) {
+        plugins.push(new LaunchServerPlugin());
+    }
+
+    const mode = (env && env.production) ? "production" : "development";
+
     return {
         name: 'server',
-        mode: 'development',
+        mode: mode,
```
webb04: can shorthand to `mode,`

PR #200, opened 2020-02-19
update adverts

webb04:
## Why are you doing this?

Ads move, we should tell the native layers when they do.


webb04:
I've already opened this PR....

PR #201, opened 2020-02-19
don't insert ad placeholders if disabled in capi

webb04:
## Why are you doing this?

There's a flag set in capi to hide ads, we should make sure we do.

### DIFF
--- a/src/ads.test.ts
+++ b/src/ads.test.ts
@@ -1,4 +1,4 @@
-import { insertAdPlaceholdersWhenEnabled } from './ads';
+import { getAdPlaceholderInserter } from './ads';
 import { ReactNode } from 'react';
 import { renderAll } from 'renderer';
 import { JSDOM } from 'jsdom';
@@ -6,7 +6,8 @@ import { Pillar } from 'pillar';
 import { compose } from 'lib';
 import { ElementKind, BodyElement } from 'item';
 
-const insertAdPlaceholders = insertAdPlaceholdersWhenEnabled(false);
+const shouldHideAdverts = false;
+const insertAdPlaceholders = getAdPlaceholderInserter(shouldHideAdverts);
 
 const textElement = (nodes: string[]): BodyElement =>
     ({
--- a/src/ads.test.ts
+++ b/src/ads.test.ts
@@ -6,7 +6,7 @@ import { Pillar } from 'pillar';
 import { compose } from 'lib';
 import { ElementKind, BodyElement } from 'item';
 
-const insertAdPlaceholders = insertAdPlaceholdersWhenEnabled(true);
+const insertAdPlaceholders = insertAdPlaceholdersWhenEnabled(false);
 
 const textElement = (nodes: string[]): BodyElement =>
     ({
--- a/src/ads.test.ts
+++ b/src/ads.test.ts
@@ -1,10 +1,12 @@
-import { insertAdPlaceholders } from './ads';
+import { insertAdPlaceholdersWhenEnabled } from './ads';
 import { ReactNode } from 'react';
 import { renderAll } from 'renderer';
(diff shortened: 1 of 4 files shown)


src/client/article.ts
```diff
@@ -30,20 +35,21 @@ function getAdSlots(): AdSlot[] {
...
-    const targetNode = document.querySelector('body') as Node;
-    const config = { attributes: true, childList: true, subtree: true };
-    const callback = function(): void {
-        const currentAdSlots = getAdSlots();
-        if (JSON.stringify(adSlots) !== JSON.stringify(currentAdSlots)) {
-            adSlots = currentAdSlots;
-            nativeClient.updateAdverts(currentAdSlots);
-        }
-    };
+    if (adSlots) {
```
JamieB-gu:
```suggestion
    if (adSlots.length > 0) {
```

src/ads.ts
```diff
@@ -34,4 +34,10 @@ function insertAdPlaceholders(reactNodes: ReactNode[]): ReactNode[] {
         .pop();
 }
 
-export { insertAdPlaceholders }
+const insertAdPlaceholdersWhenEnabled =
+    (shouldHideAdverts: boolean | undefined): (reactNodes: ReactNode[]) => ReactNode[] =>
```
JamieB-gu:
```suggestion
    (shouldHideAdverts: boolean): (reactNodes: ReactNode[]) => ReactNode[] =>
```

src/components/shared/page.tsx
```diff
@@ -94,6 +94,9 @@ const WithScript = (props: { src: string; children: ReactNode }): ReactElement =
     </>
 
 function ArticleBody({ capi, imageSalt }: BodyProps): ElementWithResources {
+
+    const insertAdPlaceholders = insertAdPlaceholdersWhenEnabled(capi.fields?.shouldHideAdverts);
```
JamieB-gu:
```suggestion
    const insertAdPlaceholders = insertAdPlaceholdersWhenEnabled(capi.fields?.shouldHideAdverts ?? false);
```

src/ads.test.ts
```diff
@@ -1,11 +1,13 @@
-import { insertAdPlaceholders } from './ads';
+import { insertAdPlaceholdersWhenEnabled } from './ads';
 import { ReactNode } from 'react';
 import { renderAll } from 'renderer';
 import { JSDOM } from 'jsdom';
 import { Pillar } from 'pillar';
 import { compose } from 'lib';
 import { ElementKind, BodyElement } from 'item';
 
+const insertAdPlaceholders = insertAdPlaceholdersWhenEnabled(false);
```
JamieB-gu: This might just be me, but I'm reading this as "ad placeholders are not enabled because `false` has been passed". However, as far as I understand it, passing `false` will guarantee that they _are_ enabled.

Would it make sense to flip this boolean inside the function, or tweak this function name?

webb04: yeah it caught me out too https://github.com/guardian/apps-rendering/pull/201/commits/902fd782682faaf25c76751771816fb689617ddc

I will change

src/components/shared/page.tsx
```diff
@@ -94,6 +94,9 @@ const WithScript = (props: { src: string; children: ReactNode }): ReactElement =
     </>
 
 function ArticleBody({ capi, imageSalt }: BodyProps): ElementWithResources {
+
+    const insertAdPlaceholders = insertAdPlaceholdersWhenEnabled(capi.fields?.shouldHideAdverts ?? false);
```
JamieB-gu:
```suggestion
    const insertAdPlaceholders =
      insertAdPlaceholdersWhenEnabled(capi.fields?.shouldHideAdverts ?? false);
```

PR #202, opened 2020-02-19
Universal Headline Component

JamieB-gu:
## Why are you doing this?

This starts the process of updating components to make use of the `Format` type introduced in #189. It also adds several stories to view these components in storybook.

I've also implemented an alternate way of handling analysis underlines (see screenshots). I may revert this pending discussions with design.

There's likely to be some duplication in the stories files that can be abstracted out. I'll wait until I've written some more to determine how that will look.

cc @SiAdcock (Started using the design system typography and spacing)

**Note:** This doesn't yet cover liveblogs, I'm going to look at those next.

## Changes

- New Headline component to handle all types of content
- New StarRating component to handle review stars
- New stories to cover Headline and StarRating
- Added new headline component to standard article
- Added some helpers to styles
- Made use of design system typography and spacing

## Screenshots

analysis


src/components/headline.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+const pillarOptions = {
+    News: Pillar.news,
+    Opinion: Pillar.opinion,
+    Sport: Pillar.sport,
+    Culture: Pillar.arts,
+    Lifestyle: Pillar.lifestyle,
+};
+
+const starRating: Record<number, number> = [1, 2, 3, 4, 5];
```
webb04: I think we should be able to handle 0 (even if there hasn't been any in the last 200 reviews)

src/components/starRating.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+        webTitle: '',
+        webUrl: '',
+        apiUrl: '',
+        references: [],
+    },
+    commentable: false,
+    tags: [],
+};
+
+const starRating: Record<number, number> = [1, 2, 3, 4, 5];
```
webb04: Same here as above.

src/components/headline.stories.tsx
JamieB-gu: 👍 updated

Is it possible for a review to be given 0 stars?

webb04: It might not be (I overheard someone saying there hasn't been in the last 200 reviews), but I think it could possibly change one day, so we might as well handle it?

src/components/starRating.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+
+// ----- Component ----- //
+
+interface Props {
+    item: Item;
+}
+
+const StarRating = ({ item }: Props): ReactElement =>
+    item.design === Design.Review ? <div>{stars(item.starRating)}</div> : <></>;
```
webb04: ah nice way to avoid `ReactElement | null`

src/styles.ts
```diff
@@ -15,6 +23,16 @@ export const headlineFontStyles = css`
     margin: 0;
 `;
 
+export const textPadding = css`
+    padding-left: ${spaceToRem(2)};
+    padding-right: ${spaceToRem(2)};
```
webb04: how much is one space? we've got `basePx` of 8px throughout the codebase.

JamieB-gu: A unitless `4`

webb04: great, we should:
```ts const basePx = spaceToRem * 2;```

File: src/components/starRating.tsx [hunk 5]
```diff
@@ -0,0 +1,52 @@
...
+
+import { Item, Design } from 'item';
+import { icons, spaceToRem } from 'styles';
+
+
+// ----- Subcomponents ----- //
+
+const starStyles = css`
+    ${icons}
+    background-color: ${brandAlt[400]};
```
SiAdcock: I think you could also go with `brandAltBackground.primary`, which is more semantic

PR #205, opened 2020-02-21
Use multiplexed thrift

davidfurey:
With service name to identify the desired service

## Why are you doing this?

Thrift has a meta-protocol that supports serving multiple services across a single link.  Adding support for it will allow us to split the interface that the native layer provide into logically separate services without needing to create a different webview<->native postMessage connection.  

The Thrift spec allows an arbitrary string to identify a service, but we are choosing to use the service name as defined in the `thrift` file.

### DIFF
--- a/src/native/thrift/nativeConnection.ts
+++ b/src/native/thrift/nativeConnection.ts
@@ -132,5 +132,6 @@ export function createAppClient<TClient extends ThriftClient<void>>(
             super(new Protocol(transport), ServiceClient.serviceName ?? "")
         }
     }
-    return new ServiceClient(new NativeConnection(getTransport(transport), NamedMultiplexedProtocol));
+    const connection = new NativeConnection(getTransport(transport), NamedMultiplexedProtocol)
+    return new ServiceClient(connection)
 }
--- a/src/native/thrift/nativeConnection.ts
+++ b/src/native/thrift/nativeConnection.ts
@@ -6,6 +6,7 @@ import {
     getProtocol,
     TransportType,
     ProtocolType,
+    TTransport,
     ITransportConstructor,
     IProtocolConstructor,
     TApplicationException,
@@ -13,6 +14,8 @@ import {
 } from '@creditkarma/thrift-server-core'
 
 import * as uuid from 'uuid';
+
+import { TMultiplexedProtocol } from './protocols'
 
 declare global {
     interface Window {
@@ -123,5 +126,11 @@ export function createAppClient<TClient extends ThriftClient<void>>(
     transport: TransportType = 'buffered', 
     protocol: ProtocolType = 'compact'
 ): TClient {
-    return new ServiceClient(new NativeConnection(getTransport(transport), getProtocol(protocol)));
+    class NamedMultiplexedProtocol extends TMultiplexedProtocol {
+        constructor(transport: TTransport) {
+            const Protocol = getProtocol(protocol)
+            super(new Protocol(transport), ServiceClient.serviceName ?? "")
+        }
+    }
+    return new ServiceClient(new NativeConnection(getTransport(transport), NamedMultiplexedProtocol));
 }
--- a/src/native/thrift/protocols.ts
+++ b/src/native/thrift/protocols.ts
@@ -165,7 +165,7 @@ export class TMultiplexedProtocol extends TProtocolDecorator {
     }
   
     writeMessageBegin(name: string, type: MessageType, seqid: number): void {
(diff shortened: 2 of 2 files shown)


src/native/thrift/protocols.ts
```diff
@@ -0,0 +1,178 @@
...
+    TType,
+    MessageType,
+    IThriftMessage,
+    IThriftStruct,
+    IThriftMap,
+    IThriftField,
+    IThriftList
+} from '@creditkarma/thrift-server-core'
+
+export abstract class TProtocolDecorator extends TProtocol {
```
davidfurey: Based on https://github.com/apache/thrift/blob/master/lib/java/src/org/apache/thrift/protocol/TProtocolDecorator.java

src/native/thrift/protocols.ts
```diff
@@ -0,0 +1,178 @@
...
+    }
+    readString(): string {
+        return this.concreteProtocol.readString();
+    }
+    skip(type: TType): void {
+        return this.concreteProtocol.skip(type);
+    }
+}
+
+export class TMultiplexedProtocol extends TProtocolDecorator {
```
davidfurey: Based on https://github.com/apache/thrift/blob/master/lib/java/src/org/apache/thrift/protocol/TMultiplexedProtocol.java

src/native/thrift/protocols.ts
```diff
@@ -0,0 +1,178 @@
...
+    static readonly separator = ":"
+    readonly serviceName: string
+    
+    constructor(protocol: TProtocol, serviceName: string) {
+        super(protocol);
+        this.serviceName = serviceName;
+    }
+  
+    writeMessageBegin(name: string, type: MessageType, seqid: number): void {
+        if (type == MessageType.CALL || type == MessageType.ONEWAY) {
```
JamieB-gu:
```suggestion
        if (type === MessageType.CALL || type === MessageType.ONEWAY) {
```

should fix the lint errors.

src/native/thrift/protocols.ts
JamieB-gu: What's the thinking behind the intermediate `TProtocolDecorator` vs extending `TProtocol` directly?

davidfurey: Discussed in person, but in summary: the ProtocolDecorator is necessary so that TMultiplexedProtocol can wrap any non-abstract TProtocol.

src/native/thrift/nativeConnection.ts
```diff
@@ -123,5 +126,12 @@ export function createAppClient<TClient extends ThriftClient<void>>(
     transport: TransportType = 'buffered', 
     protocol: ProtocolType = 'compact'
 ): TClient {
-    return new ServiceClient(new NativeConnection(getTransport(transport), getProtocol(protocol)));
+    class NamedMultiplexedProtocol extends TMultiplexedProtocol {
```
webb04: If it's now multuplexed, should `createAppClient ` be renamed? I'm assuming we'll be making two, one for `webview.thrift` and for `native.thrift`?

davidfurey: We won't create one for `webview.thrift` and `native.thrift`.  We might split `native.thrift` out into
```
service Native {
    i32 nativeThriftPackageVersion()
}

service Commercial {
    void insertAdverts(1:list<AdSlot> adSlots),
    void updateAdverts(1:list<AdSlot> adSlots)
}

service Acquistions {
    void launchFrictionScreen(),
    MaybeEpic getEpics(),
    void epicSeen()
}

service Notifications {
    void follow(1:Topic topic),
    void unfollow(1:Topic topic),
    bool isFollowing(1:Topic topic),
}

service User {
    bool isPremium()
}

service ??? {
    void launchSlideshow(1:list<Image> images, 2:i32 selectedIndex)
}
```

All of these would be app clients, that could be created by, for example:
`createAppClient<Acquistions.Client>(Acquistions.Client, 'buffered', 'compact');`

Although maybe it would make sense to just be `createClient`?

webb04: I see, that will be useful as the number of APIs grow. `createAppClient` sounds ok knowing that, as it shows it's calling into native code (although that might not be to everyone)

webb04: I was reading `multiplex` as being able to call into native and having native use the same code to call into the webview

davidfurey: Ah, I see there is scope for confusion there.

PR #208, opened 2020-02-24
Hashed JavaScript bundles

webb04:
## Why are you doing this?

Use `contenthash` to be able to invalidate cached JavaScript bundles on device.
This does require client content being built before the server is started...

## Changes

- Adds content hash to generated JavaScript files
- Creates json mapping from script to hashed script name
- Reads mapping on server launch and passes it into the rendering logic

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,12 +15,13 @@
     "lint": "node_modules/eslint/bin/eslint.js -c config/.eslintrc.js ./ --ext '.ts,.tsx'",
     "watch:server": "webpack --config webpack.config.js --config-name server --env.watch",
     "watch:client": "webpack-dev-server --config webpack.config.js --config-name client",
-    "watch": "npm run watch:server & npm run watch:client",
+    "watch": "npm run build:client; npm run watch:server; npm run watch:client",
     "build:server": "webpack --config webpack.config.js --config-name server --env.production",
     "build:client": "webpack --config webpack.config.js --config-name client",
     "build:client:prod": "webpack --config webpack.config.js --config-name clientProduction",
     "storybook": "start-storybook -s ./",
-    "deploy-storybook": "storybook-to-ghpages && ./.storybook/pushFonts.sh"
+    "deploy-storybook": "storybook-to-ghpages && ./.storybook/pushFonts.sh",
+    "prod": "npm run build:client:prod; npm run build:server"
   },
   "jest": {
     "collectCoverage": true,
--- a/package.json
+++ b/package.json
@@ -16,12 +16,12 @@
     "watch:server": "webpack --config webpack.config.js --config-name server --env.watch",
     "watch:client": "webpack-dev-server --config webpack.config.js --config-name client",
     "watch": "npm run watch:client & npm run watch:server",
-    "build:server": "webpack --config webpack.config.js --config-name server --env.production",
+    "build:server": "webpack --config webpack.config.js --config-name server",
     "build:client": "webpack --config webpack.config.js --config-name client",
+    "build:server:prod": "npm run build:client:prod; webpack --config webpack.config.js --config-name server --env.production",
     "build:client:prod": "webpack --config webpack.config.js --config-name clientProduction",
     "storybook": "start-storybook -s ./",
-    "deploy-storybook": "storybook-to-ghpages && ./.storybook/pushFonts.sh",
-    "prod": "npm run build:client:prod; npm run build:server"
+    "deploy-storybook": "storybook-to-ghpages && ./.storybook/pushFonts.sh"
   },
   "jest": {
     "collectCoverage": true,
--- a/package.json
+++ b/package.json
@@ -100,6 +100,7 @@
     "ts-jest": "^25.2.0",
     "ts-loader": "^6.2.1",
     "webpack": "^4.41.5",
+    "webpack-assets-manifest": "^3.1.1",
     "webpack-cli": "^3.3.10",
     "webpack-dev-server": "^3.10.1",
     "winston": "^3.2.1",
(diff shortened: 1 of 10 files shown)


src/server/server.ts
```diff
@@ -131,9 +133,15 @@ app.get('/*', bodyParser.raw(), serveArticle);
...
 
 const port = 3040;
-app.listen(port, () => {
-  logger.info(`Server listening on port ${port}!`);
-  if (process.env.NODE_ENV !== "production") {
-    logger.info(`Webpack dev server is listening on port 8080`);
-  }
-});
+
+fs.readFile('dist/assets/manifest.json', function(err, data) {
```
alexduf: Can we package the file within the server.js so we don't have to ship it? I seem to remember there's a way to convince webpack to import these

JamieB-gu: Are you referring to [this](https://webpack.js.org/loaders/json-loader/)? Based on the note there it looks like it's now built-in.

webb04: I think you can just do this in Node 
`const json = require('./manifest.json');`

Which I'll swap to. I initially didn't because it's synchronous, but the server will need it before starting anyway.

webb04: node requires `--resolveJsonModule` so will try the webpack loader

package.json
```diff
@@ -100,6 +100,7 @@
     "ts-jest": "^25.2.0",
     "ts-loader": "^6.2.1",
     "webpack": "^4.41.5",
+    "webpack-assets-manifest": "^3.1.1",
```
JamieB-gu: Is there an advantage to using this over the plugin [mentioned in the docs](https://webpack.js.org/guides/output-management/#the-manifest)?

webpack.config.js
```diff
@@ -105,8 +106,11 @@ const clientConfig = {
     target: 'web',
     output: {
         path: path.resolve(__dirname, 'dist/assets'),
-        filename: '[name].js',
+        filename: '[name].[contenthash].js',
     },
+    plugins: [
+        new WebpackAssetsManifest(),
```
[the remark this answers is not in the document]
webb04: Good point, I've moved the hashing to the `clientProduction` config so livereload should still work.

I've also added a `prod` script that needs running before it's started by PM2 @alexduf

package.json
webb04: We don't have any html in the project so I was a bit reluctant to revert back to having a wrapper document.

package.json
```diff
@@ -9,18 +9,19 @@
...
     "upload": "node-riffraff-artifact",
-    "test:build": "webpack --config webpack.config.js --config-name server --env.test",
+    "test:build": "npm run build:client:prod; webpack --config webpack.config.js --config-name server --env.test",
     "test": "npm run test:build && jest --verbose",
     "test:coverage": "npm run test && open coverage/index.html",
     "lint": "node_modules/eslint/bin/eslint.js -c config/.eslintrc.js ./ --ext '.ts,.tsx'",
     "watch:server": "webpack --config webpack.config.js --config-name server --env.watch",
     "watch:client": "webpack-dev-server --config webpack.config.js --config-name client",
-    "watch": "npm run watch:server & npm run watch:client",
+    "watch": "npm run build:client; npm run watch:server; npm run watch:client",
```
webb04: This initial `npm run build:client;` is for the server to read `dist/assets/manifest.json`

package.json
JamieB-gu: Hmm, I'm not sure that hash url is that reliable, I was talking about [this plugin](https://github.com/danethurber/webpack-manifest-plugin), which is suggested at the bottom of that docs page.

webb04: Oh I see, that's used by 1.1 million compared to 58K... I'll swap them over

webpack.config.js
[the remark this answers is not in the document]
alexduf: resolved, the script will be run at build time 👍

tsconfig.json
```diff
@@ -11,6 +11,10 @@
         "moduleResolution": "node",
         "noUnusedLocals": true,
         "noUnusedParameters": false,
-        "baseUrl": "src/"
+        "baseUrl": ".",
+        "paths": {
+            "*": ["src/*", "dist/*"]
```
alexduf: Is this to import the json file into the final server.js bundle?

JamieB-gu: Partly, yes. Because the `manifest.json` file exists in a separate directory (`dist`) to the rest of the source code (`src`), we need to [tell the TS compiler](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping) where it can look for it.

alexduf: What lead me to question this was the dependency we suddenly have on the dist directory. It's probably not a good idea.

So I'm reconsidering my suggestion of bundling the json in the server.js. I think the initial solution that you had where the file is loaded separately is probably more sane!

webb04:
@JamieB-gu due to https://www.npmjs.com/package/webpack-manifest-plugin#optionswritetofileemit, this server running on 3040 won't be able to load the JavaScript files. But the webpack-dev-server running on 8080 will work

alexduf:
I think the hashed assets feature may deserve to become its own module. (something like `assets.ts`).

In that module you could ensure that when loading the file, if it's not found in production the server shuts down (throws an exception). You can also avoid passing `{[bundle: string]: string} | null` everywhere, but instead pass `string => string | null`. That function would be defined in the that new assets module, and passed down to the components.

This would provide a nice isolation between the concern of loading/mapping on the manifest and the rest of the code.

src/server/assets.ts
```diff
@@ -0,0 +1,17 @@
...
+
+type AssetMapping = { [key: string]: string } | null;
+
+export const getMappings = (): AssetMapping => {
+    if (process.env.NODE_ENV !== "production") {
+        return null;
+    }
+
+    try {
+        return JSON.parse(fs.readFileSync('./dist/assets/manifest.json').toString());
```
alexduf:
```suggestion
        return JSON.parse(fs.readFileSync('./manifest.json').toString());
```

PR #210, opened 2020-02-26
GLA-1828: get the font size from native side

jeanlauliac:
## Why are you doing this?

Size can be set native size with the size selector. This is not a feature we'd want to deprecate, as indeed plenty of people requested that feature on the Editions app for example, and plenty more would probably miss it on the Live app.

Screenshot 2020-02-26 at 13 55 48

My plan is two-fold:

1. add a `getFontSize` in the protocol to allow the JS to fetch the font-size persisted in the native side and apply it immediately to the article. This PR addresses that.
2. add a `updateFontSize` in the protocol to allow the native size to do further updates to the font-size after the article has been rendered. This is for when people will slide the font size slider left and right and we want to update the size in real-time. (We wouldn't want to have to reload the whole article every time.)

To avoid having a flashing effect of the size suddenly changing size, this PR adopts the approach of hiding the article at first, and only reveal it once we have set the size. I think this is fine because the article wouldn't show while it's loading in the first place.

I'd like to validate this PR with reviewers before proceeding to add the new function in the Thrift API, hence this is not ready for merging, but requires comments.

## Changes

Set `body` to be hidden by default, then set the font size and make it visible again once we got the font-size from the Thrift API (for now, we mock that part).


JamieB-gu:
> the article wouldn't show while it's loading in the first place

Would you be able to elaborate on this? Do you mean while the JS is downloading/parsing?

The HTML should be rendered as soon as it's available, with the JS loading/parsing asynchronously some time after that.

src/client/setup.ts
```diff
@@ -1,6 +1,18 @@
...
 
 import {logger} from "../logger";
+// import { nativeClient } from 'native/nativeApi';
+
+// REPLACE by `nativeClient.getFontSize()`
+function getFontSize() {
+    return new Promise(resolve => setTimeout(() => resolve(67.5), 200));
+}
+
+async function fontSize() {
```
JamieB-gu:
```suggestion
async function fontSize(): void {
```

src/client/setup.ts
```diff
@@ -1,6 +1,18 @@
 // ----- Procedures ----- //
 
 import {logger} from "../logger";
+// import { nativeClient } from 'native/nativeApi';
+
+// REPLACE by `nativeClient.getFontSize()`
+function getFontSize() {
```
JamieB-gu: I *think* this might return the following:

```suggestion
function getFontSize(): Promise<number> {
```

src/components/shared/page.tsx
```diff
@@ -48,14 +49,15 @@ const PageStyles = css`
     ${fontFace("Guardian Icons", new None, new None, "/public/fonts/icons.otf")}
 
     font-size: 62.5%;
-    background: white;
+    background: ${neutral[97]};
```
JamieB-gu: What's the thinking behind this?

jeanlauliac:
> > the article wouldn't show while it's loading in the first place
> 
> Would you be able to elaborate on this? Do you mean while the JS is downloading/parsing?
> 
> The HTML should be rendered as soon as it's available, with the JS loading/parsing asynchronously some time after that.

Yes you're right I wasn't clear. What I mean is it would not flash because it would transition from the HTML loading/parsing (so the background color should be the default set by the native layer), to the HTML showing up, but with a hidden `body` (background color set by the `html` element which should be consistent), then to the `body` showing up with the right size. If we were to render the `body` before fetching the font-size, then it would "jump" from one font size to the final one, which probably wouldn't be a good user experience. So the tradeoff of that particular proposed solution is to make the time-to-full-content longer while preventing a "jump" in font size.

I'm not super happy with this myself, though. Another solution is to have the native layer parse the HMTL and force the `font-size`of the `html` tag to the desired/persisted font size. This is a bit problematic because it makes the logic implicit and not encoded clearly in the protocol/contract between native and web layers, but it has the benefit that the article/item would load with the correct font size right out of the box without all this `display: none` trick.

I'd love to discuss this in person, I'll send an invite if that's okay..

src/components/shared/page.tsx
jeanlauliac: Alongside the dark variant, this would allow the background color "flashing" a different color while the `body` is hidden. It's arguable whether the whole strategy is desirable or not however. Let's take some time to chat later on.

jeanlauliac:
Closing for now as it was a Spike but according with offline discussion, this PR is the final approach we would take. We just need to add a new function in the Thrift protocol such as `getSize`. See https://theguardian.atlassian.net/browse/GLA-1828 for details.

PR #212, opened 2020-02-27
Renderer.ts tests

webb04:
## Why are you doing this?
Health OKR is to reach 75% test coverage across our main server-side parsing logic.
This PR increases the coverage to above 75% for `renderer.ts` and `item.ts`


src/item.test.ts
```diff
@@ -0,0 +1,400 @@
...
+        mediaId: "",
+        mediaApiUri: "https://image.co.uk",
+        suppliersReference: "",
+        imageType: ""
+    }
+})
+
+describe('fromCapi returns correct Item', () => {
+    test('media', () => {
+        const item = fromCapi(JSDOM.fragment)(contentWithTag('type/audio'));
```
JamieB-gu: It should be possible to use partial application here to avoid having to call `fromCapi` with all arguments each time:

```ts
const f = fromCapi(JSDOM.fragment);

// Tests

const item = f(contentWithTag('type/audio'));
const item = f(reviewContent);

// etc.
```

src/renderer.test.ts
```diff
@@ -1,6 +1,68 @@
...
+const tweetElement = (): BodyElement =>
+    ({
+        kind: ElementKind.Tweet,
+        content: JSDOM.fragment('<span>Tweet contents<span>').querySelectorAll('span'),
+    })
+
+const render = (element: BodyElement): ReactNode[] =>
+    renderAll('dummySalt')(Pillar.news, [element]);
+
+const renderTextElement = compose(render, textElement);
```
JamieB-gu: Nice!

PR #225, opened 2020-03-05
remove build time gzipping

webb04:
## Why are you doing this?
The JavaScript files aren't working in production, I've got a feeling it's being gzipped by webpack and then again in s3.

I'll deploy this and test.


webb04:
This fixes the scripts in prod, looks like they're not gzipped but we can add that back at some point.

PR #226, opened 2020-03-06
update ads when fonts loaded

webb04:
## Why are you doing this?
Sometimes the ad positions change without a DOM mutation e.g. fonts loading. This should correctly reposition ads.

### DIFF
--- a/src/client/article.ts
+++ b/src/client/article.ts
@@ -14,6 +14,16 @@ import ReactDOM from 'react-dom';
 
 // ----- Run ----- //
 
+interface FontFaceSet {
+    readonly ready: Promise<FontFaceSet>;
+}
+
+declare global {
+    interface Document {
+        fonts: FontFaceSet;
+    }
+}
+
 function getAdSlots(): AdSlot[] {
     const advertSlots = document.getElementsByClassName('ad-slot');
 
@@ -55,7 +65,7 @@ function insertAds(): void {
         observer.observe(targetNode, config);
 
         try {
-            (document as any).fonts.ready.then(() => nativeClient.updateAdverts(getAdSlots()))
+            document.fonts.ready.then(() => nativeClient.updateAdverts(getAdSlots()))
         } catch (e) {
             logger.error(`font loading API not supported: ${e}`)
         }
--- a/src/client/article.ts
+++ b/src/client/article.ts
@@ -53,6 +53,9 @@ function insertAds(): void {
 
         const observer = new MutationObserver(callback);
         observer.observe(targetNode, config);
+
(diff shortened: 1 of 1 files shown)


src/client/article.ts
```diff
@@ -53,14 +53,17 @@ function insertAds(): void {
 
         const observer = new MutationObserver(callback);
         observer.observe(targetNode, config);
+
+        // handle font loading and changes that do not mutate DOM
+        setTimeout(() => nativeClient.updateAdverts(getAdSlots()), 200);
```
JamieB-gu: Every 200ms forever? If this is for content like fonts that load in relatively early, could we have this run for mebbe the first 5-10 seconds and see if that works ok?

webb04: This will only run once after 200ms. I think this probably isn't a problem when we're loading fonts from device. Maybe it does need to be a few times after 200ms just to make sure...

The mutation Observer is working well though, clicking to follow an author (changes DOM) will reposition the ads.

JamieB-gu: Ah my bad, sorry, I read that as `setInterval`.

Hmm, yeah I wonder if it's worth running it a handful of times since font load times are non-deterministic. It might be very fast on recent devices, but perhaps a touch slower on older ones?

JamieB-gu: There's the [font loading API](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Font_Loading_API) which might work for this purpose. It looks like it's mostly supported in both Safari and Chrome, although I'm not 100% sure about the font loading event in Safari...

src/client/article.ts
```diff
@@ -53,14 +53,20 @@ function insertAds(): void {
 
         const observer = new MutationObserver(callback);
         observer.observe(targetNode, config);
+
+        try {
+            (document as any).fonts.ready.then(() => nativeClient.updateAdverts(getAdSlots()))
```
webb04: classed as an experimental API so it's not in Typescript definitions.
Seems to work on iOS and says it's supported on Android though https://developer.mozilla.org/en-US/docs/Web/API/CSS_Font_Loading_API

alexanderedge: I'm up for using this and seeing how it works

JamieB-gu: We could introduce a subset of the types [suggested here](https://github.com/Microsoft/TypeScript/issues/30984#issuecomment-581953895) to avoid having to type this as `any`.

```ts
interface FontFaceSet {
    readonly ready: Promise<FontFaceSet>;
}

declare global {
    interface Document {
        fonts: FontFaceSet;
    }
}
```

PR #228, opened 2020-03-06
Image and BodyImage Components

JamieB-gu:
## Why are you doing this?

Creating standalone components for images in general, and the `BodyImage` in particular. This also encompasses greater integration with the design system.

**Note:** The tests are probably going to fail because of this error: `"global" coverage threshold for branches (75%) not met: 74.92%`... They are actually all passing, it's just the coverage that's unhappy.

cc @SiAdcock I've created a new SVG for the triangle on image captions: around line 62 in `BodyImage`.

## Changes

- Removed `BodyImage` and `FigureElement` from `renderer`
- Created new components: `Image` and `BodyImage`
- Migrated image helpers (e.g. `srcset`) into component
- Updated `Paragraph` to handle text padding directly

### DIFF
--- a/src/components/bodyImage.tsx
+++ b/src/components/bodyImage.tsx
@@ -55,7 +55,7 @@ const captionStyles = css`
     color: ${text.supporting};
 `;
 
-const BodyImage: FC<Props> = ({ figcaption, image, pillar }) =>
+const BodyImage: FC<Props> = ({ figcaption, image, pillar }: Props) =>
     <figure css={styles}>
         <Image {...image} sizes={sizes} />
         <figcaption css={captionStyles}>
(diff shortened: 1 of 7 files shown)


JamieB-gu:
The failing Snyk check is related to a dependency in master. It's not caused by this PR as there are no dependency changes made here.

src/components/bodyImage.tsx
```diff
@@ -0,0 +1,76 @@
...
+    ${textPadding}
+    padding-top: ${remSpace[2]};
+    color: ${text.supporting};
+`;
+
+const BodyImage: FC<Props> = ({ figcaption, image, pillar }: Props) =>
+    <figure css={styles}>
+        <Image {...image} sizes={sizes} />
+        <figcaption css={captionStyles}>
+            <svg
```
webb04: This seems like it could be reusable with dotcom and elsewhere, maybe one for the design system? @SiAdcock

JamieB-gu: Yup, long term it may be able to go [here](https://github.com/guardian/source/tree/master/src/core/svgs).

PR #231, opened 2020-03-10
rehydrate liveblogs client side

webb04:
## Why are you doing this?
Liveblogs require some client side functionality. This PR rehydrates liveblogs client side ready to do some of this.

## Changes

- Client side rehydration option for components
- Use design system component button in liveblogs
- Update relative times in an interval (liveblogs)
- Fetch blocks on load more (WIP: requires new mapi endpoint)
- Stubs load more requests (json omitted due to containing capi data)
- Moves the salt up the component tree to avoid needing it to rehydrate

## Followup
- Reduce bundle size, we don't need all the types (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#-type-only-imports-and-export)


src/components/liveblog/loadMore.tsx
```diff
@@ -1,46 +1,19 @@
...
-            margin-right: 16px;
-            margin-top: -2px;
-            font-size: 1.2rem;
-        }
-    }
-`;
+import { Pillar } from 'pillar';
+import { SvgPlus } from "@guardian/src-svgs"
+// import { ThemeProvider } from 'emotion-theming'
+import { Button } from '@guardian/src-button'
```
JamieB-gu: 👍

src/server/server.ts
```diff
@@ -83,9 +88,19 @@ async function serveArticle(req: Request, res: ExpressResponse): Promise<void> {
...
+              hydrationProps
+            } = Page({ content, imageSalt, getAssetLocation });
+            const props = JSON.stringify({ ...hydrationProps, imageSalt })
+              .replace(/’/g, "&rsquo;")
+              .replace(/“/g, "&ldquo;")
+              .replace(/”/g, "&rdquo;")
+              .replace(/…/g, "&hellip;");
             res.set('Link', getPrefetchHeader(resources));
             res.write('<!DOCTYPE html>');
+            res.write(`<script>window.__INITIAL__DATA__ = ${props}</script>`)
```
JamieB-gu: I think this will sidestep our bundling process and remove some of the value of our CSP. Could we achieve this another way? Either through additional requests or [`<meta>` tags](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta)?

src/components/liveblog/block.tsx
```diff
@@ -93,19 +93,29 @@ const LiveblogBlock = ({
...
         .withDefault(null)
 
     const relativeLastModified: JSX.Element | null = lastModifiedDate
         .fmap<JSX.Element | null>(date => <time>Last updated: {formatDate(date)}</time>)
         .withDefault(null)
 
+
+    const [dateJsx, setDateJsx] = useState(relativeFirstPublished(firstPublishedDate));
+
+    useEffect(() => {
```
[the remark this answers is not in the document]
webb04: Good question, it does still render and only runs when it's hydrated.

src/client/liveblog.ts
```diff
@@ -1,8 +1,35 @@
...
+import LiveblogBody from 'components/liveblog/body';
+import { createElement as h } from 'react';
+import { Content } from 'mapiThriftModels';
 
 // ----- Run ----- //
 
+declare global {
+    interface Window {
+        __INITIAL__DATA__: Content & {
+            imageSalt: string;
```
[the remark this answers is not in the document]
webb04: Replied on chat about this ^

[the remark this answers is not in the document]
alexduf: what was the outcome?

src/client/liveblog.ts
```diff
@@ -1,8 +1,35 @@
...
+        };
+    }
+}
+
 setup();
+
+const browserParser = (string: string): DocumentFragment =>
+    parse(new DOMParser())(string)
+        .toOption()
+        .fmap(frag => frag)
```
JamieB-gu: What does this do?

src/client/liveblog.ts
```diff
@@ -1,8 +1,38 @@
...
+        };
+    }
+}
+
 setup();
+
+const browserParser = (html: string): DocumentFragment =>
+    parse(new DOMParser())(html)
+        .toOption()
+        .withDefault(new DocumentFragment())
```
JamieB-gu: If you want to avoid converting to `Option` this is possible, although may be less clear...

```suggestion
    parse(new DOMParser())(html)
        .either(_ => new DocumentFragment(), identity)
```

Perhaps we should implement `withDefault` for `Result` at some point?

src/client/liveblog.ts
```diff
@@ -1,8 +1,38 @@
...
+import { createElement as h } from 'react';
+import { Content } from 'mapiThriftModels';
+import { parse } from './parser';
+import { ImageMappings } from 'components/shared/page';
 
 // ----- Run ----- //
 
+declare global {
+    interface Window {
+        __INITIAL__DATA__: Content & {
```
JamieB-gu: Does the full CAPI object get added to the window? Is there anything in a CAPI response that should be private, bearing in mind MAPI has a higher level of access? @davidfurey @alexduf ?

src/components/image.ts
```diff
@@ -62,6 +62,7 @@ interface Props {
     width: number;
     height: number;
     salt: string;
+    imageMappings?: ImageMappings;
```
JamieB-gu: ?

```suggestion
    imageMappings: Option<ImageMappings>;
```

src/components/immersive/article.tsx
```diff
@@ -76,19 +77,20 @@ const HeaderImageStyles = css`
 // ----- Component ----- //
 
 interface Props {
+    imageMappings: ImageMappings;
     imageSalt: string;
```
JamieB-gu: Is this still used here?

src/components/liveblog/body.tsx
```diff
@@ -22,32 +25,43 @@ const LiveBodyStyles = (pillarStyles: PillarStyles): SerializedStyles => css`
...
-const LiveblogBody = ({ pillar, blocks, imageSalt }: LiveblogBodyProps): JSX.Element => {
+const LiveblogBody = (props: LiveblogBodyProps): JSX.Element => {
+    const { pillar, blocks: initialBlocks, imageMappings, totalBodyBlocks } = props;
+    const [blocks] = useState(initialBlocks);
 
-    const initialBlocks = blocks.slice(0, 7);
-    const LoadMore = ({ total }: { total: number }): JSX.Element | null => total > 10
-        ? <LiveblogLoadMore pillar={pillar}/> 
+    async function loadMoreBlocks(): Promise<void> {
+        const response = await fetch('?date=2020-03-09T11%3A11%3A49Z&filter=newer');
```
JamieB-gu: Hardcoded date?

src/components/liveblog/body.tsx
```diff
@@ -22,32 +25,43 @@ const LiveBodyStyles = (pillarStyles: PillarStyles): SerializedStyles => css`
...
 
-const LiveblogBody = ({ pillar, blocks, imageSalt }: LiveblogBodyProps): JSX.Element => {
+const LiveblogBody = (props: LiveblogBodyProps): JSX.Element => {
+    const { pillar, blocks: initialBlocks, imageMappings, totalBodyBlocks } = props;
+    const [blocks] = useState(initialBlocks);
 
-    const initialBlocks = blocks.slice(0, 7);
-    const LoadMore = ({ total }: { total: number }): JSX.Element | null => total > 10
-        ? <LiveblogLoadMore pillar={pillar}/> 
+    async function loadMoreBlocks(): Promise<void> {
```
JamieB-gu: Does this need to live within `LiveblogBody`? Can it exist at the top level of this module?

src/components/liveblog/body.tsx
```diff
@@ -22,32 +25,43 @@ const LiveBodyStyles = (pillarStyles: PillarStyles): SerializedStyles => css`
...
+        const response = await fetch('?date=2020-03-09T11%3A11%3A49Z&filter=newer');
+        const buffer = Buffer.from(await response.arrayBuffer());
+        const transport = new BufferedTransport(buffer);
+        const protocol = new CompactProtocol(transport);
+        const blocks: Blocks = Blocks.read(protocol);
+        console.log(blocks);
+    }
+
+    const LoadMore = ({ total }: { total: number }): JSX.Element | null => total > 7
```
JamieB-gu: Same question about the top level.

src/components/liveblog/body.tsx
```diff
@@ -22,32 +25,43 @@ const LiveBodyStyles = (pillarStyles: PillarStyles): SerializedStyles => css`
...
 
-    const initialBlocks = blocks.slice(0, 7);
-    const LoadMore = ({ total }: { total: number }): JSX.Element | null => total > 10
-        ? <LiveblogLoadMore pillar={pillar}/> 
+    async function loadMoreBlocks(): Promise<void> {
+        const response = await fetch('?date=2020-03-09T11%3A11%3A49Z&filter=newer');
+        const buffer = Buffer.from(await response.arrayBuffer());
+        const transport = new BufferedTransport(buffer);
+        const protocol = new CompactProtocol(transport);
+        const blocks: Blocks = Blocks.read(protocol);
```
JamieB-gu: I think this instance of the variable `blocks` is shadowing the one in the enclosing scope. Is that intentional?

src/components/shared/headerImage.tsx
```diff
@@ -34,13 +35,14 @@ const Styles = (width: number, height: number): SerializedStyles => css`
 
 interface HeaderImageProps {
     image: Option<Image>;
-    imageSalt: string;
+    imageMappings?: ImageMappings;
+    imageSalt?: string;
```
JamieB-gu: Why does `imageSalt` become an optional prop here?

src/components/shared/headerImage.tsx
```diff
@@ -34,13 +35,14 @@ const Styles = (width: number, height: number): SerializedStyles => css`
 
 interface HeaderImageProps {
     image: Option<Image>;
-    imageSalt: string;
+    imageMappings?: ImageMappings;
```
JamieB-gu: ?

```suggestion
    imageMappings: Option<ImageMappings>;
```

src/components/shared/page.tsx
```diff
@@ -83,6 +85,27 @@ interface BodyProps {
...
+export interface ImageMappings {
+    [key: string]: string;
+}
+
+const getImageMappings = (imageSalt: string, capi: Content): ImageMappings => {
+    return capi.blocks?.body
+        ?.flatMap(block => block.elements
+            .filter(element => element.type === ElementType.IMAGE)
+            .flatMap(element => {
+                return element.assets.map(asset => {
```
JamieB-gu: Can this use *just* the master asset?

src/item.ts
```diff
@@ -127,6 +127,7 @@ type LiveBlock = {
 interface Liveblog extends Fields {
     design: Design.Live;
     blocks: LiveBlock[];
+    totalBodyBlocks: number;
```
JamieB-gu: 👍

src/renderer.ts
```diff
@@ -174,7 +175,8 @@ const text = (doc: DocumentFragment, pillar: Pillar): ReactNode[] =>
 interface ImageProps {
     url: string;
     alt: string;
-    salt: string;
+    imageMappings?: ImageMappings;
+    imageSalt?: string;
```
JamieB-gu: Why has the salt become optional here?

src/renderer.ts
```diff
@@ -174,7 +175,8 @@ const text = (doc: DocumentFragment, pillar: Pillar): ReactNode[] =>
 interface ImageProps {
     url: string;
     alt: string;
-    salt: string;
+    imageMappings?: ImageMappings;
```
JamieB-gu: Is this used?

src/server/server.ts
```diff
@@ -60,6 +61,10 @@ async function serveArticlePost(
 
 async function serveArticle(req: Request, res: ExpressResponse): Promise<void> {
   try {
+    // mock content from mapi
+    if (req.query.date || req.query.filter) {
```
JamieB-gu: What's this checking for?

src/server/server.ts
```diff
@@ -83,9 +88,19 @@ async function serveArticle(req: Request, res: ExpressResponse): Promise<void> {
...
+            const {
+              resources,
+              element,
+              hydrationProps
+            } = Page({ content, imageSalt, getAssetLocation });
+            const props = JSON.stringify({ ...hydrationProps, imageSalt })
+              .replace(/’/g, "&rsquo;")
+              .replace(/“/g, "&ldquo;")
+              .replace(/”/g, "&rdquo;")
+              .replace(/…/g, "&hellip;");
```
JamieB-gu: Should this live in the transform function?

src/server/server.ts
JamieB-gu: Any thoughts on this?

src/components/liveblog/body.tsx
webb04: I'm going to limit this PR to the rehydration and then after look at the mapi endpoint and fetching

src/components/liveblog/body.tsx
webb04: The state is currently scoped to `liveblogBody` and that's the only component rehydrated so far so I'm not sure.

Eventually we'd want to do `setBlocks(...blocks, newBlocks)` after we fetch them.

webb04: Oh but I can pass a function down, will try that.

src/components/shared/page.tsx
webb04: yep, good idea 👍

src/server/server.ts
webb04: Updated the comment. In the context of a webview it will request <mapi_url>?date=x&filter=newer to fetch more blocks. But we can mock that or make a request to mapi for local development.

src/server/server.ts
webb04: Could be yes. All of these problems are introduced when I stringify the capi object for rehydration, will look for a way around it.

webb04: It's not the stringify but it's node's `res.write` that is causing the problem...

webb04: So this was all happening before `<meta charset="UTF-8"/>` was set 🤦‍♂️

src/server/server.ts
webb04: I've tried some ways to use `meta` tags and also just plan JSON with script tags but couldn't find a way...

src/components/shared/page.tsx
JamieB-gu: I *think* if you're just working with master assets you can sidestep the `map` and the second `flatMap`:

```ts
const bodyBlocks = (capi: Content): Block[] =>
    capi.blocks?.body ?? [];

const mainBlocks = (capi: Content): Block[] =>
    capi.blocks?.main !== undefined ? [capi.blocks.main] : [];

const getImageMappings = (imageSalt: string, capi: Content): ImageMappings => {
    const blocks = [...bodyBlocks(capi), ...mainBlocks(capi)];

    return blocks.flatMap(block => block.elements)
        .filter((element: BlockElement) => element.type === ElementType.IMAGE)
        .reduce((mappings, elem) => {
            const asset = elem.assets.find(asset => asset.typeData?.isMaster);

            if (asset !== undefined && asset.file !== undefined) {
                const url = new URL(asset.file);
                return { ...mappings, [url.pathname]: sign(imageSalt, url.pathname) };
            }

            return mappings;
        }, {});
}
```

src/server/server.ts
JamieB-gu: Did you hit problems with `<script type="application/json">${props}</script>`?

Do we have to hydrate immediately? Since the UI won't change until the first poll request goes out (or until the user clicks "view more"), can we wait until the first request comes back to hydrate? That way we can perform a standard request for JSON and use the result to populate the body of the page.

webb04: I'll try the script json again, I think my problem before was UTF-8 related. I think the additional network request would have to be a new mapi endpoint? and another request to capi?

src/client/liveblog.ts
webb04: We've got a PR to fetch less data in MAPI before sending to apps-rendering. We should remove anything unnecessary there

src/server/server.ts
webb04: pushed a commit using `<script type="application/json">`

src/client/liveblog.ts
JamieB-gu: Ok. We can filter out anything that shouldn't be exposed publicly there. In fact, we should do that anyway since that endpoint itself will presumably be public?

src/components/shared/page.tsx
JamieB-gu: I don't think there's a need to change everything around as I did in the example above. However, I think the `flat` method returning `any` is something that could be remedied to increase our type safety.

webb04:
@JamieB-gu Used your function to avoid the any creeping in from `flat()`. Also noticed there was no mappings for contributors (Avatars and Cutouts) so added them to `imageMappings`

JamieB-gu:
Nice! 👍

src/components/shared/page.tsx
```diff
@@ -129,6 +135,25 @@ const getImageMappings = (imageSalt: string, capi: Content): ImageMappings => {
...
+    const { createdBy, lastModifiedBy, ...mainBlocks } = main as Block;
+
+    return {
+        id: contentId, type, webTitle, webUrl, apiUrl, fields, references, isHosted,
+        tags: tags.map(({ type, webTitle, webUrl, id, references, apiUrl }) => ({ type, webTitle, webUrl, id, references, apiUrl })),
+        blocks: {
+            totalBodyBlocks,
+            main: mainBlocks,
+            body: body?.map(block => {
+                const { createdBy, lastModifiedBy, ...bodyBlocks } = block;
```
JamieB-gu, line 151: Do these shadow the `createdBy` and `lastModifiedBy` variables in the outer scope?

src/components/shared/page.tsx
```diff
@@ -129,6 +135,25 @@ const getImageMappings = (imageSalt: string, capi: Content): ImageMappings => {
         return { ...blockMappings, ...getContributorMappings(capi.tags, imageSalt) }
 }
 
+const liveblogHydrationProps = (capi: Content): Content => {
+    const { id: contentId, type, webTitle, webUrl, apiUrl, fields, tags, references, isHosted, blocks } = capi;
+    const { main, body, totalBodyBlocks } = blocks as Blocks;
+    const { createdBy, lastModifiedBy, ...mainBlocks } = main as Block;
```
JamieB-gu, line 141: Are `createdBy` and `lastModifiedBy` used here?

JamieB-gu, line 141: What happens if `blocks` and/or `main` are `undefined`?

src/components/shared/page.tsx
webb04: deduped in a function

src/components/shared/page.tsx
webb04: this was a way to pull them out and use all the other fields, blocks have a lot of required fields so this seemed to make sense

webb04: refactored to avoid casting

PR #234, opened 2020-03-11
Bundle fonts with the assets

alexduf:
## Why are you doing this?

- To make things prettier on CODE and PROD
- To have prettier error messages in our logs
- To align our development environement with our production one. In prod we don't really have much choice but to have all static assets under /assets, so I'm following the same pattern in dev mode

I've also modified our teamcity build to add the `copy-fonts` step


.storybook/preview-head.html
```diff
@@ -3,30 +3,30 @@
         background: white;
     }
     
-    @font-Face{font-family:"Guardian Text Egyptian Web"; font-weight:400; src:url("./public/fonts/GuardianTextEgyptian-Reg.ttf");}
```
alexduf: I suspect the dot here was unintentional as it would have resolved a different font depending on how many slashes there are in the URL?

src/server/server.ts
```diff
@@ -53,7 +53,7 @@ async function serveArticlePost(
     res.write(html);
     res.end();
   } catch (e) {
-    logger.error(`This error occurred, but I don't know why: ${e}`);
+    logger.error(`This error occurred`, e);
```
alexduf: This prints the stack trace

src/server/server.ts
```diff
@@ -120,7 +120,7 @@ app.all('*', (request, response, next) => {
   next();
 });
 
-app.use('/public', express.static(path.resolve(__dirname, '../public')));
+app.use('/assets', express.static(path.resolve(__dirname, '../assets')));
```
alexduf: Even though `/assets` is already defined the line after, the way express works is:
If it can find the file: great, otherwise search in the next folder.

So fonts are first served, and compiled javascript will go through two hops. This has no effect in production as it's served as static files from the CDN.

PR #235, opened 2020-03-11
Dateline

JamieB-gu:
## Why are you doing this?

A new component to handle the publication date. I've matched dotcom's format for printing dates.

**Note:** It looks like dotcom use the Edition to derive the timezone (`London`, `New York` or `Sydney`) in which to show the date (does that sound right @gtrufitt?). I've haven't implemented this yet, as it may be that we want to do things differently in the apps. Therefore I've marked the timezone as `UTC` for now.

## Changes

- Created a `Dateline` component
- Added `Dateline` stories
- Migrated bylines to use `Dateline`
- Added logic to date module to format date
- Removed `dayjs` dependency
- Wrote tests for the date module

## Screenshots

byline


alexduf:
Maybe a good way to display the date would be with the user's timezone?

webb04:
> Maybe a good way to display the date would be with the user's timezone?

I wrote some client side JavaScript to do this, maybe that can be reused?

JamieB-gu:
> Maybe a good way to display the date would be with the user's timezone?
> I wrote some client side JavaScript to do this, maybe that can be reused?

Possibly. I'm trying to confirm whether we want to set timezone based on Edition or the device's timezone.

If Edition we can do one of the following:

1. Set it on the server, do nothing on the client.
2. Set UTC on the server and overwrite on the client.

If device's timezone, then we can set UTC on the server and overwrite it on the client. For 2/3 of these options, setting UTC on the server is the way to go, hence why I've done that in this PR. Once I confirm which approach to take, we can apply the client-side code (and possibly editionalise on the server if needs be).

gtrufitt:
> : It looks like dotcom use the Edition to derive the timezone (London, New York or Sydney) in which to show the date (does that sound right @gtrufitt?). 

Here's how *Frontend* does it (I assume we pull in the same to DCR for parity)

https://github.com/guardian/frontend/blob/129823d000305939b7f6700ca1a27bd18af5dd23/common/app/views/fragments/meta/dateline.scala.html#L24

Test for GUDateTimeFormat that shows US, AUS edition conversion.

https://github.com/guardian/frontend/blob/136f39c5651ed973797a8570df4cb04f79fde943/common/test/views/support/GUDateTimeFormatTest.scala#L6-L5

JamieB-gu:
Ah and the `formatTimeForDisplay` function [derives the timezone](https://github.com/guardian/frontend/blob/e5ff00c6881ab79c7f0beb1e34e15a366555d8fb/common/app/views/support/package.scala#L130) from Edition. Excellent, thanks @gtrufitt.

PR #237, opened 2020-03-12
thumbnail images

webb04:
## Why are you doing this?
New component and styles for when `role: "thumbnail"` is set.

https://www.theguardian.com/commentisfree/2020/mar/11/2020-budget-guardian-panel-verdict-rishi-sunak

## Screenshots

| Before | After |
| --- | --- |
|  |  |
| Wide |  |

### DIFF
--- a/src/components/bodyImageThumbnail.tsx
+++ b/src/components/bodyImageThumbnail.tsx
@@ -2,15 +2,12 @@
 
 import React, { FC, ReactNode } from 'react';
 import { css } from '@emotion/core';
-import { breakpoints } from '@guardian/src-foundations';
-
 import Image, { Props as ImageProps } from 'components/image';
 import { Pillar } from 'pillar';
 
-
 // ----- Setup ----- //
 
-const sizes = `(min-width: ${breakpoints.phablet}px) 620px, 100vw`;
+const sizes = `(min-width: 140px)`;
 
 
 // ----- Component ----- //
--- a/src/components/bodyImageThumbnail.tsx
+++ b/src/components/bodyImageThumbnail.tsx
@@ -1,0 +1,38 @@
+// ----- Imports ----- //
+
+import React, { FC, ReactNode } from 'react';
+import { css } from '@emotion/core';
+import { breakpoints } from '@guardian/src-foundations';
+
+import Image, { Props as ImageProps } from 'components/image';
+import { Pillar } from 'pillar';
+
+
+// ----- Setup ----- //
+
+const sizes = `(min-width: ${breakpoints.phablet}px) 620px, 100vw`;
+
+
+// ----- Component ----- //
+
+interface Props {
+    image: Omit<ImageProps, 'sizes'>;
+    figcaption: ReactNode;
+    pillar: Pillar;
+}
+
+const styles = css`
+    display: block;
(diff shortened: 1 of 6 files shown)


src/components/bodyImageThumbnail.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { css } from '@emotion/core';
+import { breakpoints } from '@guardian/src-foundations';
+
+import Image, { Props as ImageProps } from 'components/image';
+import { Pillar } from 'pillar';
+
+
+// ----- Setup ----- //
+
+const sizes = `(min-width: ${breakpoints.phablet}px) 620px, 100vw`;
```
JamieB-gu: Is the thumbnail `620px` wide from phablet and 100% of the viewport width otherwise?

src/components/image.ts
```diff
@@ -75,12 +77,21 @@ const styles = (width: number, height: number): SerializedStyles => css`
     }
 `;
 
-const Image: FC<Props> = ({ url, sizes, salt, alt, width, height, caption, credit }) => {
+const thumbnailStyles = (width: number, height: number): SerializedStyles => css`
+    float: left;
+    clear: left;
+    width: 8.75rem;
```
JamieB-gu: ?

```suggestion
    width: 8.75rem;
    height: calc(8.75rem * ${height / width});
```

src/item.ts
```diff
@@ -82,6 +82,7 @@ type Image = {
     file: string;
     width: number;
     height: number;
+    role: string;
```
JamieB-gu:
```ts
enum Role {
    Thumbnail,
}
```

```suggestion
    role: Option<Role>;
```

src/components/bodyImageThumbnail.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+// ----- Setup ----- //
+
+const sizes = `(min-width: 140px)`;
+
+
+// ----- Component ----- //
+
+interface Props {
+    image: Omit<ImageProps, 'sizes'>;
```
alexduf: What's that `Omit` type?

JamieB-gu: Docs [here](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittk). I think this is derived from the `Props` I created for the `Image` component.

I haven't yet decided whether it was a good idea or not. I'm a bit cautious about making use of utility types... it often feels like you're getting round the fact that if you'd designed your types another way you wouldn't have this problem...

src/components/bodyImageThumbnail.tsx
```diff
@@ -0,0 +1,35 @@
+// ----- Imports ----- //
+
+import React, { FC, ReactNode } from 'react';
+import { css } from '@emotion/core';
+import Image, { Props as ImageProps } from 'components/image';
+import { Pillar } from 'pillar';
+
+// ----- Setup ----- //
+
+const sizes = `(min-width: 140px)`;
```
JamieB-gu: Does this also need a "source size value", as defined [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-sizes)?

Is it possible this should be:

```ts
const sizes = '140px';
```

PR #239, opened 2020-03-13
Instagram elements

webb04:
## Why are you doing this?
Handle Instagram elements: http://localhost:8080/film/2020/mar/12/kate-beckinsale-harvey-weinstein-rant-new-york-premiere

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -351,6 +351,14 @@ const render = (salt: string, pillar: Pillar) => (element: BodyElement, key: num
 
         case ElementKind.Tweet:
             return h(Tweet, { content: element.content, pillar, key });
+
+        case ElementKind.Instagram:
+            const props = {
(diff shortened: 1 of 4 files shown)


src/renderer.ts
```diff
@@ -351,6 +351,14 @@ const render = (salt: string, pillar: Pillar) => (element: BodyElement, key: num
 
         case ElementKind.Tweet:
             return h(Tweet, { content: element.content, pillar, key });
+
+        case ElementKind.Instagram:
+            const props = {
+                dangerouslySetInnerHTML: {
```
JamieB-gu: Is there anything in the Instagram HTML that prevents us from providing a similar implementation to the one we do [for tweets](https://github.com/guardian/apps-rendering/blob/89b09fde0368f94b2e86a09d16cf2a9aa1999773/src/renderer.ts#L319)?

PR #249, opened 2020-03-24
default to opinion when news pillar used

webb04:
## Why are you doing this?
I've never seen an Opinion article with news pillar styles. It is possible to use sports pillar colours: http://localhost:8080/football/commentisfree/2019/dec/18/mesut-ozil-china-row-western-brands-be-warned-self-censorship-wont-protect-you

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/item.ts
+++ b/src/item.ts
@@ -427,9 +427,11 @@ const fromCapi = (docParser: DocParser) => (content: Content): Item => {
             ...itemFieldsWithBody(docParser, content),
         };
     } else if (isComment(tags)) {
+        const item = itemFieldsWithBody(docParser, content);
         return {
             design: Design.Comment,
-            ...itemFieldsWithBody(docParser, content),
+            ...item,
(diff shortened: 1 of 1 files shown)


JamieB-gu:
Interestingly DCR [have some code](https://github.com/guardian/dotcom-rendering/blob/0832b716b8ca59cfaa982224a04e6fd0bbb5bf29/src/lib/pillars.ts#L60) that does something similar.

This is a big reason why I think we should de-couple the concept of "Pillar" from "colour scheme". When we treat them as the same we end up altering a fundamental pieces of information about an item of content just to get the colours we want. Perhaps we could start using the `ColourScheme` type, which handles this [here](https://github.com/guardian/types/blob/3e70de6fb9443650ec14f9e8c9d99a31e5d1f078/ColourScheme.ts#L22)?

cc @gtrufitt @oliverlloyd @SiAdcock

PR #244, opened 2020-03-19
MSS-1299: gallery articles

frankie297:
## Why are you doing this?
Adding a new gallery article type

PR #250, opened 2020-03-26
Refactor Byline

JamieB-gu:
## Why are you doing this?

*Apologies for the size of this PR! Some of it is just moving or deleting files, but it's still on the large side.*

This brings the byline and related bits of UI into line with the `Format` type and design system. What we've previously called `Byline` I've now termed `Metadata`, which brings us into line with DCR and frontend. `Byline` is now *just* the byline, what was previously referred to as `Author`. I've also added several stories to cover these new components.

I've left the liveblog alone as much as possible, to avoid causing problems for #231. Instead of deleting some deprecated components I've moved them into the liveblog directory, as that's the only place they're now used. Will look at changing this in a future PR.

**Note:** ESLint seems to struggle with the `FC` type... It's not able to validate prop types, and thus I've had to add redundant `Props` annotations to some components. This is not restricted to this PR, I've encountered it before. Will look into it as a separate issue.

## Changes

- Broke `Contributor` out into own module
- Allowed passing widths into `srcset` function
- Added storybook helpers module, contains `Pillar` selector
- Deleted `Author` components, now referred to as `Byline`
- Created new `Avatar` component
- Created new `Byline` component that just displays byline
- Created new `CommentCount` component
- Created new `Follow` component, migrated tests from previous component
- Introduced generic `Metadata` component
- Added custom `Metadata` component for standard articles
- Removed some old `Byline` components
- Moved some deprecated components to liveblog, for refactoring later
- Added a number of stories for new components

## Screenshots

Description | Screenshot
-|-
Byline | byline
Comment Byline | byline-comment
CommentCount | comment-count
Follow | follow
Metadata | metadata
Comment Piece | comment
Standard Article Metadata | standard-metadata


alexduf:
This is massive! any chance this could be broken down in smaller chunks to make the review easier?
If not or if it were to take more than an hour then I'll review the PR anyway. Let me know!

webb04:
The merge conflicts with the liveblog PR are going to be messy... as that PR had to update `imageSalt` to `imageMappings`

JamieB-gu:
> any chance this could be broken down in smaller chunks

I think that might be tricky, sorry 😔. The reason it scaled out this much is that a lot of these components are inter-related.

I think it's potentially smaller that it looks though, as a lot of the changes should be new files or file deletions; hopefully easier to read than lots of changes within files?

JamieB-gu:
> The merge conflicts with the liveblog PR are going to be messy... as that PR had to update `imageSalt` to `imageMappings`

Yeah I think it's fair to merge the liveblog PR first, it's been open much longer.

webb04:
Non links are no longer italic and pillar colour:
Screen Shot 2020-03-27 at 13 16 22
Screen Shot 2020-03-27 at 13 16 58

I think that "Follow" should be a bit bigger as that will be a tappable area
Screen Shot 2020-03-27 at 13 14 43

JamieB-gu:
> Non links are no longer italic and pillar colour:

Good spot! Will take a look.

JamieB-gu:
@webb04 the font colours/weights in the byline should be fixed now.

The size of the `Follow` text is an interesting one. On the current app this text matches the size of the publish date text, which is `14px`. However, this font size doesn't exist in the design system, and on dotcom the publish date text is `12px`, so I've matched them both to that.

I think this is one of those platform inconsistencies for design to decide on as part of the design system transition. It's possible that both bits of text should be larger; or just the follow text should be larger; or that they're both fine as they are...

Worth noting as well that we're currently missing the round follow icon on apps-rendering, which would increase the tapable area.

src/components/shared/keyline.tsx
```diff
@@ -6,8 +6,8 @@ import { from } from '@guardian/src-foundations/mq';
 import { Design } from 'item';
 
 const BaseStyles = css`
-    height: 12px;
-    margin-bottom: 4px;
```
webb04: I think this is useful underneath liveblog keylines

src/image.ts
```diff
@@ -42,6 +42,11 @@ function src(salt: string, input: string, width: number): string {
     return `${imageResizer}/${service}${url.pathname}?${params.toString()}`;
 }
 
+const srcsetWithWidths = (widths: number[]) => (url: string, salt: string): string =>
```
webb04: These functions have changed in the liveblog PR 😅

I noticed we still have them duplicated in the codebase.

src/components/commentCount.tsx
```diff
@@ -0,0 +1,66 @@
...
+}
+
+const getBubbleStyles = ({ pillar }: Format): SerializedStyles => {
+    const colours = getPillarStyles(pillar);
+
+    return bubbleStyles(colours.kicker);
+}
+
+const CommentCount: FC<Props> = ({ count, commentable, ...format }: Props) => {
+    if (!commentable) {
```
webb04: This was already in master, so don't feel the need to change but we could default to false earlier on `commentable = fields?.commentable ?? false`

src/components/commentCount.tsx
```diff
@@ -0,0 +1,66 @@
...
+        return null;
+    }
+
+    return (
+        <button css={getStyles(format)}>
+            <svg css={getBubbleStyles(format)} xmlns="http://www.w3.org/2000/svg" viewBox="0 0 9 8">
+                <rect x="0" y="0" width="9" height="6" rx="1.2" />
+                <polygon points="2,6 2,8 2.5,8 4,6" />
+            </svg>
+            { count.toLocaleString() }
```
webb04: Good idea!

src/image.ts
JamieB-gu: Yeah, that's my fault 😬. Fixing that is on my list, but I've held off for a bit to avoid conflicts with #231 and #244.

src/components/shared/keyline.tsx
JamieB-gu: The `margin-bottom`?

src/components/commentCount.tsx
JamieB-gu: I'm sorry, I'm not sure I understand what you mean. This is already a `boolean` at this point isn't it, we don't need to null coalesce it?

src/components/avatar.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+    return styles(colours.inverted);
+}
+
+const Avatar: FC<Props> = ({ contributors, imageSalt, className, ...format }: Props) => {
+    const [contributor] = contributors;
+
+    if (isSingleContributor(contributors) && contributor.bylineLargeImageUrl !== undefined) {
+        return (
+            <img
```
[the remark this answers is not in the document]
JamieB-gu: Well in theory all the browsers we support can [handle `srcset`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/srcset#Browser_compatibility) 🙂

But yeah, good spot, I'd agree that's good practice, I'll change it.

src/components/shared/keyline.tsx
webb04: I guess this can be replaced by Simon's work in the design system!

PR #251, opened 2020-03-26
only request necessary CAPI fields

webb04:
## Why are you doing this?
Request only what we need from Capi. This is only for local development but it should match prod in Mapi (PR in progress)

### DIFF
--- a/src/capi.ts
+++ b/src/capi.ts
@@ -102,11 +102,20 @@ const includesTweets = (content: Content): boolean => {
 
 const capiEndpoint = (articleId: string, key: string): string => {
 
+    const fields = [
+        'headline',
+        'standfirst',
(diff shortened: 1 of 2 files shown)


src/capi.ts
```diff
@@ -101,12 +101,21 @@ const includesTweets = (content: Content): boolean => {
 // ----- Functions ----- //
 
 const capiEndpoint = (articleId: string, key: string): string => {
+    // If you need a new field here, MAPI probably also needs updating
```
JamieB-gu: I wonder if longer term this is something that could live in the `types` repo and be shared with MAPI?

webb04: yeah that would be a good idea, otherwise it would be easy for them to get out of sync

PR #252, opened 2020-03-27
make paragraph styles specific to body

webb04:
## Why are you doing this?
Some paragraphs outside of the article body have their styles overridden by component styles.

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/paragraph.tsx
+++ b/src/components/paragraph.tsx
@@ -1,7 +1,9 @@
 // ----- Imports ----- //
 
 import { FC, ReactNode } from 'react';
-import React from '@emotion/core';
+import React, { css } from '@emotion/core';
+import { body } from '@guardian/src-foundations/typography';
+import { remSpace } from '@guardian/src-foundations';
 
 // ----- Component ----- //
 
@@ -9,8 +11,14 @@ interface Props {
     children?: ReactNode;
 }
 
+const styles = css`
+    ${body.medium()};
+    overflow-wrap: break-word;
+    margin: 0 0 ${remSpace[3]};
+`;
+
 const Paragraph: FC<Props> = ({ children }: Props) =>
-    <p>{children}</p>;
+    <p css={styles}>{children}</p>;
 
 
 // ----- Exports ----- //
--- a/src/components/paragraph.tsx
+++ b/src/components/paragraph.tsx
@@ -1,7 +1,7 @@
 // ----- Imports ----- //
 
-import { FC, ReactNode } from 'react';
-import React, { css } from '@emotion/core';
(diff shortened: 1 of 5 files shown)


webb04:
Some `<p>` tags require different styles like in the standfirst. Another option would be to create a new `render` function for them or pass in a param to the existing one.

JamieB-gu:
Ah yeah I see. Our `renderText` function at the moment assumes you're rendering text within the body and applies the corresponding components and styles.

This may be a situation where we assumed two pieces of UI were more similar than they actually are. Perhaps we do need a separate incarnation of `renderText` to handle the standfirst, which provides components tailored to its needs. As it is the current `renderText` probably contains checks for stuff that doesn't make sense in the standfirst, e.g. `BLOCKQUOTE` and `H2`.

What do you think?

webb04:
> Ah yeah I see. Our `renderText` function at the moment assumes you're rendering text within the body and applies the corresponding components and styles.
> 
> This may be a situation where we assumed two pieces of UI were more similar than they actually are. Perhaps we do need a separate incarnation of `renderText` to handle the standfirst, which provides components tailored to its needs. As it is the current `renderText` probably contains checks for stuff that doesn't make sense in the standfirst, e.g. `BLOCKQUOTE` and `H2`.
> 
> What do you think?

Good point, I'll make a new render function as it's quite limited inside standfirsts to things like paragraphs and lists

JamieB-gu:
Coincidentally I've just hit the same thing over in #250. The solution I went with was to create a much simpler `renderText` function inside the `byline` component.

src/components/paragraph.tsx
```diff
@@ -1,19 +1,18 @@
...
-
 // ----- Component ----- //
 
 interface Props {
     children?: ReactNode;
 }
 
 const styles = css`
-    ${body.medium()}
+    ${body.medium()};
```
JamieB-gu: Doesn't `body.medium` generate a set of CSS properties with trailing `;` for you?

src/renderer.ts
```diff
@@ -146,7 +145,7 @@ const textElement = (pillar: Pillar) => (node: Node, key: number): ReactNode =>
         case '#text':
             return transform(text, pillar);
         case 'SPAN':
-            return text;
+            return h('span', null, text);
```
JamieB-gu: If a `<span>` has no styles, does it make sense to just simplify it to a text node?

webb04: Could do, I thought this could break some styles so I'll revert. I just thought that this could be a really annoying bug for anyone expecting to see a span

PR #255, opened 2020-03-31
more robust is para check

webb04:
## Why are you doing this?

Fix paragraphs not showing in production.

### DIFF
--- a/src/ads.ts
+++ b/src/ads.ts
@@ -1,10 +1,13 @@
 import { createElement as h, ReactNode, ReactElement } from 'react';
+import Paragraph from 'components/paragraph';
 
 function insertPlaceholders(reactNodes: ReactNode[]): ReactNode[] {
     const adIndices = [3, 9];
     const flattenedNodes = reactNodes.flat();
 
-    const isPara = (node: { type: { name?: string } }): boolean => node?.type?.name === 'Paragraph';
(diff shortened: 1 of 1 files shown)


src/ads.ts
```diff
@@ -1,10 +1,13 @@
 import { createElement as h, ReactNode, ReactElement } from 'react';
+import Paragraph from 'components/paragraph';
 
 function insertPlaceholders(reactNodes: ReactNode[]): ReactNode[] {
     const adIndices = [3, 9];
     const flattenedNodes = reactNodes.flat();
 
-    const isPara = (node: { type: { name?: string } }): boolean => node?.type?.name === 'Paragraph';
+    const isPara = (node: ReactElement): boolean =>
+        node?.type === Paragraph;
```
JamieB-gu: As discussed offline, this works because `flattenedNodes` is implicitly `any`... In reality it's a `ReactNode[]`.

`ReactElement` is a subtype of `ReactNode`, and therefore technically you can't pass the items in `flattenedNodes` to this function.

However, I think this will work because `node.type` will exist for `Paragraph` elements, and therefore we'll get `true`. For any scenario where `node.type` doesn't exist (e.g. `typeof node === string`), optional chaining will return `undefined` (I think), and the expression will evaluate to `false`.

PR #256, opened 2020-04-01
line height tweaks

webb04:
## Changes

- Less space between bullet and text
- bullet points slightly closer (matches dotcom)
- Small increase to line-height in standfirst (matches dotcom)
- Smaller h2 font size (matches dotcom)
- Replace all bullets in transform. Some article use bullet points as article breaks http://localhost:8080/world/2020/apr/01/golden-eagles-us-dying-why

## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |
|  |  |


src/renderer.ts
```diff
@@ -122,7 +121,7 @@ const listItemStyles: SerializedStyles = css`
 `
 
 const HeadingTwoStyles = css`
-    font-size: 1.4rem;
+    font-size: 1.25rem;
     font-weight: 700;
```
JamieB-gu, line 125:
```suggestion
    ${headline.xxsmall({ fontWeight: 'bold' })}
```

PR #257, opened 2020-04-01
Audio and Embed elements

webb04:
## Why are you doing this?
- Render audio elements
- Render embed elements

## Screenshot

| Audio | Embed |
| --- | --- |
| Screen Shot 2020-04-01 at 16 09 11 |  |

### DIFF
--- a/src/item.ts
+++ b/src/item.ts
@@ -51,7 +51,8 @@ const enum ElementKind {
     Interactive,
     RichLink,
     Tweet,
-    Instagram
+    Instagram,
+    Audio
 }
 
 enum Role {
@@ -109,6 +110,9 @@ type BodyElement = {
     content: NodeList;
 } | {
     kind: ElementKind.Instagram;
+    html: string;
+} | {
+    kind: ElementKind.Audio;
     html: string;
 };
 
@@ -260,6 +264,10 @@ const parseElement =
         case ElementType.INSTAGRAM:
             const instagramHtml = element.instagramTypeData?.html ?? "";
             return new Ok({ kind: ElementKind.Instagram, html: instagramHtml });
+
+        case ElementType.AUDIO:
+            const audioHtml = element.audioTypeData?.html ?? "";
+            return new Ok({ kind: ElementKind.Audio, html: audioHtml });
 
         default:
             return new Err(`I'm afraid I don't understand the element I was given: ${element.type}`);
--- a/src/item.ts
+++ b/src/item.ts
@@ -222,10 +222,10 @@ const parseIframe = (docParser: DocParser) => (html: string): string | undefined
         return undefined;
     }
 
-    iframe?.removeAttribute('frameborder');
-    iframe?.setAttribute('style', 'border: none');
-    iframe?.setAttribute('sandbox', 'allow-scripts')
-    return iframe?.outerHTML;
+    iframe.removeAttribute('frameborder');
+    iframe.setAttribute('style', 'border: none');
+    iframe.setAttribute('sandbox', 'allow-scripts')
+    return iframe.outerHTML;
 }
 
 const parseElement =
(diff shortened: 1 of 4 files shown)


src/renderer.ts
```diff
@@ -374,6 +374,7 @@ const render = (pillar: Pillar, imageMappings: ImageMappings) =>
         case ElementKind.Tweet:
             return h(Tweet, { content: element.content, pillar, key });
 
+        case ElementKind.Audio:
```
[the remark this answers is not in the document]
webb04: Unfortunately all we get is:
```\n            <iframe\n                height=\"380\"\n                width=\"300\"\n                src=\"https://open.spotify.com/embed/track/\"\n                frameborder=\"0\"\n                allowfullscreen\n            ></iframe>\n        ```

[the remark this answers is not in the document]
Comment (JamieB-gu) [reply src/renderer.ts:377]: Is it possible to parse this, as we do other HTML, into a `DocumentFragment` and apply a view that we control?

For example, we can avoid use of the [deprecated `frameborder`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-frameborder), and add [`sandbox` restrictions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox).

File: src/item.ts [hunk 2]
```diff
@@ -90,6 +92,11 @@ type Image = {
     role: Option<Role>;
 }
 
+type Audio = {
+    kind: ElementKind.Audio;
+    html: string;
```
JamieB-gu, line 97: Now that we're parsing this into a `DocumentFragment`, can I suggest we simply pull out the iframe URL and change this to:

```suggestion
    src: string;
```

And then in `renderer.ts` we can do:

```ts
interface AudioProps {
    src: string;
}

const audioStyles = css`
    border: none;
`;

const Audio: FC<AudioProps> = ({ src }) =>
    <iframe src={src} sandbox="allow-scripts" height="380" width="300" />
```

I'm not too clear what the rules are for `height` and `width`. If we can't decide on them ourselves and need to derive those from the CAPI response each time, we can pull those out when we grab the `src` from the `iframe`, so that we have:

```ts
// item.ts
type Audio = {
    kind: ElementKind.Audio;
    src: string;
    height: number;
    width: number;
}

// renderer.ts
interface AudioProps {
    src: string;
    height: number;
    width: number;
}
```

src/item.ts
```diff
@@ -205,6 +218,21 @@ const parseImage = (docParser: DocParser) => (element: BlockElement): Option<Ima
...
+    if (!iframe || !src) {
+        return undefined;
+    }
+
+    return {
+        src,
+        width: iframe.getAttribute('width') ?? "300",
+        height: iframe.getAttribute('height') ?? "380"
+    }
+}
```
JamieB-gu, line 234:
```suggestion
const parseIframe = (docParser: DocParser) => (html: string): Result<string, Audio> => {
    const iframe = docParser(html).querySelector('iframe');
    const src = iframe?.getAttribute('src');

    if (!iframe || !src) {
        return new Err('No iframe within audioTypeData.html');
    }

    return new Ok({
        kind: ElementKind.Audio,
        src,
        width: iframe.getAttribute('width') ?? "300",
        height: iframe.getAttribute('height') ?? "380",
    });
}
```

src/item.ts
```diff
@@ -257,10 +285,31 @@ const parseElement =
...
+        case ElementType.AUDIO:
+            const { html: audioHtml } = element.audioTypeData ?? {};
+            if (!audioHtml) {
+                return new Err('No html field on audioTypeData')
+            }
+            return fromNullable(parseIframe(docParser)(audioHtml))
+                .fmap<Result<string, Audio>>(({ src, width, height }) => {
+                    return new Ok({ kind: ElementKind.Audio, src, width, height })
+                })
+                .withDefault(new Err('No iframe within audioTypeData.html'))
```
JamieB-gu, line 311:
```suggestion
            return parseIframe(docParser)(audioHtml);
```

PR #265, opened 2020-04-06
Generate thrift packages

webb04:
## Why are you doing this?
This is still WIP but I wanted to discuss it before going too far. 

We communicate with MAPI using thrift. So far we've just needed the `Content` type that's provided [here](https://github.com/guardian/content-api-models/) but we'll need new custom types to send things like comment counts. We also use thrift for things like liveblog block fetching, so we need some custom types for that too.

- The bash script can be moved to GitHub actions and can use Docker
- This approach would let us generate `src/mapiThriftModels` automatically rather than manually
- I think it's a good idea for `appsRendering.thrift` to live in this repo
- Is it a good idea to publish the scala packages from this repo? How would that work?
- Files can be renamed


JamieB-gu:
👍 

It's possible this sort of thing could live in the [`types` repo](https://github.com/guardian/types)?

webb04:
> 👍
> 
> It's possible this sort of thing could live in the [`types` repo](https://github.com/guardian/types)?

We could do, although would you expect scripts to generate the types in there? and I imagine this would only be used by apps-rendering and mapi.

appsRendering.sh
```diff
@@ -0,0 +1,17 @@
+# scala package
+git clone https://github.com/guardian/content-api-models
+cd content-api-models
+cat ../appsRendering.thrift >> models/src/main/thrift/content/v1.thrift
```
alexduf, line 4: I think it might make sense to do that the other way around.

consider `content-api-models` as a folder that must reflect how the repo was when it was cloned.

Then you can read it, but shouldn't write anything in it. You can copy the files into a temporary directory if you want, then concatenate everything into one big thrift file

Also don't forget to add that folder to the .gitignore

api-models/src/main/thrift/appsRendering.thrift
```diff
@@ -0,0 +1,17 @@
...
+namespace scala com.gu.mobile.ssr.models
+
+struct appsRendering {
+
+    1: required v1.Content content
+
+    2: optional i32 commentCount 
+}
+
+struct appsRenderingBlocks {
```
JamieB-gu, line 12: Is this for the client-side requests?

api-models/src/main/thrift/appsRendering.thrift
```diff
@@ -0,0 +1,17 @@
+include "v1.thrift"
+
+namespace scala com.gu.mobile.ssr.models
```
JamieB-gu, line 3:
```suggestion
namespace scala com.gu.mobile.ar.models
```

build.sbt
```diff
@@ -0,0 +1,36 @@
+import java.io.File
+lazy val apiModels = project.in(file("api-models"))
+  .settings(
+    name := "api-models",
+    organization := "com.gu",
+    version := "0.1", 
+    scalaVersion := "2.12.11",
```
JamieB-gu, line 7: You've got a couple of mentions of Scala 2.13 in the shell script above, is that important?

alexduf:
>Is this something that should live apps-rendering? 

I think it makes sense for the API to own the models associated with it

JamieB-gu:
> > Is this something that should live apps-rendering?
> 
> I think it makes sense for the API to own the models associated with it

Is that also an argument for having this live in MAPI?

alexduf:
> Is that also an argument for having this live in MAPI?

Not really as MAPI is a client of this API

JamieB-gu:
Would it make sense to have all the API model code live in its own directory?

alexduf:
Right! I think we've got the generation of scala code and publication of the jars sorted out.
We now need to address the other comments in this PR and merge.

Regarding the generation of the typescript models, I'm wondering if we couldn't use Scrooge to do that? It would be yet another way to generate typescript types, but it would be nicely integrated with this repo. It would also generate all the CAPI types, so we could get rid of our current way of generating typescript types.

webb04:
@alexduf we can use Scrooge, thrift or the credit karma library. This was working before, but it was generating one huge file https://github.com/guardian/apps-rendering/pull/265/files#diff-53eb2fc614d2c36d7f9c182451ef4f86R13.

I think I used the credit karma library originally to generate the models in `mapiThriftModels`

api-models/src/main/thrift/appsRendering.thrift
webb04: yep we can rename the type, but this is for liveblog fetching. You would be able to setState using this format and render the list of blocks.

alexduf:
I think this can be merged in its current state. It will add support for scala models, but not for typescript. We'll come back to it when the SBT plugin is released

build.sbt
```diff
@@ -0,0 +1,67 @@
...
+      pushChanges
+    ),
+
+    Compile / runThrift := {
+      val appsRenderingFile = baseDirectory.value / "src" / "main" / "thrift" / "appsRendering.thrift"
+      val importDirectoriesOptions = (Compile / scroogeUnpackDeps).value.map { dependency =>
+        s"-I ${dependency.getPath}"
+      }.mkString(" ")
+      val outputDirOption = s"-o ${target.value}"
+      val exitCode = Process(s"thrift --gen js:ts ${importDirectoriesOptions} ${outputDirOption} ${appsRenderingFile.getPath}", baseDirectory.value) !
```
webb04, line 61: Will Thrift be available everywhere this is run?

alexduf: it will have to be!

alexduf:
This is ready for one last review I believe!

README.md
```diff
@@ -89,6 +89,12 @@ npm run storybook
...
-1. Import all types into a thrift file `v1.thrift`
-2. `node_modules/.bin/thrift-typescript --target thrift-server --outDir definitions v1.thrift`
+### Generating Apps rendering API models
+
+In the SBT repl:
+```sbtshell
+project scalaApiModels
+release
+project tsApiModels
+releaseNpm <versionNumber>
```
webb04, line 99: could the npm version drift from the sbt package version?

webb04: Also do you need npm credentials?

alexduf: > could the npm version drift from the sbt package version?
technically yes, although I recommend we use the same. I haven't managed to massage the scrooge plugin into releasing both scala (maven) and typescript (npm) at the same time

>Also do you need npm credentials?
Good shout I'll add that

webb04: LGTM, can't approve as I originally opened the PR

alexduf: ha, thanks!

PR #270, opened 2020-04-08
increase cutout image width

webb04:
## Why are you doing this?
Cutout images were low quality compared to dotcom and templates. Also added some more space between the top of the cutout and the headline.


webb04:
There's quite a few pixels related to this code so I might leave that for a wider cleanup.

PR #271, opened 2020-04-08
Separate Format Module

JamieB-gu:
## Why are you doing this?

Breaking `Format` out into its own module. There are a few reasons why:

- Improve code clarity
- Less reason to import the entire `item` module in rendering code, which may reduce client bundle sizes
- Preparation for future changes, including integration with the [guardian/types](https://github.com/guardian/types) repo

## Changes

- Pulled `Format` type into its own module
- Updated imports
- Dropped `SpecialReport` and `GuardianLabs` from `Design` since we don't think they're used


JamieB-gu:
Failing on test coverage, as expected from #244.

PR #272, opened 2020-04-08
Write a test for renderCaption

frankie297:
## Why are you doing this?
Write a test for the renderCaption function for galleries

### DIFF
--- a/src/renderer.test.ts
+++ b/src/renderer.test.ts
@@ -75,7 +75,12 @@ const instagramElement = (): BodyElement =>
 const render = (element: BodyElement): ReactNode[] =>
     renderAll({})(Pillar.news, [element]);
 
+const renderCaption= (element: BodyElement): ReactNode[] =>
+    renderMedia({})(Pillar.news, [element]);
+
 const renderTextElement = compose(render, textElement);
+
+const renderCaptionElement = compose(renderCaption, imageElement)
 
 describe('renderer returns expected content', () => {
     test('ImageElement returns null for no url', () => {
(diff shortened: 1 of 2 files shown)


src/renderer.test.ts
```diff
@@ -126,6 +126,16 @@ describe('renderer returns expected content', () => {
 
         expect(text.flat().length).toBe(7);
     });
+
+    test ('Renders caption node types', () => {
+        const text = renderCaption(JSDOM.fragment('this caption contains <em>html</em>, <br>html, <strong>html</strong>'),  Pillar.news);
```
webb04, line 131:
```suggest
const text = renderCaption(JSDOM.fragment('this caption contains'),  Pillar.news);
```

We're only testing the first item so I guess we don't need to have the other types?

webb04: Ah I realise this will likely reduce the test coverage though

frankie297: 👍

src/renderer.ts
```diff
@@ -507,4 +507,5 @@ export {
     ImageElement,
     getHref,
     renderMedia,
+    renderCaption,
```
JamieB-gu, line 510: Little bit cautious about exporting functions solely for use in the tests. Should we be testing the public API rather than a private one?

src/renderer.test.ts
```diff
@@ -75,8 +75,13 @@ const instagramElement = (): BodyElement =>
 const render = (element: BodyElement): ReactNode[] =>
     renderAll({})(Pillar.news, [element]);
 
+const renderCaption= (element: BodyElement): ReactNode[] =>
```
JamieB-gu, line 78:
```suggestion
const renderCaption = (element: BodyElement): ReactNode[] =>
```

PR #275, opened 2020-04-15
darken Culture colours to improve colour contrast

webb04:
## Why are you doing this?

This darkens the pillar colours for Opinion and Art articles. Before they would fail in Chrome accessibility audits due to low contrast. We use the kicker colours for links.

Do the colours need to be updated in Source? These new colours seem quite dark.


webb04:
Yes it diverges from the designs and from dotcom (who also have the same contrast warning).

I think it must just be below the threshold, so maybe we close this?

JamieB-gu:
> Yes it diverges from the designs and from dotcom (who also have the same contrast warning).
> 
> I think it must just be below the threshold, so maybe we close this?

If it diverges I think we need to discuss with design first, because a change like this will have wide-reaching consequences.

webb04:
I'd be interested to hear what @SiAdcock thinks as it seems weird to see a contrast warning for the arts kicker on a white background.

webb04:
Reverted the Opinion change for the time being and used the colour suggested by Ben https://theguardian.atlassian.net/browse/MSS-1586

SiAdcock:
@webb04 All I know is that Sport and Culture kicker contrast issues are long standing on web, and I'd love to see them fixed! But @JamieB-gu is right, it needs input from design.

https://whocanuse.com/?b=f6f6f6&c=a1845c&f=17&s=b
https://whocanuse.com/?b=f6f6f6&c=0084c6&f=17&s=b

PR #277, opened 2020-04-16
chromatic github action

webb04:
## Why are you doing this?

Run chromatic when pushing to GitHub.

Do we need the GitHub action as it looks like we've got a UI tests job running? @oliverlloyd

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,0 +1,17 @@
+name: "Chromatic"
+on: push
+
+jobs:
+    test:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v1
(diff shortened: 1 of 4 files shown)


.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,17 @@
+name: "Chromatic"
+on: push
```
JamieB-gu, line 2: Do we want all branches, or just PRs?

oliverlloyd: Chromatics docs:

```
GitHub actions can run based on any github event, but we recommend to run the workflow containing the chromatic step on push event. The action will work on pull-request events too, although it comes with some caveats. All other events will not.
```

PR #279, opened 2020-04-16
Refactor HeaderImage

JamieB-gu:
## Why are you doing this?

This is the first step in some improvements to the way we handle images. It (hopefully) simplifies the header image as a single component using the `Format` type. I'm trying to keep PRs small, so this doesn't include everything that I plan to change; for example, the new `Img` component will be used more widely in future PRs.

## Changes

- Created single component for header images using the `Format` type
- Updated all the articles to use the new component
- Removed redundant styles
- Created new `Img` component
- Removed now-unused `ImageElement` component from renderer


src/renderer.ts
```diff
@@ -251,51 +250,6 @@ const text = (doc: DocumentFragment, pillar: Pillar): ReactNode[] =>
...
-    height: number;
-    captionString: string;
-    caption: DocumentFragment;
-    credit: string;
-    pillar: Pillar;
-}
-
-const imageStyles = (width: number, height: number): SerializedStyles => css`
-    height: calc(100vw * ${height / width});
-    background: ${neutral[97]};
```
webb04, line 269: Should we keep this? we have it in the templates to show where an image will load in

JamieB-gu:
> noticed this on liveblogs

Ah, that was actually intentional. The full width image looked quite odd to me so I implemented a design closer to what dotcom do. That may or may not have been a good idea...

src/renderer.ts
JamieB-gu: Yes good spot, looks like I lost that somewhere moving code around.

benwuersching:
@JamieB-gu @webb04 On iPad in portrait view we always have main article images full-width, so it seems more consistent to keep it the same on liveblogs. Dotcom doesn't do images full-width by default, it's just a different setup.

JamieB-gu:
Ok, updated.

PR #285, opened 2020-04-17
Chromatic Continue On Error

JamieB-gu:
## Why are you doing this?

Prevents chromatic/storybook issues from blocking the build while we stabilise the action. From the docs [here](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error).

## Changes

- Added `continue-on-error` to chromatic build


JamieB-gu:
This was due to #284; turns out that had nothing to do with Chromatic so I'm closing this.

PR #286, opened 2020-04-20
match current ad placeholder logic

webb04:
## Why are you doing this?
Match current logic on apps.
This should hopefully be easier to configure if requirements change.

### DIFF
--- a/src/ads.ts
+++ b/src/ads.ts
@@ -1,8 +1,19 @@
 import { createElement as h, ReactNode, ReactElement } from 'react';
 import Paragraph from 'components/paragraph';
 
+function getAdIndices(): number[] {
+    const adEveryNParagraphs = 6;
+    const firstAdIndex = 3;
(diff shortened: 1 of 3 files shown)


src/ads.ts
```diff
@@ -1,8 +1,19 @@
...
 import Paragraph from 'components/paragraph';
 
+function getAdIndices(): number[] {
+    const adEveryNParagraphs = 6;
+    const firstAdIndex = 3;
+    const totalAds = 15;
+
+    const indiciesAfterFirstAd = [...Array(totalAds-1).keys()]
+        .map(index => firstAdIndex + (adEveryNParagraphs * ++index))
+    return [firstAdIndex, ...indiciesAfterFirstAd];
```
JamieB-gu, line 11: 👍 

JS could do with a [standard library](https://docs.python.org/3/library/functions.html#func-range)...

PR #288, opened 2020-04-20
Consolidate Images

JamieB-gu:
## Why are you doing this?

Following on from #279, this consolidates our handling of images into a single data type, reducing duplication. Image data and parsing should now all be handled by the `image.ts` module, and I've fixed the issue of duplicated logic in the `image.ts` component by deleting it.

This PR also fixes a couple of bugs:

- Thumbnail images had an invalid "sizes" attribute
- Captions and credits were being merged during parsing, which made it impossible to manipulate and style credits separately (this was flagged by @frankie297 in #244).

## Changes

- Created new `Image` data type to represent image data
- Migrated `Role` to image module
- Migrated image parsing to image module
- Deleted old `Image` component
- Stopped credits being merged with captions during parsing
- Migrated `BodyImage` and `BodyImageThumbnail` over to use `Img` component
- Fixed invalid "sizes" attribute in thumbnail images


src/image.ts
```diff
@@ -46,11 +72,46 @@ const srcsetWithWidths = (widths: number[]) => (url: string, mappings: ImageMapp
...
+        return new Some({
+            src: asset.file,
+            alt: fromNullable(data?.alt),
+            width: asset.typeData.width,
+            height: asset.typeData.height,
+            caption: fromNullable(data?.caption).fmap(docParser),
+            credit: fromNullable(data?.credit).andThen<string>(
+                c => data?.displayCredit ? new Some(c) : new None()
+            ),
+            nativeCaption: data?.caption ?? '',
```
webb04, line 99: Should we make this optional like caption and credit? https://github.com/guardian/bridget/blob/master/thrift/native.thrift#L14

JamieB-gu: Good idea, done 👍

PR #291, opened 2020-04-21
use vw in height calculation

webb04:
1. set height on headerImage to avoid jumping content and sometimes incorrect ad positions
2. pass string as alt value on image element

### DIFF
--- a/src/components/img.ts
+++ b/src/components/img.ts
@@ -26,7 +26,7 @@ const Img: FC<Props> = ({ image, sizes, imageMappings, className }) =>
         sizes,
         srcSet: srcset(image.src, imageMappings),
         src: src(imageMappings, image.src, 500),
-        alt: image.alt,
+        alt: image.alt.withDefault(''),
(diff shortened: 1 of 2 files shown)


src/components/img.ts
```diff
@@ -26,7 +26,7 @@ const Img: FC<Props> = ({ image, sizes, imageMappings, className }) =>
         sizes,
         srcSet: srcset(image.src, imageMappings),
         src: src(imageMappings, image.src, 500),
-        alt: image.alt,
+        alt: image.alt.withDefault(''),
```
JamieB-gu, line 29: Yeah I noticed this... Interesting that it doesn't throw a type error 🤔

webb04: the good old `[object Object]`

JamieB-gu: Ok, it appears it's not throwing a type error because of `styledH`... If you switch to JSX it *does* throw an error.

PR #293, opened 2020-04-22
Consistent News Colours

JamieB-gu:
## Why are you doing this?

Brings the news colours into line with the other pillar palettes.

## Changes

- Used `news[800]` rather than `neutral[97]` for news `soft`

Before | After
-|-
before | after


benwuersching:
This looks good to me @JamieB-gu 👍

PR #295, opened 2020-04-22
support photo essay articles

webb04:
## Changes

- Support new image role "halfWidth"
- Remove old anchor styles on immersive articles
- Render as immersive if an article contains a displayHint of "photoEssay"
- Abstract `bodyImageProps`

## Screenshots

`HalfWidthBodyImage`

Screen Shot 2020-04-22 at 17 45 09

### DIFF
--- a/src/components/bodyImageHalfWidth.tsx
+++ b/src/components/bodyImageHalfWidth.tsx
@@ -4,18 +4,19 @@ import React, { FC } from 'react';
 import { css, SerializedStyles } from '@emotion/core';
 import Img from 'components/img';
 import { BodyImageProps as Props } from 'image';
-import { remSpace } from '@guardian/src-foundations';
+import { remSpace, breakpoints } from '@guardian/src-foundations';
 
 // ----- Setup ----- //
 
-const size = `calc(50% - ${remSpace[1]})`;
+const figureWidth = `calc(50% - ${remSpace[1]})`;
+const size = `(min-width: ${breakpoints.phablet}px) 310px, 50vw`;
 
 // ----- Component ----- //
 
 const styles = css`
     margin: ${remSpace[1]} 0 0 0;
     display: inline-block;
-    width: ${size};
+    width: ${figureWidth};
 
     + .halfWidth, + .halfWidth + .halfWidth + .halfWidth {
         margin-left: ${remSpace[2]};
@@ -27,7 +28,7 @@ const styles = css`
 `;
 
 const imgStyles = (width: number, height: number): SerializedStyles => css`
-    height: calc(${size} * ${height / width});
+    height: calc(${figureWidth} * ${height / width});
     width: 100%;
 `;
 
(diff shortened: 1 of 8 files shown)


src/components/bodyImageHalfWidth.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+import React, { FC } from 'react';
+import { css, SerializedStyles } from '@emotion/core';
+import Img from 'components/img';
+import { BodyImageProps as Props } from 'image';
+import { remSpace } from '@guardian/src-foundations';
+
+// ----- Setup ----- //
+
+const size = `calc(50% - ${remSpace[1]})`;
```
JamieB-gu, line 11: Unfortunately percentages [aren't allowed](https://html.spec.whatwg.org/multipage/images.html#source-size-value) in the `sizes` attribute; this might work with `px`, `rem` or `vw` instead.

src/components/bodyImageHalfWidth.tsx
```diff
@@ -0,0 +1,52 @@
...
+// ----- Component ----- //
+
+const styles = css`
+    margin: ${remSpace[1]} 0 0 0;
+    display: inline-block;
+    width: ${size};
+
+    figcaption {
+        display: none;
+    }
```
JamieB-gu, line 22: Does this mean that you can pass a `figcaption`, it will be rendered in the DOM, but will be hidden by CSS? Would it be preferable not to accept `children` for half width images?

src/image.ts
```diff
@@ -98,7 +105,16 @@ const parseImage = (docParser: (html: string) => DocumentFragment) =>
...
+                role => {
+                    switch(role) {
+                        case 'thumbnail':
+                            return new Some(Role.Thumbnail)
+                        case 'halfWidth':
+                            return new Some(Role.HalfWidth)
+                        default:
+                            return new None()
+                    }
+                }
```
JamieB-gu, line 117: 👍

src/components/bodyImageHalfWidth.tsx
webb04: I wonder if the output of calc works? I just tested increasing the screen size from phone to tablet and it requested two sized images `1000` and `2000`

JamieB-gu: `calc` is valid according to that spec, it comes under the definition of "math functions".

src/components/bodyImageHalfWidth.tsx
```diff
@@ -0,0 +1,48 @@
...
+const size = `(min-width: ${breakpoints.phablet}px) 310px, 50vw`;
+
+// ----- Component ----- //
+
+const styles = css`
+    margin: ${remSpace[1]} 0 0 0;
+    display: inline-block;
+    width: ${figureWidth};
+
+    + .halfWidth, + .halfWidth + .halfWidth + .halfWidth {
```
gtrufitt, line 21: tenor-18653901

PR #298, opened 2020-04-24
Render Immersive Interactives

webb04:
## Changes

- Immersive Interactive articles
- Render interactive atoms

## Screenshots
Screen Shot 2020-04-24 at 18 40 16


src/renderer.ts
```diff
@@ -363,18 +398,36 @@ const render = (format: Format, imageMappings: ImageMappings) =>
 
             return h('div', props);
         }
-    }
+
+        case ElementKind.ContentAtom:
+            if (element.atom.atomType === AtomType.INTERACTIVE) {
```
JamieB-gu, line 403: Does this only apply to atoms that appear on full-page interactives? Or is it any interactive atom, including ones that appear in standard articles?

src/components/interactive/article.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const Styles = css`
+
+    h1, h2, h3, h4, h5, h6 {
+        margin: 0;
+    }
+
+    .meta__extras {
+        display: none;
+    }
```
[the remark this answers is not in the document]
webb04: In the current templates we have very few styles: https://github.com/guardian/mobile-apps-article-templates/blob/master/ArticleTemplates/assets/scss/interactive.scss

It looks like dotcom have some more: https://github.com/guardian/frontend/blob/3095a28e951b63ae993aa3d3fcc59a3cf83afc4b/static/src/stylesheets/head.interactive.scss

This sometimes makes it harder for interactive developers inheriting different styles on different platforms. They usually have to apply app specific styles when there's an `ios` or `android` class on the body. We probably won't have this in apps-rendering so I moved in some of the common app styles they use (we have social links in the topbar so they aren't needed inside `meta__extras`)

It would be really useful if we could share the same styles on dotcom/apps for immersive interactives. @gtrufitt @oliverlloyd

src/renderer.ts
```diff
@@ -312,12 +345,14 @@ const imageComponentFromRole = (role: Role): FC<BodyImageProps> => {
...
-const render = (format: Format, imageMappings: ImageMappings) =>
+const render = (format: Format, imageMappings: ImageMappings, excludeStyles = false) =>
     (element: BodyElement, key: number): ReactNode => {
     switch (element.kind) {
 
         case ElementKind.Text:
-            return text(element.doc, format);
+            return excludeStyles
+                ? Array.from(element.doc.childNodes).map(plainTextElement)
+                : text(element.doc, format)
```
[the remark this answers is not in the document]
webb04: Let me test a few more today, that might be possible.

src/renderer.ts
webb04: This will appear on other article types too I think. 

Due to the API we'll need to handle the other types differently, and will have to make some new components:

`const html = element.atom.data.recipe?.html`

src/renderer.ts
```diff
@@ -363,18 +398,36 @@ const render = (format: Format, imageMappings: ImageMappings) =>
 
             return h('div', props);
         }
-    }
+
+        case ElementKind.ContentAtom:
+            if (element.atom.atomType === AtomType.INTERACTIVE) {
+                const html = element.atom.data.interactive?.html
+                const styles = element.atom.data.interactive?.css
+                const js = element.atom.data.interactive?.mainJS
```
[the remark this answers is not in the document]
webb04: Sounds reasonable to me if an atom only had html and css. I'm not sure if they're always expected to have the three fields though.

 The `atomsJS` and `atomsCSS` fields in Mapi aren't optional.

src/components/interactive/article.tsx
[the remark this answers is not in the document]
gtrufitt: Yeah, this would be really good - removing that uncertainty of rendering for Visuals would be 💯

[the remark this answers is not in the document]
JamieB-gu: If we could have an identical foundation (e.g. a shared stylesheet) of styles that they're working against on both platforms, that would be amazing! Could mean developing against dotcom and having the apps version "just work" most of the time 🤔

src/renderer.ts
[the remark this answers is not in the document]
JamieB-gu: So would it be reasonable to fail the rendering of the atom if any of these are missing?

src/renderer.ts
JamieB-gu: Perhaps we'll come up with a new data type for atoms we support:

```
const enum AtomKind {
    Interactive,
    Recipe,
}

interface Atom {
    kind: AtomKind;
    js: string;
    css: string;
    html: string;
}
```

webb04:
I've replicated the current behaviour on apps: minimal styles and setting a classname of `ios` or `android` on the body for interactives to apply app specific styles.

`interactive atoms` unfortunately rely on existing page styles which means we will have to duplicate a few styles in `interactiveAtom` http://localhost:8080/world/ng-interactive/2020/apr/27/the-traffic-data-that-shows-the-road-into-and-out-of-covid-19-lockdown

src/renderer.ts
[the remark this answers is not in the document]
webb04: I've seen some interactive atoms that rely on page styles. This is now a bit tricky when we style individual components rather than having page stylesheets.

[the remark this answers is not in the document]
JamieB-gu: I think if we want to fix that short term for interactives we can just "polyfill" those styles in at the page level.

For interactive atoms that appear on other articles I agree that's a problem. I think best case would be to solve this upstream in the atoms, because relying on legacy styles available in frontend is brittle and probably something we'd like to avoid.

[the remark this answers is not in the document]
webb04: yeah, hopefully it's not much more than link colours and line-heights but we may have to add more styles to the `interactiveAtom` component.

We should meet with Sean Clarke before releasing this to production.

webb04:
We can probably still merge this, but we shouldn't release to production until dotcom-rendering has implemented `interactive` type articles.

We should meet with dotcom and interactive developers to discuss:
- Available css/classnames on the page that's consistent across dotcom and apps
- Backwards compatibility of older interactives
- Dark mode

@JamieB-gu @gtrufitt

src/client/interactive.ts
```diff
@@ -0,0 +1,7 @@
+const body = document.querySelector("body");
+const userAgent = navigator.userAgent || navigator.vendor;
+
+if (/android/i.test(userAgent)) {
+    body?.classList.add("android");
+    body?.classList.remove("ios");
+}
```
JamieB-gu, line 7: Are we likely to run up against Chrome's plans to [phase out](https://wicg.github.io/ua-client-hints/) UA strings?

src/components/standfirst.tsx
```diff
@@ -121,7 +121,7 @@ function content(standfirst: DocumentFragment, item: Item): ReactNode {
 
 const Standfirst: FC<Props> = ({ item }) =>
     item.standfirst.fmap<ReactElement | null>(standfirst =>
-        <div css={getStyles(item)}>{content(standfirst, item)}</div>,
+        <div className="standfirst" css={getStyles(item)}>{content(standfirst, item)}</div>,
```
JamieB-gu, line 124: What do full-page interactives use this class for?

src/components/shared/page.tsx
```diff
@@ -269,7 +285,7 @@ function Page({ content, imageSalt, getAssetLocation }: Props): ElementWithResou
                 <meta name="viewport" content="initial-scale=1, maximum-scale=5" />
                 <link rel="stylesheet" href="native://fontSize.css" />
             </head>
-            <body>
+            <body className="ios">
```
JamieB-gu, line 288: What do full-page interactives use this class for?

src/item.ts
```diff
@@ -32,7 +34,8 @@ const enum ElementKind {
     Instagram,
     Audio,
     Embed,
-    Video
+    Video,
+    ContentAtom
```
JamieB-gu, line 38: What's the difference between an "atom" and a "content atom"?

src/item.ts
```diff
@@ -67,6 +70,11 @@ type Video = {
     width: string;
 }
 
+type ContentAtom = {
+    kind: ElementKind.ContentAtom;
+    atom: Atom;
```
JamieB-gu, line 75: How much information do we need from this type? Could we create a narrower type that looks something like this:

```ts
const enum AtomKind {
    Interactive,
    Recipe,
}

interface Atom {
    kind: AtomKind;
    js: string;
    css: string;
    html: string;
}
```

src/renderer.ts
```diff
@@ -363,18 +400,42 @@ const render = (format: Format, imageMappings: ImageMappings) =>
...
+
+                if (!html || !styles || !js) {
+                    return null;
+                }
```
JamieB-gu, line 412: 👍

src/renderer.ts
```diff
@@ -363,18 +400,42 @@ const render = (format: Format, imageMappings: ImageMappings) =>
...
+
+                const atom = h(InteractiveAtom, { html, styles, js, format });
+
+                if (format.design !== Design.Interactive) {
+                    return h('iframe', null, atom);
```
JamieB-gu, line 417: 👍 👍

src/server/server.ts
```diff
@@ -11,6 +11,9 @@ import { renderToString } from 'react-dom/server';
 import bodyParser from 'body-parser';
 import fetch from 'node-fetch';
 
+import Page from 'components/shared/page';
+import { getConfigValue } from 'server/ssmConfig';
+import { CapiError, capiEndpoint, getContent } from 'capi';
```
JamieB-gu, line 16: Have these been imported twice?

src/components/standfirst.tsx
webb04: An example I came across was making the standfirst text white here https://www.theguardian.com/society/ng-interactive/2020/apr/29/how-humans-have-reacted-to-pandemics-through-history-a-visual-guide

This is apps-specific for backwards compatibility so hopefully this won't be needed.

src/client/interactive.ts
webb04: Possibly, this is for backwards compatibility but if we update our styles and classnames to match what dotcom-rendering do, we shouldn't need this.

src/components/shared/page.tsx
webb04: this would get updated by the client side code using `userAgent`. But I imagine most interactives have the same styles for `ios` and `android` so this would be an enhancement really to apply those styles straight away.

src/item.ts
webb04: the block type is `ContentAtom` in capi. Maybe we could name it `Atom`

JamieB-gu: We seem to refer to it mostly as "atom" so `Atom` sounds good.

webb04:
@JamieB-gu fixed the checks other than Snyk. Last question would be do we want to merge in the fallbacks like "className: 'ios'" to best support interactives now, or do we remove them with the idea of agreeing on a plan with dotcom/interactives going forward?

JamieB-gu:
> fixed the checks other than Snyk
This can be ignored since #312.

> do we want to merge in the fallbacks
Are we currently filtering out interactives in MAPI? If yes, I have a preference for removing the classes and coming up with a new solution. I think we should only add them in if not having them blocks us getting our first production articles out.

src/components/shared/page.tsx
```diff
@@ -156,6 +157,17 @@ function ArticleBody({ capi, imageSalt, getAssetLocation }: BodyProps): ElementW
     const liveblogScript = getAssetLocation('liveblog.js');
     const mediaScript = getAssetLocation('media.js');
 
+    if (item.design === Design.Interactive && item.display === Display.Immersive) {
```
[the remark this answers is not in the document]
webb04: Yes there is, they still expect page elements like the header to be there but because it's `interactive` the interactive atoms won't be iframed so their styles can override page styles.

This is starting to explain some of the issues we have. Our page structure is very different to dotcom (classnames and elements on the page). So interactive developers would have to add lots of unique styles for apps by targeting ".ios" or "android"


Immersive interactives should be more of a blank page, where the atoms provide their own header. standfirst etc.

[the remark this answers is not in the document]
webb04: Actually according to @duarte `interactive` articles are more common than `immersive interactives`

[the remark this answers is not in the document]
duarte: @JamieB-gu one thing to look out for is that many Immersive articles are not Interactive — so they're independent switches: Interactive is a composer file type, while Immersive is a display hint available to both standard articles and Interactives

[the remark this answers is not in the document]
JamieB-gu: So in the context of these changes, a piece of content that is `Design.Interactive` but _isn't_ `Display.Immersive` won't render as an interactive? @webb04

src/bodyElement.ts
```diff
@@ -215,18 +227,40 @@ const parse =
...
+
+            if (!atoms) {
+                return new Err('No atom data returned by capi')
+            }
+
+            const id = element.contentAtomTypeData?.atomId
+            const atom = atoms.interactives?.find(interactive => interactive.id === id);
+
+            if (!atom?.data?.interactive) {
+                return new Err('No atom matched')
```
JamieB-gu, line 240:
```suggestion
                return new Err(`No atom matched this id: ${id}`);
```

src/components/atoms/interactiveAtom.tsx
```diff
@@ -0,0 +1,38 @@
...
+const InteractiveAtom: FC<InteractiveAtomProps> = (props: InteractiveAtomProps): ReactElement => {
+    const { html, styles, js, format } = props;
+    const pillarStyles = getPillarStyles(format.pillar);
+    const style = h('style', { dangerouslySetInnerHTML: { __html: styles } });
+    const script = js ? h('script', { dangerouslySetInnerHTML: { __html: js } }) : null;
+    const markup = styledH('figure', { css: InteractiveAtomStyles(pillarStyles), dangerouslySetInnerHTML: { __html: html } });
+    return <>
+        {style}
+        {script}
+        {markup}
```
JamieB-gu, line 34: In case the script relies on an element being present?

```suggestion
        {markup}
        {script}
```

webb04: good spot! although I think I did this for backwards compatibility. See the styles targeting h2's after script tags...

Screen Shot 2020-05-07 at 16 03 11

PR #299, opened 2020-04-27
Tidy Server Module

JamieB-gu:
## Why are you doing this?

Tidying up the server module before adding some features. Thought this was preferable to causing large diffs on more impactful changes. There shouldn't be any change in functionality.

## Changes

- Tidied up the server module

### DIFF
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -1,13 +1,17 @@
 // ----- Imports ----- //
 
 import path from 'path';
-import express, {NextFunction, Request, Response as ExpressResponse} from 'express';
+import express, {
+    NextFunction,
+    Request,
+    Response as ExpressResponse,
+} from 'express';
 import compression from 'compression';
 import { renderToString } from 'react-dom/server';
 import bodyParser from 'body-parser';
 import {
-  BufferedTransport,
-  CompactProtocol
+    BufferedTransport,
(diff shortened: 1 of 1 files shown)


src/server/server.ts
```diff
@@ -22,119 +26,125 @@ import { App, Stack, Stage } from './appIdentity';
...
-    const duration = Date.now() - start;
-    logger.info(`HTTP ${request.method} ${request.path} returned ${response.statusCode} in ${duration}ms`)
-  });
 
-  next();
-});
+const app = express();
 
+app.use(bodyParser.raw({ limit: '50mb' }));
 app.use('/assets', express.static(path.resolve(__dirname, '../assets')));
```
alexduf, line 134: I thought this was used to serve assets locally?

src/server/server.ts
```diff
@@ -22,119 +26,125 @@ import { App, Stack, Stage } from './appIdentity';
...
 app.use('/assets', express.static(path.resolve(__dirname, '../dist/assets')));
 app.use(compression());
```
alexduf, line 136: That would probably be useful to keep :)

src/server/server.ts
JamieB-gu: Currently it serves the fonts locally, yes.

PR #300, opened 2020-04-28
render advertisementFeatures

webb04:
Render AdvertisementFeatures. Also known as Guardian Labs.

- Use textSans font
- Use labs colours rather than pillar colours
- Remove tags and follow options
- Added lab stories

## Screenshots

Screen Shot 2020-04-28 at 15 03 59

### DIFF
--- a/src/components/advertisementFeature/article.tsx
+++ b/src/components/advertisementFeature/article.tsx
@@ -1,0 +1,79 @@
+// ----- Imports ----- //
+
+import React, { ReactNode } from 'react';
+import { css } from '@emotion/core';
+import { neutral, background } from '@guardian/src-foundations/palette';
+import { from, breakpoints } from '@guardian/src-foundations/mq';
+
+import HeaderImage from 'components/headerImage';
+import Series from 'components/shared/articleSeries';
+import Headline from 'components/headline';
+import Standfirst from 'components/standfirst';
+import Body from 'components/shared/articleBody';
+import Metadata from 'components/standard/metadata';
+import { darkModeCss, articleWidthStyles } from 'styles';
+import { Keyline } from 'components/shared/keyline';
+import { Standard, Review, getFormat } from 'item';
+import { ImageMappings } from 'components/shared/page';
+
+
+// ----- Styles ----- //
+
+const Styles = css`
+    background: ${neutral[97]};
+`;
+
+const DarkStyles = darkModeCss`
+    background: ${background.inverse};
+`;
+
+const BorderStyles = css`
+    background: ${neutral[100]};
+    ${darkModeCss`background: ${background.inverse};`}
+
+    ${from.wide} {
+        width: ${breakpoints.wide}px;
+        margin: 0 auto;
+    }
+`;
+
+
+// ----- Component ----- //
+
+interface Props {
+    imageMappings: ImageMappings;
+    item: Standard | Review;
+    children: ReactNode[];
+}
+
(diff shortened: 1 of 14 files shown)


src/components/headline.tsx
```diff
@@ -76,6 +76,11 @@ const commentStyles = css`
     padding-bottom: ${remSpace[1]};
 `;
 
+const advertisementFeatureStyles = css`
+    ${styles}
+    ${textSans}
```
JamieB-gu, line 81: Would it be preferable to use a design system font here?

src/components/paragraph.tsx
```diff
@@ -1,24 +1,35 @@
...
 }
 
-const styles = css`
-    ${body.medium()}
-    overflow-wrap: break-word;
-    margin: 0 0 ${remSpace[3]};
-`;
+const styles = (design: Design): SerializedStyles => {
+    const advertisementFeature = design === Design.AdvertisementFeature
+        ? textSans
```
JamieB-gu, line 19: Would it be preferable to use a design system font here? E.g. `textSans.medium()`?

```ts
const font = design === Design.AdvertisementFeature ? textSans.medium() : body.medium();

return css`
    ${font}
    overflow-wrap: break-word;
    margin: 0 0 ${remSpace[3]};
`;
```

src/components/standfirst.tsx
```diff
@@ -76,6 +76,11 @@ const media = css`
     }
 `;
 
+const advertisementFeature = css`
+    ${styles}
+    ${textSans}
```
JamieB-gu, line 81: Would it be preferable to use a design system font here?

src/components/advertisementFeature/article.tsx
```diff
@@ -0,0 +1,79 @@
...
+        margin: 0 auto;
+    }
+`;
+
+
+// ----- Component ----- //
+
+interface Props {
+    imageMappings: ImageMappings;
+    item: Standard | Review;
```
JamieB-gu, line 45: Are these the only two types of articles that can be advertisement features?

src/components/headline.tsx
webb04: I had a problem here as I wanted to keep the headline the same size as other articles but that would require larger textSans sizes in the design system https://theguardian.design/2a1e5182b/p/930d69-typography/b/06df96

@SiAdcock

webb04: Would be happy to try and add that to the design system if it's useful? I imagine Dotcom could have the same problem when adding glabs

src/components/paragraph.tsx
webb04: `textSans.medium()` is a lot smaller than `headline.medium()`. Even the largest textSans is smaller that `headline.medium()`

src/components/headline.tsx
JamieB-gu: Is the issue that you need a `34px` version of `Text Sans`, and the largest currently available is `24px`?

webb04: yep, I could go into source and find the rem value used for that but the current way would also do that.

JamieB-gu: I've opened [this](https://github.com/guardian/source/issues/343) on Source.

webb04:
@JamieB-gu The larger `textSans` sizes have been merged into the design system so they should be available from the next release hopefully. 

I've seen us use the older `textSans` in other places to `AdvertisementFeatures` so I've made a ticket to move them all over to using the design system fonts https://theguardian.atlassian.net/browse/AR-25.

Should this be merged beforehand?

PR #302, opened 2020-04-29
use inverted colour for anchors on dark backgrounds

webb04:
## Changes

- Improve contrast of anchors on media pages 
- Use strong elements to keep inline display

## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |
|  |  |

### DIFF
--- a/src/components/anchor.tsx
+++ b/src/components/anchor.tsx
@@ -2,7 +2,7 @@
 
 import React, { FC, ReactNode } from 'react';
 import { SerializedStyles, css } from '@emotion/core';
-import { Format } from '@guardian/types/Format';
+import { Format, Design } from '@guardian/types/Format';
 import { neutral } from '@guardian/src-foundations/palette';
 
 import { darkModeCss } from 'styles';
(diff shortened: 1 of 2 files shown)


src/components/anchor.tsx
```diff
@@ -18,18 +18,32 @@ interface Props {
...
+            return css`
+                color: ${kicker};
+                border-bottom: 0.0625rem solid ${neutral[86]};
+            `
+    }
+}
+
 const Anchor: FC<Props> = ({ format, children, href, className }: Props) =>
-    <a css={[styles(getPillarStyles(format.pillar).kicker), className]} href={href}>
+    <a css={[styles, className, colour(format)]} href={href}>
```
JamieB-gu, line 46:
```suggestion
    <a css={[styles, colour(format), className]} href={href}>
```

src/components/figCaption.tsx
```diff
@@ -76,11 +77,11 @@ const captionElement = (format: Format) => (node: Node, key: number): ReactNode
     const children = Array.from(node.childNodes).map(captionElement(format));
     switch (node.nodeName) {
         case 'STRONG':
-            return <h2 css={captionHeadingStyles} key={key}>{children}</h2>;
+            return <strong css={captionHeadingStyles} key={key}>{children}</strong>;
```
JamieB-gu, line 80: The choice of `h2` here was for semantic reasons (plus they're `h2` elements on dotcom). If the aim is to get inline styles can we use `display: inline` instead?

frankie297:
You might want to check with @benwuersching but I'm pretty sure we aren't supposed to be using pillar colours for anchors anymore? I think it's supposed to be the same colour as the article copy with an underline (same colour)

frankie297:
Like this....
image

benwuersching:
@webb04 @JamieB-gu @frankie297 For a general look at the design of dark mode media pages (and links on them), please see here:

Galleries:
https://app.zeplin.io/project/58ee27a3e08030c8234f88a3/dashboard?seid=5d67f6e1f555509eedb2b361

Podcasts:
https://app.zeplin.io/project/58ee27a3e08030c8234f88a3/dashboard?seid=5d694f5ff555509eedc232e4

Videos:
https://app.zeplin.io/project/58ee27a3e08030c8234f88a3/dashboard?seid=5d67f67646e50e9b4232f97d

webb04:
Good spot, I know it's like that on dark mode. Should be easy for us to change either way

| Current app | Current app dark mode |
| --- | --- |
|  |  |

frankie297:
> For a general look at the design of dark mode media pages (and links on them), please see here:

@benwuersching Does this only apply to Dark Mode then?

benwuersching:
@frankie297 @webb04 Ah yes! In light mode, we do want to keep the pillar colour for links/anchors etc. if currently used. If we currently use any colours that are hard to read on black background, then let's find a more suitable colour. It should come, however, from the range of options within the pillar in question.

webb04:
Is this ok to go @frankie297 @JamieB-gu.

I don't think we have any dark mode styles yet so I'll save them for another ticket.

PR #303, opened 2020-04-30
Split Up Item Module

JamieB-gu:
## Why are you doing this?

The `Item` module is quite large and unfocused. This splits out some of the data types into more tightly focused modules of their own.

These changes will help to enable some upcoming work on the liveblog API.

## Changes

- Created `bodyElement` module
- Created `liveBlock` module
- Created `docParser` module to house the `DocParser` type
- Migrated the CAPI date parser to the `capi` module
- Updated imports

### DIFF
--- a/src/item.ts
+++ b/src/item.ts
@@ -13,8 +13,8 @@ import {
 } from 'mapiThriftModels';
 import { Format, Pillar, Design, Display } from 'format';
 import { Image as ImageData, parseImage } from 'image';
-import { LiveBlock, Body as LiveBody, parseLiveBlocks } from 'liveBlock';
-import { parseElements } from 'bodyElement';
+import { LiveBlock, parseLiveBlocks } from 'liveBlock';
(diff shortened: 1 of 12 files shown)


src/item.ts
```diff
@@ -115,20 +41,20 @@ interface Liveblog extends Fields {
 
 interface Review extends Fields {
     design: Design.Review;
-    body: Body;
+    body: LiveBody;
```
webb04, line 44: Why `LiveBody` this implies Liveblog body for me

JamieB-gu:
> Interesting to see Chromatic flag two captions changes?

Yeah not sure why that is, maybe they're left over from a previous change?

JamieB-gu:
The Snyk issue is not related to this PR; there are no dependency changes.

src/item.ts
JamieB-gu: As discussed offline, you're right this is a mistake 🙈. Fixed now.

PR #307, opened 2020-05-04
test format type

webb04:
## Changes

- Add tests for `format`
- Increase test coverage minimums to 75%

### DIFF
--- a/src/item.test.ts
+++ b/src/item.test.ts
@@ -636,14 +636,6 @@ describe('video elements', () => {
 });
 
 describe('format', () => {
-    test('get format returns correct fields', () => {
-        const item = f(immersive);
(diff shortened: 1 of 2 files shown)


src/item.test.ts
```diff
@@ -565,4 +633,26 @@ describe('video elements', () => {
...
     });
+});
+
+describe('format', () => {
+    test('get format returns correct fields', () => {
+        const item = f(immersive);
+        const format = getFormat(item);
+        expect('design' in format).toBe(true);
+        expect('display' in format).toBe(true);
+        expect('pillar' in format).toBe(true);
```
JamieB-gu, line 644: I think this *should* be caught by the compiler, as `getFormat` returns `Format`.

PR #308, opened 2020-05-04
npm audit fix

webb04:


webb04:
This seems to be the problem: https://github.com/lodash/lodash/pull/4745

JamieB-gu:
Is this covered by #304?

webb04:
So the vulnerability hasn't been patched yet in Lodash so neither npm audit or Snyk will be able to patch it. I've opened this to unblock open PRs https://github.com/guardian/apps-rendering/pull/309

JamieB-gu:
Snyk offers their own patches for situations where the fix hasn't been implemented in the dependency yet. It looks like that's what #304 has applied.

webb04:
I see, it's not promising that the Snyk check is failing on that PR still

JamieB-gu:
I *think* that's because `snyk protect` isn't being run as part of the GitHub Action, but I'm not 100% on that, looking into it now.

PR #309, opened 2020-05-04
remove snyk check on pr

webb04:
This seems to be the problem: https://github.com/lodash/lodash/pull/4745.

We can add this check back once there's a fix.


alexduf:
I'm not sure I understand, we're disabling snyk on master so it doesn't block us?
Shouldn't we make snyk non blocking instead?

PR #310, opened 2020-05-04
add font smoothing to page styles

webb04:
## Screenshots

| Before | After |
| --- | --- |
|  |  |


JamieB-gu:
Would you mind adding a bit to the description to give some background to this?

sndrs:
this is worth checking in dark mode too. 

I can never remember which looks better for which (`antialias` or `subpixel-antialiased`/`default`), but light on dark usually seems to work better with the opposite setting to dark on light, in my experience.

webb04:
Closing as pages won't be viewed on `Mac OS X/macOS.` very much. https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth

PR #313, opened 2020-05-06
Editorial Palette

JamieB-gu:
## Why are you doing this?

A speculative PR based on our conversation about an editorial colour palette. I'm not sure about the API here so I'd appreciate some feedback 🙂

This creates some simple functions to lookup headline font and background colours, and the border colour based on the `Format` type. I've also added some functions to generate the next level above this, i.e. CSS declarations for font and background colour based on the `Format` type.

## Changes

- Created editorialPalette module
- Functions to get headlineFont, headlineBackground and border colours
- Functions to apply color and background declarations
- Applied new palette functions to `Headline` component

### DIFF
--- a/src/components/headline.tsx
+++ b/src/components/headline.tsx
@@ -10,7 +10,8 @@ import { Format, Display, Design } from '@guardian/types/Format';
 import { Item } from 'item';
 import { wideContentWidth, articleWidthStyles } from 'styles';
 import StarRating from 'components/starRating';
-import { headlineFontColour, headlineBackgroundColour, border } from 'editorialPalette';
+import { border } from 'editorialPalette';
+import { headlineFontColour, headlineBackgroundColour } from 'editorialStyles';
 
 
 // ----- Component ----- //
--- a/src/components/headline.tsx
+++ b/src/components/headline.tsx
@@ -11,7 +11,7 @@ import { Item } from 'item';
 import { wideContentWidth, articleWidthStyles } from 'styles';
 import StarRating from 'components/starRating';
 import { border } from 'editorialPalette';
-import { headlineFontColour, headlineBackgroundColour } from 'editorialStyles';
+import { headlineTextColour, headlineBackgroundColour } from 'editorialStyles';
 
 
 // ----- Component ----- //
@@ -22,7 +22,7 @@ interface Props {
 
 const styles = (format: Format): SerializedStyles => css`
     ${headline.medium()}
-    ${headlineFontColour(format)}
+    ${headlineTextColour(format)}
     ${headlineBackgroundColour(format)}
     padding-bottom: ${remSpace[9]};
     margin: 0;
--- a/src/components/headline.tsx
+++ b/src/components/headline.tsx
@@ -62,7 +62,7 @@ const analysisStyles = (format: Format): SerializedStyles => css`
     ${headline.medium({ lineHeight: 'regular', fontWeight: 'light' })}
 
     span {
-        box-shadow: inset 0 -0.1rem ${border(format).light};
+        box-shadow: inset 0 -0.1rem ${border.primary(format)};
         padding-bottom: 0.2rem;
     }
 `;
--- a/src/components/headline.tsx
+++ b/src/components/headline.tsx
@@ -3,15 +3,14 @@
 import React, { ReactElement } from 'react';
 import { css, SerializedStyles } from '@emotion/core';
 import { headline, textSans } from '@guardian/src-foundations/typography';
-import { background, neutral, text } from '@guardian/src-foundations/palette';
 import { remSpace } from '@guardian/src-foundations';
 import { between, from } from '@guardian/src-foundations/mq';
-import { Display, Design } from '@guardian/types/Format';
+import { Format, Display, Design } from '@guardian/types/Format';
 
 import { Item } from 'item';
-import { darkModeCss as darkMode, wideContentWidth, articleWidthStyles } from 'styles';
-import { getPillarStyles, PillarStyles } from 'pillarStyles';
+import { wideContentWidth, articleWidthStyles } from 'styles';
 import StarRating from 'components/starRating';
+import { headlineFontColour, headlineBackgroundColour, border } from 'editorialPalette';
 
 
 // ----- Component ----- //
@@ -20,26 +19,18 @@ interface Props {
     item: Item;
 }
 
-const darkStyles = darkMode`
-    background: ${background.inverse};
-    color: ${neutral[86]};
-`;
-
-const styles = css`
+const styles = (format: Format): SerializedStyles => css`
     ${headline.medium()}
+    ${headlineFontColour(format)}
+    ${headlineBackgroundColour(format)}
     padding-bottom: ${remSpace[9]};
-    color: ${text.primary};
     margin: 0;
 
     ${articleWidthStyles}
-
-    ${darkStyles}
 `;
 
 const immersiveStyles = css`
     ${headline.medium({ fontWeight: 'bold' })}
-    background-color: ${neutral[7]};
-    color: ${neutral[100]};
     font-weight: 700;
     padding: ${remSpace[1]} ${remSpace[2]} ${remSpace[6]} ${remSpace[2]};
     margin: calc(100vh - 5rem) 0 0;
@@ -64,67 +55,52 @@ const immersiveStyles = css`
             width: ${wideContentWidth}px;
         }
     }
-
-    ${darkStyles}
 `;
 
-const analysisStyles = ({ kicker }: PillarStyles): SerializedStyles => css`
-    ${styles}
+const analysisStyles = (format: Format): SerializedStyles => css`
     ${headline.medium({ lineHeight: 'regular', fontWeight: 'light' })}
 
     span {
-        box-shadow: inset 0 -0.1rem ${kicker};
+        box-shadow: inset 0 -0.1rem ${border(format).light};
         padding-bottom: 0.2rem;
     }
 `;
 
-const mediaStyles = css `
-    ${styles}
-    color: ${neutral[100]};
+const mediaStyles = css`
     ${headline.medium({ fontWeight: 'medium' })}
-    
-    ${darkMode`
-        color: ${neutral[86]};
-    `}
 `
 
-const featureStyles = ({ featureHeadline }: PillarStyles): SerializedStyles => css`
-    ${styles}
+const featureStyles = css`
     ${headline.medium({ fontWeight: 'bold' })}
-    color: ${featureHeadline};
 `;
 
 const commentStyles = css`
-    ${styles}
     ${headline.medium({ fontWeight: 'light' })}
     padding-bottom: ${remSpace[1]};
 `;
 
 const advertisementFeatureStyles = css`
-    ${styles}
     ${textSans.xxxlarge({ lineHeight: 'regular' })}}
 `;
 
-const getStyles = (item: Item): SerializedStyles => {
-    const pillarStyles = getPillarStyles(item.pillar);
-
-    if (item.display === Display.Immersive) {
-        return immersiveStyles;
+const getStyles = (format: Format): SerializedStyles => {
+    if (format.display === Display.Immersive) {
+        return css(styles(format), immersiveStyles);
     }
 
-    switch (item.design) {
+    switch (format.design) {
         case Design.Analysis:
-            return analysisStyles(pillarStyles);
+            return css(styles(format), analysisStyles(format));
         case Design.Feature:
-            return featureStyles(pillarStyles);
+            return css(styles(format), featureStyles);
         case Design.Comment:
-            return commentStyles;
+            return css(styles(format), commentStyles);
         case Design.Media:
-            return mediaStyles;
+            return css(styles(format), mediaStyles);
         case Design.AdvertisementFeature:
-            return advertisementFeatureStyles;
+            return css(styles(format), advertisementFeatureStyles);
         default:
-            return styles;
+            return styles(format);
     }
 }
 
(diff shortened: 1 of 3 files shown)


JamieB-gu:
As discussed offline, will look into making this an object.

JamieB-gu:
Following discussions with @oliverlloyd it looks like creating an object keyed on `Format` would be complex and come with a lot of boilerplate. Here's a small sample of what that might look like:

```ts
const defaultPalette = {
    headlineFont: {
        light: text.primary,
        dark: neutral[86],
    },
};

const immersiveHeadlineFont = {
    ...defaultPalette.headlineFont,
    light: neutral[100],
};

const palette: { [ k: string ]: typeof defaultPalette } = {
    [ [ Design.Feature, Display.Immersive, Pillar.News ].toString() ]: {
        ...defaultPalette,
        headlineFont: immersiveHeadlineFont,
    },
    [ [ Design.Feature, Display.Immersive, Pillar.Opinion ].toString() ]: {
        ...defaultPalette,
        headlineFont: immersiveHeadlineFont,
    },
    [ [ Design.Feature, Display.Immersive, Pillar.Sport ].toString() ]: {
        ...defaultPalette,
        headlineFont: immersiveHeadlineFont,
    },
    [ [ Design.Feature, Display.Immersive, Pillar.Culture ].toString() ]: {
        ...defaultPalette,
        headlineFont: immersiveHeadlineFont,
    },
    [ [ Design.Feature, Display.Immersive, Pillar.Lifestyle ].toString() ]: {
        ...defaultPalette,
        headlineFont: immersiveHeadlineFont,
    },
}
```

Instead, I've looked at addressing Oliver's other concern with exporting CSS: I've split out any CSS-related functions into an `editorialStyles` module. The `editorialPalette` module now just export functions that generate hex strings.

I've also added a brief example of a `Palette` type, which we may or may not want to use. Let me know what you think @oliverlloyd @SiAdcock @gtrufitt @liywjl @webb04.

src/editorialPalette.ts
```diff
@@ -80,32 +83,20 @@ const border = (format: Format): Colour => {
...
-    `;
-
-const headlineFontColour = compose(fontColour, headlineFont);
-const headlineBackgroundColour = compose(backgroundColour, headlineBackground);
+const palette = (format: Format): Palette =>
+    ({
+        headlineFont: headlineFont(format),
+        headlineBackground: headlineBackground(format),
+        border: border(format),
+    });
```
oliverlloyd, line 91: _This_. It's funny how it looks so simple when you see it here now but I guess the solutions that seem simple are the best ones!

One nitpick, I wouldn't name the function the same as the property it is setting, I normally go for some kind of prefix like `decideHeadlineBackground()` or `getHeadlineBackground()` but that's maybe because I lean a lot on using `CMD+D` to select similar text when editing. Defo not a blocker.

src/editorialPalette.ts
```diff
@@ -0,0 +1,102 @@
...
+} from '@guardian/src-foundations/palette';
+import { Format, Design, Display, Pillar } from '@guardian/types/Format';
+
+
+// ----- Types ----- //
+
+interface Colour {
+    light: string;
+    dark: string;
+}
```
oliverlloyd, line 21: This might be a bit restrictive for DCR. I don't have a specific example but my feeling is we often stray outside the bounds of 'neat'...

liywjl: I am assuming this is for dark mode?

JamieB-gu: > I am assuming this is for dark mode?

Yep, many colours have a dark mode variant, the most prominent examples being the article background and body copy colour.

src/editorialPalette.ts
JamieB-gu: I suspect DCR might make use of this (with a theme), and AR might make use of the individual functions/CSS declarations.

gtrufitt:
Love it, I'd be really keen to start using this and evolve it together!

src/editorialPalette.ts
```diff
@@ -0,0 +1,102 @@
...
+
+interface Palette {
+    headlineFont: Colour;
```
SiAdcock, line 24: Minor: I might suggest `headlineText` for better consistency with Source.

JamieB-gu: Good idea! Updated.

SiAdcock: It's worth spending a bit of time thinking about how the naming system will scale. I know @oliverlloyd has pointed to the potential problem with the `Colour` abstraction defining a `light` and a `dark` mode. This could become a bit verbose or restrictive for DCR that only has a light mode. There might be ways for DCR to abstract this away in their application, but it's worth bearing in mind.

I'll give you a peek at Source's colour token naming system:

```
[theme][property].[component][variant][state][mode]
```

- theme: the colour theme to which this token applies
- property: the style property for which this token is intended
- component: the design system component to which this token is applied
- variant: there are sometimes multiple variants of each token. For example, text in a component may be primary (core) or secondary (supporting)
- state: user interaction that may have affected the appearance, such as hover or checked
- mode: the colour mode (i.e. inverse represents dark mode)

Note that in most cases, not all of these parts will be present.

Examples:

```
text.primary
```

This is a text colour, for primary text. There is no theme specified, so we assume it’s for the default theme. Note that it’s not for a specific component. “Primary” is the variant, telling us it should be applied to the most important or core text. There is no state, so we assume there has been no user interaction. There is no mode, so we assume the default colour mode ("light mode")

```
brandBackground.cardArticleHover
```

It is a background colour on brand theme, for the article card, in the hover state.

JamieB-gu: This is very thorough, thanks!

Two questions:

1. What does the dot signify in the hierarchy (I know it represents a method or property lookup, but why at that level)?
2. How would you define the **property** for text in a headline? Would it be `headlineText.primary`?

SiAdcock: The position of the dot was my best guess at where to split the palette to make it more treeshakable. It's likely that most applications won't be using multiple themes on one page, and those that used multiple themes might not want all the colours for each theme. I think you can ignore this consideration for your purpose.

In your example `text` is the property... it represents the CSS property to which this toekn will be applied (well, technically it's the CSS `color` property, but `color` is a bit vague!) 

I think `headlineText` is a good name. Abstracted it's `[component][property]` whereas Source goes with `[property][component]` but I don't think it matters too much.

There's only one headline per article, so you won't have variants like "primary headline" and "supporting headline".

It's worth taking a look at other design systems to see how they name their tokens. Then write down what your system is, so you can (attempt to!) keep it consistent:

- [Salesforce](https://www.lightningdesignsystem.com/design-tokens/)
- [Adobe](https://spectrum.adobe.com/page/design-tokens/)
- [Duet](https://www.duetds.com/tokens/)

JamieB-gu: > There's only one headline per article

Except when there [isn't](https://www.theguardian.com/technology/2020/jan/20/jabra-elite-75t-review-small-and-long-lasting-airpod-beaters) 🙂

I'm happy to use `text.headlinePrimary` where `headline` is the **component**. I derived from what you said here:

> the _design system component_ to which this token is applied

that this was meant to refer to specific components that exist in the design system. If not, and it won't cause confusion, let's follow the same conventions?

src/components/headline.tsx
```diff
@@ -3,14 +3,14 @@
...
-import { background, neutral, text } from '@guardian/src-foundations/palette';
 import { remSpace } from '@guardian/src-foundations';
+import { Format } from '@guardian/types/Format';
 
 import { Item } from 'item';
-import { darkModeCss as darkMode } from 'styles';
-import { getPillarStyles, PillarStyles } from 'pillarStyles';
 import StarRating from './starRating';
 import { Display, Design } from 'format';
+import { border } from 'editorialPalette';
```
SiAdcock, line 12: Do you intend to move `border` to `editorialStyles` also?

src/editorialPalette.ts
SiAdcock: Yep that sounds fine to me! I'd be interested to hear what people think of the ergonomics of splitting imports on CSS property

```js
import { text } from 'editorialPalette'

const headline = css`
  color: ${text.headlinePrimary(format)};
`

const standfirst = css`
  color: ${text.standfirstPrimary(format)};
`
```

vs

```js
import { textHeadline, textStandfirst } from 'editorialPalette'

const headline = css`
  color: ${textHeadline(format)};
`

const standfirst = css`
  color: ${textStandfirst(format)};
`
```

The second API is more verbose but perhaps more explicit? It would be so cool if you could do away with all those function calls, but I understand it might become difficult to maintain.

```js
import { palette } from 'editorialPalette'

const { text } = palette(format)

const headline = css`
  color: ${text.headlinePrimary};
`

const standfirst = css`
  color: ${text.standfirstPrimary};
`
```

src/components/headline.tsx
JamieB-gu: I wasn't planning to _move_ it, but we could generate CSS declarations for it there as well. Would have to be `border-color` specifically rather than `border` I think?

Now that you've flagged this it occurs to me we could also name it something more generic in the `palette` module, like `line`?

src/editorialPalette.ts
JamieB-gu: In your second example wouldn't you still need a reference to `primary`?

I don't have a strong preference for one of those APIs over the other. I like two things about both:

- They take `Format` and give you back one or more colours.
- You can consume individual colours, _or_ retrieve the entire palette as an object. I think this will be important for flexibility (see [this comment](https://github.com/guardian/apps-rendering/pull/313#discussion_r421421338)).

src/components/headline.tsx
SiAdcock: > Would have to be border-color specifically rather than border I think

Yeah, maybe. I was gonna suggest dropping `colour` or `color` from names because colour is implied by the fact it's in the editorial palette. It's up to you though, maybe there will be non-colour tokens here also (spacing, border radiuses)? In which case, specific is better.

> we could also name it something more generic in the palette module, like line

I think there's a distinction between `border` and `line`. Borders border elements, lines divide elements. Borders are encapsulated within a component, lines are defined at the layout level.

JamieB-gu: > I was gonna suggest dropping `colour` or `color` from names because colour is implied by the fact it's in the editorial palette

Sorry, I was probably unclear there. I was thinking something along these lines:

1. The `editorialPalette` module would contain a colour token, called either `border` or `line`, which is _just_ a hex colour.

2. The `editorialStyles` module could contain a CSS declaration of the form:

```ts
const border = css`
  border-color: ${palette.border};
`;
```

When I said `border-color` I meant the CSS property rather than the name for the variable in our code 🙂. I think you're right, this is _just_ for colours.

> Borders border elements, lines divide elements

Ok, we have a semantic distinction between them so we should provide different names to avoid confusion, sounds good 👍.

src/editorialPalette.ts
SiAdcock: > In your second example wouldn't you still need a reference to primary?

Yep, sorry, an omission.

Cool. You're right, it's probably much of a muchness and you can potentially provide both APIs for flexibility

```js
import { p } from 'palette'
const { text } = p(format)

//...
css`
  color: ${text};
`
```

```js
import { text } from 'palette'

// ...
css`
  color: ${text(format)};
`
```

src/components/headline.tsx
SiAdcock: Right, yeah, me got confused there! Only 3 cups of tea this morning. You makes a lot of sense 👍

src/editorialPalette.ts
JamieB-gu: We seem to have a few ways we could define the API based upon what you've suggested above 🤔

1. Object with all the colours for a given **property**, with keys for every **component**-**variant** combination.
```ts
text(format).headlinePrimary
```

2. Object with all the colours for a given **property**, with keys for every **component** that give you objects with all the **variant**s
```ts
text(format).headline.primary
```

3. An object for each **property**-**component** combination, where the keys of each object are the **variant**s
```ts
textHeadline(format).primary
```

4. Function for every **property**-**component**-**variant** combination, which return a single hex colour as a `string`
```ts
textHeadlinePrimary(format)
```

SiAdcock: Can we also consider:

5. A function for the entire **palette** that returns an object with each **property**, each of which has keys for every **component-variant** combination

```js
palette(format).text.headlinePrimary
```

?

JamieB-gu: Oh sorry, yes that's also an API we'd want. With the above I was asking _which one_ of those four we should go with for individual colour lookups. I think we'd want the `palette` API to exist _in addition_ to that?

SiAdcock: My preference would be option 1. It's the same API as Source and I kinda like the ergonomics. Also there's some value in consistency as people would be able to switch between the Source palette and the Editorial palette without having to think too much.

JamieB-gu: Ok, sounds good, I'm happy with the API.

My only concern is that we'd be generating a lot of unused colours with every call to the `text` function. For example, if we use it in the `Headline` component, we're only going to pull out one or two colours. We could solve this with memoisation, or by dependency injecting the full palette as DCR are planning to do.

I wonder if a _fifth_ option that looks a lot like (1) would be preferable:

```ts
text.headlinePrimary(format)
```

SiAdcock: Oh gosh, yes. More performant and preserves the ergonomics. I'm on board 🚢

JamieB-gu: Ok, I've updated the code to use this new API.

src/editorialPalette.ts
```diff
@@ -0,0 +1,137 @@
...
+    };
+    background: {
+        headlinePrimary: string;
+        headlineInverse: string;
+    };
+    border: {
+        primary: string;
+        inverse: string;
+    };
+}
```
alexduf, line 33: Shouldn't these be declared using the `Colour` type?

PR #314, opened 2020-05-07
AR-18: Add anchor storybook component

frankie297:
## Why are you doing this?
Adds a storybook component for Anchors

### DIFF
--- a/src/components/anchor.stories.tsx
+++ b/src/components/anchor.stories.tsx
@@ -1,0 +1,41 @@
+// ----- Imports ----- //
+
+import React, {FC} from 'react';
+import Anchor from "./anchor";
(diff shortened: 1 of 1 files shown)


frankie297:
@JamieB-gu I just had to fix a linting issue where you can't pass children in as props, it would be good if you could just take another look and make sure what i am doing is not crazy 😄

JamieB-gu:
> I just had to fix a linting issue where you can't pass children in as props

Looks good.

If the `<span>` is there to wrap the `children`, I think you may be able to use `{copy()}` directly or, if that causes you problems, wrap it in a [fragment](https://reactjs.org/docs/fragments.html): `<>{copy()}</>`.

PR #315, opened 2020-05-07
dark mode colours

webb04:
## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |

### DIFF
--- a/src/components/bodyImage.tsx
+++ b/src/components/bodyImage.tsx
@@ -8,6 +8,7 @@ import { neutral } from '@guardian/src-foundations/palette';
 
 import Img from 'components/img';
 import { BodyImageProps } from 'image';
+import { darkModeCss } from 'styles';
 
 // ----- Setup ----- //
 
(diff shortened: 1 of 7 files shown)


src/components/bodyImage.tsx
```diff
@@ -29,6 +30,11 @@ const imgStyles = (width: number, height: number): SerializedStyles => css`
     ${from.phablet} {
         height: calc(620px * ${height / width});
     }
+
+    ${darkModeCss`
+        color: ${neutral[60]};
+        background-color: ${neutral[20]};
+    `}
```
JamieB-gu, line 37: Are these already covered by the changes you've made to the styles within the `Img` component?

webb04: I think header image might be using `Img`

JamieB-gu: Yeah it is, they both are.

As you've applied these styles within the `Img` component, are they needed here?

PR #320, opened 2020-05-13
fence interactive atom

webb04:
iframed interactive atoms were'nt working. It looks like they needed some extra css/js.

- https://github.com/guardian/frontend/blob/master/common/app/views/fragments/atoms/interactive.scala.html
- https://github.com/guardian/frontend/blob/88cfa609c73545085c3e5f3921631ec344a3eb83/common/app/templates/inlineJS/nonBlocking/interactiveFonts.scala.js

- https://github.com/guardian/frontend/blob/88cfa609c73545085c3e5f3921631ec344a3eb83/common/app/templates/inlineJS/nonBlocking/interactiveResize.scala.js

Example of fenced interactive atom e.g. iframed interactive atom in an article that isn't `type: interactive`: http://localhost:8080/world/2020/may/13/coronavirus-real-care-home-death-toll-double-official-figure-study-says

| Before | After |
| --- | --- |
|  |  |


src/renderer.ts
```diff
@@ -419,9 +419,40 @@ const render = (format: Format, imageMappings: ImageMappings, excludeStyles = fa
             const { html, css: styles, js } = element;
 
             const atom = h(InteractiveAtom, { html, styles, js, format });
+            const fenced = `
+                <html>
+                    <head>
+                        <meta charset="utf-8">
+                        <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
+                        <script>
+                            const fonts = [].slice.apply(window.parent.document.styleSheets)
```
JamieB-gu, line 428: Does this exist to convert a `StyleSheetList` to an `Array`? Would it be preferable to use `Array.from`?

src/renderer.ts
```diff
@@ -419,9 +419,40 @@ const render = (format: Format, imageMappings: ImageMappings, excludeStyles = fa
...
+                                .filter(sheet => sheet.ownerNode.attributes?.getNamedItem('data-emotion-css')?.value?.includes('PageStyles'))
+                                .map(sheet => sheet.ownerNode.textContent)
+                                .join(' ')
+                            var css = document.createElement('style');
+                            css.textContent = fonts;
+                            document.head.appendChild(css);
```
JamieB-gu, line 434: Is the purpose of this is to make the fonts available in the iframe? If so, since we _have_ the list of `@font-face` rules available on the server, would it make sense to insert them here directly rather than relying on client-side JS?

src/renderer.ts
```diff
@@ -419,9 +419,40 @@ const render = (format: Format, imageMappings: ImageMappings, excludeStyles = fa
...
+                        </script>
+                        <style>
+                            ${styles}
+                        </style>
+                    </head>
+                    <body>
+                        ${html}
+                        <script>
+                            ${js}
+                            function resize() {
```
JamieB-gu, line 444: What does this do out of curiosity 🤔?

src/renderer.ts
```diff
@@ -419,9 +419,40 @@ const render = (format: Format, imageMappings: ImageMappings, excludeStyles = fa
...
+                    </body>
+                </html>
+            `;
 
             if (format.design !== Design.Interactive) {
-                return h('iframe', null, atom);
+                return h('iframe', { srcdoc: fenced });
             } else {
                 return atom;
             }
```
JamieB-gu, line 458: Am I correct in thinking that the `atom` value is only generated for full-page interactives, and the `fenced` value is generated in all other cases? If so, would it make sense to make their generation part of this `if` condition so that we avoid doing additional work?

```ts
if (format.design === Design.Interactive) {
    return h(InteractiveAtom, { html, styles, js, format });
}

const fenced = ...;
return h('iframe', { srcdoc: fenced });
```

src/renderer.ts
webb04: It resizes the iframed interactive atom to the correct height. Otherwise it would default to the standard 150px iframe height

src/renderer.ts
webb04: Good idea, that makes sense

PR #326, opened 2020-05-18
Parser Converts Image URLs

JamieB-gu:
## Why are you doing this?

This is the groundwork for some further changes coming in later PRs. I've split the work up in this way in an attempt not to change too much at once, as this results in very large PRs. This follows on in turn from some other work in #288, #303 and #322.

In brief, this PR:

- Paves the way for the upcoming liveblog API
- Enables some work to allow storybook stories with images
- Should allow us to drop `sig-ignores-params`, which is the source of some complex VCL in Fastly and a potential cause of cache-splitting

A big part of this PR is a conceptual shift in the way we handle image URLs. Currently we're passing the URLs CAPI provides us all the way through to the components, and letting them handle re-writing them into image resizer URLs. With these changes the _parser_ will now handle translating CAPI image URLs into image resizer URLs. The components will now take these formatted URLs and focus on building them into markup. This follows from the principle that the parsing step exists to translate data from the structure provided by CAPI into a structure that best fits our rendering code.

**Note:** In the interests of keeping this PR as small as possible (it's still not _that_ small) I've opted to temporarily break liveblog hydration rather than implement a big chunk of the new liveblog API to keep it working. This will be fixed when that work appears in a forthcoming PR.

FYI @oliverlloyd @gtrufitt @philmcmahon 

## Changes

- Parser now produces Image Resizer URLs
- Removed `ImageMappings`
- Used `Context` to wrap `DocParser` and `salt`
- Contributor module now handle parsing of `Contributor`s from CAPI
- Temporarily breaks liveblog hydration with mock salt

### DIFF
--- a/src/components/liveblog/avatar.tsx
+++ b/src/components/liveblog/avatar.tsx
@@ -1,20 +1,18 @@
 // ----- Imports ----- //
 
-import React from 'react';
+import React, { ReactElement } from 'react';
 import { css, SerializedStyles } from '@emotion/core';
 
 import { Contributor, isSingleContributor } from 'contributor';
-import { src } from 'image';
-import { ImageMappings } from 'components/shared/page';
 
 
 // ----- Styles ----- //
 
-const imageWidth = 68;
+const imageWidth = '4rem';
 
 const AvatarStyles = (bgColour: string): SerializedStyles => css`
-    width: ${imageWidth}px;
-    height: ${imageWidth}px;
+    width: ${imageWidth};
+    height: ${imageWidth};
     background-color: ${bgColour};
     border-radius: 100%;
     float: left;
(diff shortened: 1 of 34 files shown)


src/components/liveblog/avatar.tsx
```diff
@@ -1,20 +1,18 @@
...
 
 import { Contributor, isSingleContributor } from 'contributor';
-import { src } from 'image';
-import { ImageMappings } from 'components/shared/page';
 
 
 // ----- Styles ----- //
 
-const imageWidth = 68;
+const imageWidth = '4rem';
```
webb04, line 11: Any use of using the design system? Can't see a `remSpace[16]` though

src/components/opinion/cutout.tsx
```diff
@@ -31,24 +24,27 @@ const ImageStyles = css`
...
     const [contributor] = contributors;
 
-    if (isSingleContributor(contributors) && contributor.bylineLargeImageUrl) {
-        const imgSrc = src(imageMappings, contributor.bylineLargeImageUrl, imageWidth*3);
-        return (
-            <div css={[className, Styles]}>
-                <img css={ImageStyles} src={imgSrc} alt={contributor.webTitle}/>
-            </div>
-        );
+    if (!isSingleContributor(contributors)) {
```
webb04, line 33: These logic is duplicated in a few places, maybe a future cleanup

src/contributor.ts
```diff
@@ -1,22 +1,47 @@
...
     cs.length === 1
 
+const parseContributors = (salt: string, content: Content): Contributor[] =>
+    articleContributors(content).map(contributor => ({
+        id: contributor.id,
+        apiUrl: contributor.apiUrl,
+        name: contributor.webTitle,
+        image: fromNullable(contributor.bylineLargeImageUrl).fmap(url => ({
+            srcset: contributorSrcset(url, salt),
+            src: src(salt, url, 64),
```
webb04, line 36: why 64?

src/image.ts
```diff
@@ -65,18 +65,18 @@ function src(imageMappings: ImageMappings, input: string, width: number): string
         quality: defaultQuality.toString(),
         fit: 'bounds',
         'sig-ignores-params': 'true',
-        s: imageMappings[url.pathname],
+        s: sign(salt, url.pathname),
```
webb04, line 68: So this will be pulled out from the rehydration?

src/components/liveblog/avatar.tsx
JamieB-gu: My understanding of the spacing system in Source was that it was designed for margins and paddings rather than widths, but I could be wrong 🤔

src/components/opinion/cutout.tsx
JamieB-gu: Yeah good point.

src/contributor.ts
JamieB-gu: No strong reason, we need a size for the fallback image and that's one of the values in the list used for the `srcset`. Could be something else.

src/image.ts
JamieB-gu: This code won't appear in the client-side bundle, it only runs on the server.

PR #327, opened 2020-05-19
MSS-1378: Add a DPR value to the image url

frankie297:
## Why are you doing this?
Add a dpr value to `srcset` and reduce the quality of images to `45` to allow for retina displays

Worked on this with @faresite and @JamieB-gu


src/image.ts
```diff
@@ -62,7 +63,8 @@ function src(imageMappings: ImageMappings, input: string, width: number): string
 
     const params = new URLSearchParams({
         width: width.toString(),
-        quality: defaultQuality.toString(),
+        quality: width > 500 ? lowerQuality.toString() : defaultQuality.toString(),
+        dpr: width > 500 ? '2' : '1',
```
JamieB-gu, line 67: Does this assume that no image will ever be larger than 500 CSS pixels?

JamieB-gu:
> Is there a reason why we use a higher dpr for larger images? does it improve quality or reduce file sizes?

I'm not familiar with the consequences of the `dpr` param, but I think @paperboyo and @philmcmahon did some work to establish that the best balance of bandwidth/perceived quality was to drop the `quality` param to `45` when working with a `DPR` of `2x`.

The `srcset` attribute supports `DPR` values of higher than `2x`. Alternatively you can use [`resolution`](https://developer.mozilla.org/en-US/docs/Web/CSS/resolution) with media queries if you switch to `<picture>`/`<source>`, which I believe is what dotcom does.

paperboyo:
> is there a reason why we use a higher dpr for larger images? does it improve quality or reduce file sizes?

Yeah, the perceptible quality is higher, we lower quality parameter to not baloon the filesize too much (you can compress denser images more without much negative visual impact). Obviously, all this depends on device size and viewing distance…

> maybe even dpr: 3 if supported

Dotcom has `1` and `2`. I don’t think we need more as there are diminishing returns here (you wouldn’t be able to discern beyond a certain point). Also, we would need to painstakingly establish another quality value for any other dpr than 1 & 2 (otherwise the size might balloon).

In the ideal world with perfect vision, infinitely fast and quick network and free caching, we would send as many variations of an image as there are renders of it on readers’ physical devices (in real pixels). In reality that would be ☠️ . 

We’ve established drp and quality coefficients out of the 🎩 , but my naive ideal world scenario for doing it [is a tad bonkers](https://github.com/guardian/frontend/issues/4638#issuecomment-45886985).

paperboyo:
> Is there a reason why we use a higher dpr for larger images?

&

> Does this assume that no image will ever be larger than 500 CSS pixels?

I’m sorry, but I can’t read code. If the above is true (dpr variance applies only to images over certain size), then I don’t understand this, it’s not what dotcom does, nor does this make sense as far as I can tell or follows any established practice. This would create an inconsistency in image quality that I don’t think would look good (if anything, but I would still strongly advise against, it’s the smaller images that could be denser, not the other way around).

All imagery (dotcom has some holes here, but is pretty good) should undergo the same dpr/quality variance regime.

Feel free to ping me whenever I can be of any help (unfortunately not around code). There is [this lengthy chat](https://chat.google.com/room/AAAAOQXYrpo/8KbukywZBUc), that we didn’t follow on, but seeing Daily’s constant fight with image quality and projects introducing a secondary, client-side resizers to help with performance (sic!), I think the whole department could be better. I certainly would love to learn as I don’t even understand the difference between scrset and Picture Element (other than that the former works for `background-image` while the latter doesn’t… or haven’t 5 years ago). cc @philmcmahon

JamieB-gu:
> I don’t even understand the difference between scrset and Picture Element

@paperboyo [this](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images) is my go-to primer whenever I have to remember how responsive images work 🙂

paperboyo:
Thanks, @JamieB-gu. This link following from yours has a useful overview of strategies of picking how many different dprs it makes sense to support, it even mentions [basing it on frequency](https://cloudfour.com/thinks/responsive-images-101-part-9-image-breakpoints/#setting-image-breakpoints-based-on-most-frequent-requests) of requests. (we’ve done it more [this way](https://cloudfour.com/thinks/responsive-images-101-part-9-image-breakpoints/#winging-it-aka-matching-your-layout-breakpoints))

I also made a mistake thinking `srcset` is different to Picture Element. It’s `image-set` I meant.

JamieB-gu:
There are a number of changes in `contributor.ts` and `image.ts` that actually came from #326. Is it possible a merge/rebase has caused some problems somewhere?

src/components/img.ts
```diff
@@ -26,18 +30,45 @@ const styles = css`
...
+        h(Source, {
+            sizes,
+            srcSet: image.dpr2Srcset,
+            dpr: Dpr.Two,
+        }),
+        h(Source, {
+            sizes,
+            srcSet: image.srcset,
+            dpr: Dpr.One,
+        }),
```
JamieB-gu, line 61: It occurred to me that this could actually be simpler if we inlined it rather than creating the `Source` component:

```suggestion
        h('source', {
            sizes,
            srcSet: image.dpr2Srcset,
            media: '(-webkit-min-device-pixel-ratio: 1.25), (min-resolution: 120dpi)',
        }),
        h('source', {
            sizes,
            srcSet: image.srcset,
        }),
```

That way we can drop the `Source` component, the `mediaQuery` function, _and_ we no longer need import `Dpr`.

frankie297:
@JamieB-gu @webb04 I just want to write a test before I merge this PR, incase you was wondering why I was leaving it hanging.

frankie297:
@JamieB-gu @webb04 can you just review my latest change adding a unit test before i merge? :) thakns

frankie297:
> 👍 , although FYI that `srcset` function will actually take a `media.guim` URL from CAPI and _generate_ an `i.guim` URL

thanks for that, have changed it 👍

frankie297:
> Looks good, would be interesting to see the differences in image quality and size

I was going to add screenshots but they just look exactly the same so not sure they would really add much to the PR

paperboyo:
What would be the easiest way to see the results of this for someone with just a web browser?

JamieB-gu:
@paperboyo Teleporter has an option to open a dotcom page in apps-rendering

paperboyo:
Everything looks brilliant (as much as I can atest, that is). One thing is that dpr seems to not be explicitly expressed in the URL (as is in dotcom), but instead, the width change. Not sure if it’s got any practical consequences apart from for ppl inspecting URLs in the Network tab, though 😉 .

JamieB-gu:
> dpr seems to not be explicitly expressed in the URL

Do you mean the `dpr` query parameter? As far as I can tell that returns an image with a width double the one requested in the `width` query parameter, hence why the `w` values in the `srcset` are also double that `width` parameter:

```html
<source srcset="https://i.guim.co.uk/image.jpg?width=620&dpr=2 1240w" />
```

 We've opted instead to leave the `width` parameter matched up with the `w` values in `srcset`:

```html
<source srcset="https://i.guim.co.uk/image.jpg?width=1240 1240w" />
```

We think these two implementations *should* produce the same result.

paperboyo:
Me thinks so too. Thanks for the explanation! (maybe caching would be the only reason to keep URLs consistent, but apps have differnt salt and, maybe, different resolutions to dotcom, so this may not apply)

JamieB-gu:
> but apps have different salt

Do they?

> different resolutions to dotcom

In apps-rendering at least, the choice of resolutions is arbitrary. They could match dotcom if we decided that would be useful.

paperboyo:
Above considerations are only about cache, I think? Less fragmentation should be better, no? If salt is the same: great!

philmcmahon:
I think it's going to be hard to share the same images between apps/dotcom without repeatedly keeping one or the other up to date... I think @paperboyo is right - better to have dotcom and apps use `?width=10&dpr=2` rather than apps using `?width=20` instead, but they are identical in terms of the image itself.

PR #332, opened 2020-05-20
add blockquote component with styles

webb04:
## Why are you doing this?

The current apps and dotcom have some specific blockquote styles to make them stand out. We can convert the icon to an SVG when the existing ticket gets picked up.

## Screenshots

| Before | After light mode | After dark mode |
| --- | --- | --- |
|  |  |  |

### DIFF
--- a/src/components/blockquote.tsx
+++ b/src/components/blockquote.tsx
@@ -1,7 +1,7 @@
 // ----- Imports ----- //
 
 import React, { FC, ReactNode } from 'react';
-import { css } from '@emotion/core';
+import { css, SerializedStyles } from '@emotion/core';
 import { Format } from '@guardian/types/Format';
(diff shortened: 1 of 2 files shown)


src/components/blockquote.tsx
```diff
@@ -0,0 +1,49 @@
...
+interface Props {
+    children?: ReactNode;
+    format: Format;
+}
+
+const styles = (format: Format) => css`
+    font-style: italic;
+    position: relative;
+    margin: 1rem 0 2rem 0;
+    padding: 0 1.5rem;
```
JamieB-gu, line 22:
```suggestion
    margin: ${remSpace[4]} 0 ${remSpace[8]} 0;
    padding: 0 ${remSpace[6]};
```

PR #334, opened 2020-05-20
use object fit to keep aspect ratio of images

webb04:
Recommended by lighthouse best practices audit.
Would be good to try out the latest audits in version 6 https://web.dev/lighthouse-whats-new-6.0/.

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/bodyImage.tsx
+++ b/src/components/bodyImage.tsx
@@ -26,7 +26,8 @@ const imgStyles = (width: number, height: number): SerializedStyles => css`
     background: ${neutral[97]};
     display: block;
     width: 100%;
- 
+    object-fit: cover;
+
     ${from.phablet} {
         height: calc(620px * ${height / width});
     }


src/components/bodyImage.tsx
```diff
@@ -26,7 +26,8 @@ const imgStyles = (width: number, height: number): SerializedStyles => css`
     background: ${neutral[97]};
     display: block;
     width: 100%;
- 
+    object-fit: cover;
```
JamieB-gu, line 29: In theory the CSS is sizing this box to be the exact dimensions of the image, but doesn't hurt to be safe 👍

webb04: Maybe it's related to the side paddings. Removing `width: 100%` also maintains the aspect-ratio, but with it they aren't the exact dimensions

PR #335, opened 2020-05-20
Liveblog Updates Endpoint

JamieB-gu:
## Why are you doing this?

This adds an endpoint for retrieving liveblog updates, at `/<articleId>/live-blocks`, with a JSON response of the form:

```ts
type LiveUpdates = {
    newBlocks: LiveBlock[];
    updatedBlocks: LiveBlock[];
}
```

Currently the endpoint will only work for `GET` requests in local development; I haven't added a `POST` equivalent for MAPI to hit yet. The client also hasn't been updated to use this endpoint. Both of these changes will come in later PRs.

FYI @oliverlloyd @gtrufitt 

## Changes

- Introduced `JsonSerialisable` type
- Added converter to make `BodyElement` `JsonSerialisable`
- Migrated `isElement` to `lib` for reuse
- Added converter to make `LiveBlock` `JsonSerialisable`
- Added functions to `liveBlock` to retrieve recent updates
- Made JS in `InteractiveAtom` an `Option`
- Added module for parsing query params sent to server
- Added server endpoint for new and updated liveblog blocks
- Added converter to make `Option` `JsonSerialisable`

### DIFF
--- a/src/liveBlock.ts
+++ b/src/liveBlock.ts
@@ -71,9 +71,9 @@ const filterBlocks = (filterFunc: (block: Block) => boolean) => (content: Conten
 
 const blocksSince =
     (getDate: (block: Block) => CapiDateTime | undefined) =>
-    (since: number): (content: Content) => (context: Context) => LiveBlock[] =>
+    (since: Date): (content: Content) => (context: Context) => LiveBlock[] =>
 {
-    const isRecent = compose(moreRecentThan(new Date(since)), getDate);
+    const isRecent = compose(moreRecentThan(since), getDate);
 
     return compose(parseMany, filterBlocks(isRecent));
 }
--- a/src/liveBlock.ts
+++ b/src/liveBlock.ts
@@ -1,10 +1,19 @@
 // ----- Imports ----- //
 
 import { IBlock as Block } from 'mapiThriftModels/Block';
-import { Option } from 'types/option';
+import { ICapiDateTime as CapiDateTime } from 'mapiThriftModels/CapiDateTime';
+import { IContent as Content } from 'mapiThriftModels/Content';
+import { Option, toSerialisable as optionToSerialisable } from 'types/option';
 import { Context } from 'types/parserContext';
-import { Body, parseElements } from 'bodyElement';
+import JsonSerialisable from 'types/jsonSerialisable';
+import {
+    Body,
+    parseElements,
+    toSerialisable as bodyElementToSerialisable,
+} from 'bodyElement';
 import { maybeCapiDate } from 'capi';
+import { partition } from 'types/result';
+import { compose } from 'lib';
 
 
 // ----- Types ----- //
@@ -21,6 +30,26 @@ type LiveBlock = {
 
 // ----- Functions ----- //
 
+const serialiseLiveBlock = ({
+    id,
+    isKeyEvent,
+    title,
+    firstPublished,
+    lastModified,
+    body,
+}: LiveBlock): JsonSerialisable =>
+    ({
+        id,
+        isKeyEvent,
+        title,
+        firstPublished: optionToSerialisable(firstPublished),
+        lastModified: optionToSerialisable(lastModified),
+        body: partition(body).oks.map(bodyElementToSerialisable),
+    });
+
+const toSerialisable = (blocks: LiveBlock[]): JsonSerialisable =>
+    blocks.map(serialiseLiveBlock);
+
 const parse = (context: Context) => (block: Block): LiveBlock =>
     ({
         id: block.id,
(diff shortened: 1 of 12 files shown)


src/liveBlock.ts
```diff
@@ -31,13 +60,38 @@ const parse = (docParser: DocParser) => (block: Block): LiveBlock =>
...
+{
+    const isRecent = compose(moreRecentThan(new Date(since)), getDate);
+
+    return compose(parseMany, filterBlocks(isRecent));
+}
+
+const newBlocksSince = blocksSince(block => block.firstPublishedDate);
+const updatedBlocksSince = blocksSince(block => block.lastModifiedDate);
+
+const recentBlocks = (num: number) => (content: Content): (docParser: DocParser) => LiveBlock[] =>
```
webb04, line 84: Will we also need a `blocksAround` when accessing key events?

src/liveBlock.ts
```diff
@@ -21,6 +30,26 @@ type LiveBlock = {
...
+    title,
+    firstPublished,
+    lastModified,
+    body,
+}: LiveBlock): JsonSerialisable =>
+    ({
+        id,
+        isKeyEvent,
+        title,
+        firstPublished: optionToSerialisable(firstPublished),
```
alexduf, line 45: Isn't the fact we have to pass options to a special function a smell that we'd be better off without the `Option` type?
Could `MyType | undefined | null` suffice?

src/server/server.ts
```diff
@@ -128,6 +147,47 @@ async function serveArticle(req: Request, res: ExpressResponse): Promise<void> {
...
+
+const recentLiveBlocks = (content: Content): LiveUpdates => ({
+    newBlocks: serialiseLiveBlocks(recentBlocks(10)(content)(docParser)),
+    updatedBlocks: [],
+});
+
+async function liveBlocks(req: Request, res: ExpressResponse): Promise<void> {
+    try {
+        const articleId = req.params.articleId || defaultId;
+        const since = parseNumber(req.query.since);
```
alexduf, line 163: I think this should be a date. It's fine while developing, but in prod we should be consistent with MAPI

src/server/server.ts
```diff
@@ -128,6 +147,47 @@ async function serveArticle(req: Request, res: ExpressResponse): Promise<void> {
     }
 }
 
+const liveBlockUpdates = (since: number, content: Content): LiveUpdates => ({
+    newBlocks: serialiseLiveBlocks(newBlocksSince(since)(content)(docParser)),
+    updatedBlocks: serialiseLiveBlocks(updatedBlocksSince(since)(content)(docParser)),
+});
+
+const recentLiveBlocks = (content: Content): LiveUpdates => ({
+    newBlocks: serialiseLiveBlocks(recentBlocks(10)(content)(docParser)),
```
webb04, line 156: Could this be something we could configure?
`const blocksPerRequest = 10`

src/liveBlock.ts
```diff
@@ -21,6 +30,26 @@ type LiveBlock = {
 
 // ----- Functions ----- //
 
+const serialiseLiveBlock = ({
+    id,
+    isKeyEvent,
```
webb04, line 35: I think there's another type of block `isSummaryEvent` which we currently style to look like keyEvents in the templates. Maybe we should add this to the model?

src/liveBlock.ts
JamieB-gu: Yup, this is just the standard polling endpoint to start. There will be some more to implement to handle the other user interactions we support nowadays.

src/server/server.ts
JamieB-gu: Of the format: `2020-05-21T10:00:00Z`?

src/server/server.ts
JamieB-gu: Yup, see no reason why not, that's why I included it as a param rather than hardcoding it. This is really just a fallback for the moment to make the API well-formed (i.e. what happens when you hit the endpoint with no query param), I don't know if anything on the client will actually hit this.

src/liveBlock.ts
JamieB-gu: Wasn't aware of this, do you have an example of what that looks like? 🤔

src/liveBlock.ts
JamieB-gu: I feel the `Option` type offers us a lot of benefits that I wouldn't want to throw away for the sake of avoiding a bit of work to serialise them.

What this says to me is that some of structures we use to represent our data don't fit into the subset of types that count as valid JSON (`Option` isn't the only example of this). I don't think that should force us to lose the richness of our data modelling by mandating 1:1 JSON compatibility.

src/liveBlock.ts
webb04: https://www.theguardian.com/world/live/2020/may/20/coronavirus-live-news-brazil-sees-record-daily-deaths-as-world-bank-warns-60m-to-fall-into-extreme-poverty?page=with:block-5ec5aece8f08a1782fa7e955#block-5ec5aece8f08a1782fa7e955

It looks just like a key event. You can see them by going to capi and searching `is-summary`. It looks like those classes are only added in the the big html fields and not in the blocks api. Maybe in the blocks api they are always set to be `key-events`

JamieB-gu: It *looks* like the blocks API provides `attributes.summary: true` for those, so we can probably use that? Seems like the changes we have to make for them is to added the headline `"Summary"` and include them in the key events (which also means adding key events styles)?

webb04: I think you can have a block that is either “Summary”, “Key Event” or both. Both look the same (yellow title and top border colour) but only key events go into the bullet point links at the top.

JamieB-gu: Yeah spoke too soon, *some* of them have `attributes.summary: true`, others just have `attributes.keyEvent: true`. So what you're saying is that they're *not* a key event, so don't appear in the key events list, but we *do* apply the same styles as key events?

webb04: Yes that's my understanding. We added that logic to the templates a while ago @benwuersching

JamieB-gu: Ok, good to know. Unrelated to this PR so I've created a ticket.

src/liveBlock.ts
alexduf: I wasn't aware of the `replacer` feature, that might be a good candidate?

>I feel the Option type offers us a lot of benefits that I wouldn't want to throw away for the sake of avoiding a bit of work to serialise them.

That's fair!

src/server/server.ts
alexduf: yep!

src/liveBlock.ts
JamieB-gu: As an aside, this problem might go away if we were to move away from classes and methods to data types and pure functions. See [this related discussion](https://github.com/guardian/types/issues/5#issuecomment-632045893) that's also going on today.

src/server/server.ts
JamieB-gu: Done!

PR #336, opened 2020-05-21
style tweaks for immersive articles

webb04:
## Changes

- Opinion colour for Opinion headline
- Only use heading caption styles for media articles
- Immersive headline left padding on wide breakpoint
- Reduce height of header image so that the bottom navbar doesn't block the top of the headline
- Reduce immersive standfirst font size slightly and use `renderStandfirstText`
- Add min height to immersive headlines to handle short headlines

## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |

### DIFF
--- a/src/components/headerImage.tsx
+++ b/src/components/headerImage.tsx
@@ -74,7 +74,7 @@ const imgStyles = (width: number, height: number): SerializedStyles => css`
 
 const immersiveImgStyles = css`
     display: block;
-    height: 100vh;
+    height: 80vh;
     object-fit: cover;
     width: 100vw;
 `;
--- a/src/components/standfirst.tsx
+++ b/src/components/standfirst.tsx
@@ -54,7 +54,7 @@ const thinHeadline = css`
 
 const immersive: SerializedStyles = css`
     ${styles}
-    ${headline.xsmall({ fontWeight: 'light' })}
+    ${headline.xxsmall({ fontWeight: 'light' })}
     margin-top: ${remSpace[3]};
 `;
 
@@ -110,7 +110,7 @@ function content(standfirst: DocumentFragment, item: Item): ReactNode {
             <>
                 {rendered}
                 <address>
-                    <p>By {renderText(byline, format)}</p>
+                    <p>By {renderStandfirstText(byline, format)}</p>
                 </address>
             </>
         ).withDefault(rendered);
--- a/src/components/standfirst.tsx
+++ b/src/components/standfirst.tsx
@@ -54,7 +54,7 @@ const thinHeadline = css`
 
 const immersive: SerializedStyles = css`
     ${styles}
-    ${headline.xxsmall({ fontWeight: 'light' })}
+    ${headline.xsmall({ fontWeight: 'light' })}
     margin-top: ${remSpace[3]};
 `;
 
--- a/src/components/standfirst.tsx
+++ b/src/components/standfirst.tsx
@@ -7,7 +7,7 @@ import { background, neutral, text } from '@guardian/src-foundations/palette';
 import { remSpace } from '@guardian/src-foundations';
 
 import { Item, getFormat } from 'item';
-import { renderText, renderStandfirstText } from 'renderer';
+import { renderStandfirstText } from 'renderer';
 import { darkModeCss as darkMode } from 'styles';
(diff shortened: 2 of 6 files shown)


src/components/headerImage.tsx
```diff
@@ -74,7 +74,7 @@ const imgStyles = (width: number, height: number): SerializedStyles => css`
 
 const immersiveImgStyles = css`
     display: block;
-    height: 100vh;
+    height: 80vh;
```
JamieB-gu, line 77: Would you be able to bump this up very slightly? @blongden73 told me the idea is the bottom of the page should be flush with the baseline of the second line of text, i.e. the top of the nav should hit (but not cover) approximately the bottom of the "m" in the word "masked" in your screenshot.

src/components/standfirst.tsx
```diff
@@ -54,7 +54,7 @@ const thinHeadline = css`
 
 const immersive: SerializedStyles = css`
     ${styles}
-    ${headline.xsmall({ fontWeight: 'light' })}
+    ${headline.xxsmall({ fontWeight: 'light' })}
```
JamieB-gu, line 57: Dotcom has `xsmall` for the standfirst.

src/editorialPalette.ts
```diff
@@ -65,6 +65,8 @@ const textHeadlineInverse = (_: Format): Colour =>
 const backgroundHeadlinePrimary = (format: Format): Colour => {
     if (format.display === Display.Immersive) {
         return neutral[7];
+    } else if (format.pillar === Pillar.Opinion) {
+        return opinion[800];
```
JamieB-gu, line 69: 👍

src/components/headerImage.tsx
webb04: Is this another app vs dotcom style question. The rule for the current templates was to show the first two lines of the headline...

Sorry for another tag @benwuersching 😀

src/components/standfirst.tsx
webb04: hmm it does, where did that come from because `frontend` uses 1.125rem whereas that would be 1.5rem 😅

src/components/headerImage.tsx
benwuersching: Showing two lines of headline sounds good, however, I'd have to check with Ben L what his reasonings were. // And here I go off on a tangent: Looking at dotcom, I'm still trying to get my head around what exactly is supposed to happen at the top of a picture essay. How deep is the picture supposed to be on compact devices and tablet portrait? On dotcom, it's currently not as deep as on immersives (not full screen), more like square. Is that right?

webb04: I see what you mean @benwuersching on dotcom. It looks like dotcom-rendering is more fullscreen and shows the first two lines of the headline https://www.theguardian.com/artanddesign/2020/may/21/mind-the-gap-two-metres-apart-masked-ridley-road-hackney-a-photo-essay?dcr

Seen as that's a newer design I assume that's the way they are going?

src/components/standfirst.tsx
JamieB-gu: What was the conclusion on this?

src/components/headerImage.tsx
benwuersching: @webb04 Hello again (was away yesterday). Yes, picture essay should look as feature-y as possible. But I still have to better understand the thinking behind dotcom-rendering. For picture essays, dotcom goes "more fullscreen" as you write, but not quite as fullscreen as immersive. Here is a desktop comparison (it's like they show one headline line less).
picture_essay_desktop
immersive_desktop

src/components/standfirst.tsx
webb04: It seemed a little big to me at the time so thought about reducing it to the size on other articles. Happy to revert though.
IMG_5301

JamieB-gu: I'd strongly advocate for matching dotcom if we can, it makes it easier to share.

webb04: Ok I've reverted that one. I guess as we gradually role article types out to users we can have some more general design reviews too

PR #339, opened 2020-05-26
wrap related content in anchor

webb04:
- Makes whole grey box linkable
- Removes kicker colour on link (closer to current apps) as the related content may be a different pillar to the current article @benwuersching

Screen Shot 2020-05-26 at 11 53 54

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -298,23 +298,17 @@ const richLinkWidth = '8.75rem';
 const richLinkStyles = css`
     background: ${neutral[97]};
     padding: ${basePx(1)};
+    text-decoration: none;
+    border-bottom: none;
+    color: ${neutral[7]};
 
     h1 {
         margin: ${basePx(0, 0, 2, 0)};
-        font-size: 1rem;
-    }
-
-    p {
-        margin: ${basePx(1, 0)};
+        font-size: ${remSpace[4]};
     }
 
     span {
         display: none;
-    }
-
-    a {
-        text-decoration: none;
-        border-bottom: none;
     }
 
     float: left;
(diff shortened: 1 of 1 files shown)


src/renderer.ts
```diff
@@ -298,25 +298,19 @@ const richLinkWidth = '8.75rem';
...
+    color: ${neutral[7]};
 
     h1 {
         margin: ${basePx(0, 0, 2, 0)};
-        font-size: 1rem;
-    }
-
-    p {
-        margin: ${basePx(1, 0)};
+        font-size: ${remSpace[4]};
```
JamieB-gu, line 307: Are we able to use the design system typography for this?

src/renderer.ts
```diff
@@ -329,17 +323,17 @@ const richLinkStyles = css`
     ${darkModeCss`
         background-color: ${neutral[20]};
         color: ${neutral[60]};
-        a {
+
+        h1 {
             color: ${neutral[60]};
-            border-bottom: 0.0625rem solid ${neutral[60]};
+            border-bottom: none;
```
JamieB-gu, line 329: If you leave this unset does a border appear?

src/renderer.ts
```diff
@@ -329,17 +323,17 @@ const richLinkStyles = css`
...
     `}
 `;
 
-const RichLink = (props: { url: string; linkText: string; format: Format }): ReactElement =>
-    styledH('aside', { css: richLinkStyles },
-        h('h1', null, props.linkText),
-        h(Anchor, { href: props.url, format: props.format }, 'Read more'),
+const RichLink = (props: { url: string; linkText: string }): ReactElement =>
+    styledH('a', { href: props.url, css: richLinkStyles },
+        styledH('aside', null, [h('h1', null, props.linkText), 'Read more'])
```
JamieB-gu, line 336: Would it be possible to swap the `a` and `aside` here? So that the top level element is an `aside` and the `a` lives within that?

src/renderer.ts
webb04: Nope that can be removed

benwuersching:
@webb04 Hello Jamie, you tagged me on this one. Looking at the screenshot, I'm a bit confused. Is this a new proposed styling for this component? If so, then we should refine it a bit. I can help with that.

webb04:
@benwuersching that's the current apps-rendering implementation but we can change that if we have a better design.

At the moment we only have the pillar colour available for the article (the linked story might be a different colour but we don't know that yet)

benwuersching:
@webb04 I quickly put this together: https://www.figma.com/file/oRmb9eJKgm2cuhUfgsUI0X/in-article-rich-link?node-id=1%3A100

webb04:
Thanks @benwuersching that looks a lot better

PR #343, opened 2020-05-26
set width on comment bylines

webb04:
IMG_5227


src/components/byline.tsx
```diff
@@ -40,6 +40,7 @@ const anchorStyles = (kicker: string, inverted: string): SerializedStyles => css
 
 const commentStyles = (kicker: string): SerializedStyles => css`
     color: ${kicker};
+    width: 75%;
```
JamieB-gu, line 43: Is this needed for the wider breakpoints?

webb04: I think so yes, it could be more than 75% but I think that's a real edge case to have a byline that long

PR #344, opened 2020-05-27
add white background and padding to interactive atoms and embeds in dark mode

webb04:
This is the current behaviour in dark mode apps. There must be some JavaScript setting inline padding styles for some interactive atoms, so I've needed to use `!important` there.

## Changes

- Adds a white background to interactive atoms on standard articles (not type: Interactive) in dark mode
- Adds white background to embeds in dark mode
- adds .5rem padding to these elements in dark mode

## Screenshots

| Embed | Interactive atom on standard article |
| --- | --- |
|  |  |

Dark mode is now supported in both Android and iOS live apps. This fallback ensures embeds and interactive atoms are readable, but it would be nicer if they had dark mode styles to blend into the page better.

The article pages in apps-rendering are currently not used in production but they should be fairly soon. I would be interested in hearing whether people are ok with these fallbacks or if we can better support this content in apps-rendering (the new mobile apps article templates) 

@pixelsquid @tinius @seanclarkeguardian @superfrank @mchv @JamieB-gu

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -432,7 +432,13 @@ const render = (format: Format, excludeStyles = false) =>
                 },
             };
 
-            const figureCss = css`margin ${remSpace[4]} 0`;
+            const figureCss = css`
+                margin ${remSpace[4]} 0;
+                ${darkModeCss`
+                    background: white;
+                    padding: ${remSpace[2]};
+                `}
+            `;
             const captionStyles = css`
                 ${textSans.xsmall()}
                 color: ${textColour.supporting};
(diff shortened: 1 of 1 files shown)


webb04:
> Looks sensible.
> 
> > There must be some JavaScript setting inline background styles/padding for some interactive atoms
> 
> Some of them are running JS that sets a background colour on the `body`? If they're doing that, do we need to overwrite those styles? In that scenario they're setting their own background colour, so they'll have whatever background they've set and not the dark mode one?

Maybe it was just padding actually. But if interactive atoms could set their own background that would be great.

webb04:
> If it's just `padding`, do we need `!important` on the `background` declaration?

Gave it another try and it is required.

PR #345, opened 2020-05-27
Live Blocks On Client

JamieB-gu:
## Why are you doing this?

This begins the work to request new live blocks on the client. These changes are quite contained: on page load the JS will make a request to get the most recent 7 blocks (which matches what the server has rendered for the page). It will then put the liveblog body under the control of client-side React, using the blocks retrieved as the data.

This is a follow-on from #335, and fixes the client-side rendering that was temporarily disabled in that PR. It also temporarily removes client-side hydration, hardcoding the `Format` and total blocks for now. A fix for *this* will come in a further PR; it requires some restructuring that would balloon the size of these changes.

**Note:** I'm not particularly happy with the conversions from JSON applied here; they create the potential for runtime errors due to the `any` return type of `JSON.parse`. In theory it should be fine because we control both sides of the API, but I still think we could do better. This isn't the highest priority right now, but I may look into this again down the line. One alternative would be to use Thrift, but that may cause a big spike in the bundle size.

## Changes

- Added a deserialiser for `BodyElement`
- Introduced new `fromString` function for `date` module
- Added a deserialiser for `LiveBlock`
- Rewrote liveblog client script to load initial blocks
- Temporarily hardcoded `Format` and total blocks for liveblog client
- Fixed bug where blocks article element was duplicated
- Temporarily removed hydration props

### DIFF
--- a/src/client/liveblog.ts
+++ b/src/client/liveblog.ts
@@ -1,45 +1,50 @@
 // ----- Imports ----- //
 
-import setup from 'client/setup';
-import { fromCapiLiveBlog, getFormat } from 'item';
 import ReactDOM from 'react-dom';
-import LiveblogBody from 'components/liveblog/body';
 import { createElement as h } from 'react';
-import { Content } from 'mapiThriftModels';
-import { parse } from './parser';
+import { Format, Design, Display, Pillar } from '@guardian/types/Format';
+
+import setup from 'client/setup';
+import { fromSerialisable } from 'liveBlock';
+import { parse } from 'client/parser';
+import LiveblogBody from 'components/liveblog/body';
+
+
+// ----- Setup ----- //
+
+const domParser = new DOMParser();
+
+const format: Format = {
+    design: Design.Live,
+    display: Display.Standard,
+    pillar: Pillar.News,
+}
+
+
+// ----- Functions ----- //
+
+const docParser = (html: string): DocumentFragment =>
+    parse(domParser)(html)
+        .toOption()
+        .withDefault(new DocumentFragment());
+
+const deserialise = fromSerialisable(docParser);
 
 
 // ----- Run ----- //
 
-declare global {
-    interface Window {
-        __INITIAL__DATA__: Content;
-    }
-}
-
+// Set up interactives, Twitter embeds etc.
 setup();
 
-const browserParser = (html: string): DocumentFragment =>
-    parse(new DOMParser())(html)
-        .toOption()
-        .withDefault(new DocumentFragment())
+// Load the initial group of blocks.
+fetch(`${window.location}/live-blocks`)
+    .then(res => res.json())
+    .then(({ newBlocks }) => {
+        const blocks = deserialise(newBlocks);
 
-
-try {
-    const hydrationProps = JSON.parse(document.getElementById('hydrationProps')?.textContent ?? "");
-
-    if (hydrationProps) {
-        const item = fromCapiLiveBlog({
-            docParser: browserParser,
-            salt: 'mockSalt',
-        })(hydrationProps);
-        const { blocks, totalBodyBlocks } = item;
-
-        ReactDOM.hydrate(
-            h(LiveblogBody, { format: getFormat(item), blocks, totalBodyBlocks }),
-            document.querySelector('#blocks'),
+        ReactDOM.render(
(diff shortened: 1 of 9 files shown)


src/client/liveblog.ts
```diff
@@ -1,45 +1,50 @@
...
+setup();
 
-    if (hydrationProps) {
-        const item = fromCapiLiveBlog({
-            docParser: browserParser,
-            salt: 'mockSalt',
-        })(hydrationProps);
-        const { blocks, totalBodyBlocks } = item;
+// Load the initial group of blocks.
+fetch(`${window.location}/live-blocks`)
```
webb04, line 40: Can you use the relative path? 

`fetch(`/live-blocks`)`

src/client/liveblog.ts
```diff
@@ -1,45 +1,50 @@
...
+    .then(res => res.json())
+    .then(({ newBlocks }) => {
+        const blocks = deserialise(newBlocks);
 
-        ReactDOM.hydrate(
-            h(LiveblogBody, { format: getFormat(item), blocks, totalBodyBlocks }),
-            document.querySelector('#blocks'),
+        ReactDOM.render(
```
webb04, line 45: How comes we're no longer hydrating?

src/server/server.ts
```diff
@@ -154,7 +152,7 @@ const liveBlockUpdates = (since: Date, content: Content, context: Context): Live
 });
 
 const recentLiveBlocks = (content: Content, context: Context): LiveUpdates => ({
-    newBlocks: serialiseLiveBlocks(recentBlocks(10)(content)(context)),
+    newBlocks: serialiseLiveBlocks(recentBlocks(7)(content)(context)),
```
webb04, line 155: I think the current app requests 10 when fetching blocks.

src/client/liveblog.ts
JamieB-gu: That will request `localhost:8080/live-blocks` I'm afraid.

src/server/server.ts
JamieB-gu: Which request are you talking about here, we have a few 🙂. The "View More" request? The polling request? The pull-to-refresh?

src/client/liveblog.ts
JamieB-gu: At the moment because technically the data returned by `/live-blocks` might not be the same as what the server has rendered. There are a few sources of delay between the time the HTML is rendered on the server and the time the `fetch` request goes out:

- App caching
- Fastly caching
- Document request/response/render cycle

Liveblogs move quickly, and blocks may have been changed or new blocks added in that time. It might be that this is fine, or it might be that we delay this request until the first polling interval, or it might be best to store the initial few blocks using the `application/json` mechanism used currently. I haven't quite figured it out yet, but this implementation is pretty minimal at the moment and will probably change. There will _have_ to be a mechanism for accessing the `Format` and total block count regardless, so it might be that the mechanism used for that is used for this too.

src/server/server.ts
webb04: I think all of the requests that load more blocks fetch 10. I'm not sure why the initial 7 is different :)

JamieB-gu: Also, it's different on dotcom, I think it varies depending on the piece 🤷‍♂️.

Anyway, as mentioned above this is currently replacing client-side the initial seven rendered on the server; if I make it 10 three new blocks will pop in 🙂. The way this works may change in future.

src/client/liveblog.ts
webb04: Oh so you need the id passed in the request?

src/client/liveblog.ts
webb04: We will still SSR the initial blocks though? so a cached home front can still display those offline.

src/client/liveblog.ts
JamieB-gu: Well we need some way of identifying the article so we can send a request to CAPI for the blocks. We *could* do that by pulling it out of the "referer" header I suppose...

src/client/liveblog.ts
JamieB-gu: The initial blocks are still rendered on the server, yes.

PR #346, opened 2020-05-27
Change fontSize.css URL

maxspencer:
## Why are you doing this?

Because the current URL (`native://fontSize.css`) caused mixed content errors/warnings in the web console and the request may be blocked depending on the security policy.

Changing it to `https://` should allow it to go ahead and be intercepted in the native layer.

## Changes

- The URL used to load the base font size stylesheet.

### DIFF
--- a/src/components/shared/page.tsx
+++ b/src/components/shared/page.tsx
@@ -189,7 +189,7 @@ function Page({ content, imageSalt, getAssetLocation }: Props): ElementWithResou
                 <title>{content.id}</title>
                 <meta id="twitter-theme" name="twitter:widgets:theme" content="light" />
                 <meta name="viewport" content="initial-scale=1, maximum-scale=5" />
-                <link rel="stylesheet" href="native://fontSize.css" />
+                <link rel="stylesheet" href="https://localhost/fontSize.css" />
             </head>
             <body>
                 { element }


src/components/shared/page.tsx
```diff
@@ -189,7 +189,7 @@ function Page({ content, imageSalt, getAssetLocation }: Props): ElementWithResou
                 <title>{content.id}</title>
                 <meta id="twitter-theme" name="twitter:widgets:theme" content="light" />
                 <meta name="viewport" content="initial-scale=1, maximum-scale=5" />
-                <link rel="stylesheet" href="native://fontSize.css" />
+                <link rel="stylesheet" href="https://localhost/fontSize.css" />
```
JamieB-gu, line 192: This will (I think) attempt to make a request to `localhost` on port `80` in all circumstances. That won't work locally or when hitting MAPI in a web browser, but we don't expect this request to work anyway in those cases so that's fine.

However, what will this make a request to in the webviews? Is there a server running on `localhost:80` in the apps? If not, is it possible we want `href="/fontSize.css"`, which will make a relative request and pick up the protocol/domain/port from the parent document (in DEV `localhost:8080/fontSize.css`, in PROD `https://mobile.guardianapis.com/fontSize.css`)?

maxspencer: On Android there's no server running on `localhost:80` but I think that's OK because we are going to intercept this request in the native layer anyway. That means the webview won't try to make a request to `localhost:80` in the normal way.

We'll need to check what will happen on iOS (@faresite) but my assumption is that there would be no server at `localhost:80` there either and the request would fail (and that won't matter because the base font size is provided differently on iOS).

I thought that for iOS this `localhost` URL might be preferable to using `https://fontSize.css` because, as Fares pointed out, that could trigger a DNS lookup on `fontSize.css` which is unnecessary. AFAIK `.css` is not a real TLD so that DNS lookup will always fail, but they're adding new TLDs left right and centre these days so perhaps another another good reason no to use `https://fontSize.css` is that it might one day actually point to something we don't control!

JamieB-gu: Yeah that's a good point, in URL terms that's TLD `css`, domain `fontSize`.

Is there any reason why we wouldn't want it to be `https://mobile.guardianapis.com/fontSize.css`?

maxspencer: `https://mobile.guardianapis.com/fontSize.css` would work fine for Android - we would be able to intercept it just the same way.

On iOS we want this request to fail as quickly/quietly as possible and this is where I think it might not be quite so good? I might be over-thinking it but it seems likely that the `localhost` request will fail pretty quick whereas using `mobile.guardianapis.com` will cause an actual request to be made and to fail with a 404?

JamieB-gu: Yeah, fair point, and stylesheets are render blocking so this isn't ideal.

Ok, I can't see any blockers for doing this. It feels strange to me to request a resource that we _know_ doesn't exist, but I guess we're limited by the platforms here.

maxspencer:
I've made the change to the Android repo such that any request where the path is `fontSize.css` will be intercepted: https://github.com/guardian/android-news-app/pull/6062

I'm happy for anyone to modify this PR further to use a different URL 👍

src/components/shared/page.tsx
faresite: How do you feel about adding a custom attribute like`data-platform="android"` to tags that are there for one platform only? I could basically remove these tags from the html document native side?

JamieB-gu: That's possible. Only downside I can see is that it's one more piece of HTML manipulation on the iOS side 🤔

maxspencer:
Just clearing out my GitHub notifications and noticed this is still open. What do we need to do with ti?

JamieB-gu:
> What do we need to do with it?

I have a feeling we may have come up with another solution? @faresite how does iOS currently handle the resizing? And @maxspencer does resizing on Android work without this?

maxspencer:
@JamieB-gu IIRC this was an Android-only thing, @faresite can confirm but I thin font sizes are set in a different way on iOS.


The font resize buttton doesn't work on Android yet (I'm going to resolve that soon) but when it does the desired font size will be signalled to the templates via this mechanism.

JamieB-gu:
> font resize buttton

After moving the slider, will a user have to reload the article to see the font size update?

maxspencer:
> > font resize buttton
> 
> After moving the slider, will a user have to reload the article to see the font size update?

Yeah, I'll "invalidate" the view after the new size preference is saved

JamieB-gu:
> I'll "invalidate" the view after the new size preference is saved

Oh so from the user's perspective they'll see the page resize in front of them?

If we hit `localhost` is it possible that any other app or system process that's running a server on port 80 might respond? I'm not familiar to what extent Android and iOS lock this down but on macOS, for example, that would probably happen. Presumably on Android we're intercepting that request so it won't be an issue, but what about iOS?

maxspencer:
> > I'll "invalidate" the view after the new size preference is saved
> 
> Oh so from the user's perspective they'll see the page resize in front of them?

Correct
 
> If we hit `localhost` is it possible that any other app or system process that's running a server on port 80 might respond? I'm not familiar to what extent Android and iOS lock this down but on macOS, for example, that would probably happen. Presumably on Android we're intercepting that request so it won't be an issue, but what about iOS?

On Android we'll intercept it as you say. I can't say definitively what will happen on iOS, but I suspect nothing, perhaps we can merge this to unblock the Android side and then test for any effect on iOS?

maxspencer:
@JamieB-gu would it be better to make the request to the MAPI domain?

JamieB-gu:
> would it be better to make the request to the MAPI domain?

That should work. It would get captured on Android and most likely result in a 404 on iOS.

maxspencer:
OK cool, happy to make that change, can you help with my permissions though - I can no longer merge against this repo:
image

PR #348, opened 2020-05-27
format byline links

webb04:
`bylineHtml` only provides us with a relative profile url. Changing to the absolute path ensures iOS and Android can correctly link to profile pages.

### DIFF
--- a/src/components/byline.tsx
+++ b/src/components/byline.tsx
@@ -103,11 +103,18 @@ const getAnchorStyles = (format: Format): SerializedStyles => {
     }
 }
 
+const getProfileLink = (node: Node): string => {
+    const href = getHref(node).withDefault('');
+    return href.startsWith('profile/')
+        ? `https://www.theguardian.com/${href}`
+        : href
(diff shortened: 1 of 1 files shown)


src/components/byline.tsx
```diff
@@ -103,11 +103,18 @@ const getAnchorStyles = (format: Format): SerializedStyles => {
     }
 }
 
+const getProfileLink = (node: Node): string => {
+    const href = getHref(node).withDefault('');
+    return href.startsWith('profile/')
+        ? `https://www.theguardian.com/${href}`
```
JamieB-gu, line 109: Without this does it point to: `https://mobile.guardianapis.com/profile/journalist`? The apps presumably don't intercept MAPI URLs?

webb04: Yeah the apps seem to handle profile links on dotcom but not from mapi. Think it makes sense to make the update here rather than in two places!

Really I think capi should be returning the absolute link in this case.

PR #350, opened 2020-05-28
better related content designs

webb04:
## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |


src/renderer.ts
```diff
@@ -313,6 +313,23 @@ const richLinkWidth = '8.75rem';
...
+        margin-bottom: 0;
+        padding-bottom: 0;
+    }
+
+    p::before {
+        ${icons}
+        content: '\\e005';
+        border-radius: 100%;
+        border: solid 1px ${neutral[7]};
+        padding: 4px 5px 5px 5px;
```
webb04, line 329: The icon font has additional space above the character. Hopefully we can solve this if we move to svgs.

src/renderer.ts
```diff
@@ -313,6 +313,23 @@ const richLinkWidth = '8.75rem';
...
+        font-size: 12px;
```
JamieB-gu, line 330: We might want these in `rem` to avoid scaling issues?

src/renderer.ts
```diff
@@ -346,7 +367,7 @@ const richLinkStyles = css`
 
 const RichLink = (props: { url: string; linkText: string }): ReactElement =>
     styledH('aside', { css: richLinkStyles },
-        styledH('a', { href: props.url }, [h('h1', null, props.linkText), 'Read more'])
+        styledH('a', { href: props.url }, [h('h1', null, props.linkText), h('p', null, 'Read more')])
```
JamieB-gu, line 370: I know this is wrapped in a link, but should we make this a `<button>` semantically?

src/renderer.ts
webb04: I'm a bit worried about the circle looking less round at some sizes, I think we can definitely use rem sizing for the svg though https://github.com/guardian/source/issues/371

PR #356, opened 2020-05-29
Add Content Security Policy

JamieB-gu:
## Why are you doing this?

CSP's are important for security; they limit what resources (scripts, styles, iframes etc.) are accepted by your page. For a more detailed explanation see [this MDN article](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP).

This implements a CSP using the `<meta>` tag, as I'm not sure the HTTP header will play well with the way iOS inserts the HTML string into the webview (happy to be told otherwise though @faresite).

There are a few techniques used here:
- I've set [`default-src` to `'self'`](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP#Example_1), so any resources not explicitly described in the policy will only be accepted from our domain (MAPI)
- The interactives domain is allowed to load styles, scripts and fonts, and make AJAX requests (via [`connect-src`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/connect-src))
- Images can be loaded from MAPI or from `i.guim.co.uk` (the Fastly image service)
- If tweets are present on the page then some combination of styles, images, scripts and iframes can be loaded from some combination of the following Twitter domains: `platform.twitter.com`, `syndication.twitter.com`, `pbs.twimg.com` and `cdn.syndication.twimg.com`. Images are also allowed to be loaded from `data:` URLs in this case, because Twitter appear to load some of their SVGs in this way.

If anything I've described above sounds incorrect, or doesn't line up with the way the policy is written please let me know! Likewise with the choice of hashing algorithm/library; I've used `crypto` from Node's standard library, and `sha256` as demonstrated in some of [MDN's examples](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src#Unsafe_inline_styles). I went with hashing rather than use of nonces because we'd like to cache this content, and caching would prevent the nonce being request-unique. @adamnfish @katebee @davidfurey 

I've tried to lock this down as much as possible without breaking functionality. That said, this policy is much stricter than it was before so there may one or two things that it breaks, e.g. a stray interactive domain that I'm not aware of. If you spot anything like this, please let me know.

The implementation for extracting Emotion's styles draws from the [Emotion SSR docs](https://emotion.sh/docs/ssr#advanced-approach). DCR do [something similar](https://github.com/guardian/dotcom-rendering/blob/449144470ee68e16402b38bbd6f75f871c5c42db/src/amp/server/document.tsx#L35) for AMP articles.

**Note:** This currently blocks a use of `eval` that exists in the Webpack boilerplate code, which may cause problems for our client-side bundle. I'm looking into this.

FYI @gtrufitt @oliverlloyd @liywjl 

## Changes

- Added `create-emotion-server` as a dependency
- Added a `csp` module to build the CSP from asset hashes
- Added Twitter to CSP only if tweets are present
- Added interactives to CSP
- Extracted Emotion into single style tag for hashing
- Built html doc as string to work with Emotion's `extractCritical`

### DIFF
--- a/src/server/csp.ts
+++ b/src/server/csp.ts
@@ -29,7 +29,7 @@ const extractInteractiveAssets = (elements: BodyElement[]): Assets =>
                 styles: [ ...styles, elem.css ],
                 scripts: elem.js.fmap(js => [ ...scripts, js ]).withDefault(scripts),
             };
-        };
+        }
 
         return { scripts, styles };
     }, { scripts: [], styles: [] });
--- a/src/server/csp.ts
+++ b/src/server/csp.ts
@@ -1,0 +1,73 @@
+// ----- Imports ----- //
+
+import { createHash } from 'crypto';
+import { Design } from '@guardian/types/Format';
+
+import { BodyElement, ElementKind } from 'bodyElement';
+import { partition, Result } from 'types/result';
+import { Item } from 'item';
+import { compose } from 'lib';
+
+
+// ----- Types ----- //
+
+interface Assets {
+    scripts: string[];
+    styles: string[];
+}
+
+
+// ----- Functions ----- //
+
+const assetHash = (asset: string): string =>
+    createHash('sha256').update(asset).digest('base64');
+
+const extractInteractiveAssets = (elements: BodyElement[]): Assets =>
+    elements.reduce<Assets>(({ scripts, styles }, elem) => {
+        if (elem.kind === ElementKind.InteractiveAtom) {
+            return {
+                styles: [ ...styles, elem.css ],
+                scripts: elem.js.fmap(js => [ ...scripts, js ]).withDefault(scripts),
+            };
+        };
+
+        return { scripts, styles };
+    }, { scripts: [], styles: [] });
+
+const getElements = (item: Item): Result<string, BodyElement>[] =>
+    item.design === Design.Live ? item.blocks.flatMap(block => block.body) : item.body;
+
+const getValidElements = (item: Item): BodyElement[] =>
+    partition(getElements(item)).oks;
+
+const interactiveAssets = compose(extractInteractiveAssets, getValidElements);
+
+const assetHashes = (assets: string[]): string =>
+    assets.map(asset => `'sha256-${assetHash(asset)}'`).join(' ');
+
+const buildCsp = ({ styles, scripts }: Assets, twitter: boolean): string => `
+    default-src 'self';
+    style-src ${assetHashes(styles)} https://interactive.guim.co.uk ${twitter ? 'https://platform.twitter.com' : ''};
+    img-src 'self' https://i.guim.co.uk ${twitter ? 'https://platform.twitter.com https://syndication.twitter.com https://pbs.twimg.com data:' : ''};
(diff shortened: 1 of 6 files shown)


src/server/page.tsx
```diff
@@ -145,22 +164,33 @@ function page(
 ): Page {
     const item = fromCapi({ docParser, salt: imageSalt })(content);
     const shouldHideAds = content.fields?.shouldHideAdverts ?? false;
+    const hasTwitter = includesTweets(content);
     const clientScript = scriptName(item).fmap(getAssetLocation);
-    const html = (
-        <html lang="en" css={PageStyles}>
+    const emotionCache = createCache();
+    const { extractCritical } = createEmotionServer(emotionCache);
+    const { html: body, css, ids } = compose(extractCritical, renderToString)(
```
webb04, line 171: What benefit does this give us?

src/server/csp.ts
```diff
@@ -0,0 +1,73 @@
...
+const interactiveAssets = compose(extractInteractiveAssets, getValidElements);
+
+const assetHashes = (assets: string[]): string =>
+    assets.map(asset => `'sha256-${assetHash(asset)}'`).join(' ');
+
+const buildCsp = ({ styles, scripts }: Assets, twitter: boolean): string => `
+    default-src 'self';
+    style-src ${assetHashes(styles)} https://interactive.guim.co.uk ${twitter ? 'https://platform.twitter.com' : ''};
+    img-src 'self' https://i.guim.co.uk ${twitter ? 'https://platform.twitter.com https://syndication.twitter.com https://pbs.twimg.com data:' : ''};
+    script-src 'self' ${assetHashes(scripts)} https://interactive.guim.co.uk ${twitter ? 'https://platform.twitter.com https://cdn.syndication.twimg.com' : ''};
```
webb04, line 52: dotcom are a bit more relaxed here, there's probably a reason for it 
`script-src https: 'unsafe-inline' 'unsafe-eval'`

webb04: That could be commercial or other third parties though.

src/server/page.tsx
JamieB-gu: > Extracted Emotion into single style tag for hashing

By default Emotion scatters `<style>` tags across the DOM. To prevent the CSP from blocking these we'd need to somehow find all of these tags and hash their content. Achieving this would mean using one of a number of techniques, e.g. walking the React tree, applying string searches/regexes to the HTML string, or parsing the whole tree into a DOM and extracting them using the DOM APIs. I felt these all had greater complexity/performance downsides than using the built-in Emotion API to extract the CSS.

src/server/csp.ts
JamieB-gu: Maybe, their client-side platform is more complex than ours.

JamieB-gu:
The CSP was breaking client-side code in development (probably not in production) because webpack's default sourcemap in [`development` mode is `eval`](https://webpack.js.org/configuration/mode/#mode-development). This [`devtool` setting](https://webpack.js.org/configuration/devtool) includes `eval` in the outputted client bundle, which is not allowed by our CSP.

I could have applied different CSPs for DEV and PROD, but I felt that would likely cause debugging difficulties in future. Therefore I've made use of a different (and probably more useful) sourcemap for development.

PR #362, opened 2020-06-03
Use image component for avatar and cutout images

webb04:
## Why are you doing this?

Use the `` component for cutout and avatar images. This way we can get the benefits of the picture element, the srcset and dpr queries.

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/contributor.ts
+++ b/src/contributor.ts
@@ -18,7 +18,7 @@ type Contributor = {
 
 // ----- Functions ----- //
 
-const contributorSrcset = srcsetWithWidths([32, 64, 128, 192, 256]);
+const contributorSrcset = srcsetWithWidths([32, 64, 128, 192, 256, 400, 600]);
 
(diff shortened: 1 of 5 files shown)


src/contributor.ts
```diff
@@ -34,6 +31,14 @@ const parseContributors = (salt: string, content: Content): Contributor[] =>
         image: fromNullable(contributor.bylineLargeImageUrl).fmap(url => ({
             srcset: contributorSrcset(url, salt, Dpr.One),
             src: src(salt, url, 64, Dpr.One),
+            dpr2Srcset: contributorSrcset(url, salt, Dpr.Two),
```
JamieB-gu, line 34: We might want to offer a couple of higher res widths on line 21 in this file. The cutout is somewhat larger than the avatar and so could probably benefit on DPR 2 screens? Maybe `400` and `600`?

PR #367, opened 2020-06-04
Remove launch slideshow class from contributor images

webb04:
## Why are you doing this?

We only want body and header images in the native slideshows.

### DIFF
--- a/src/components/img.ts
+++ b/src/components/img.ts
@@ -38,7 +38,7 @@ const Img: FC<Props> = ({ image, sizes, className }) =>
         styledH('img', {
             src: image.src,
             alt: image.alt.withDefault(''),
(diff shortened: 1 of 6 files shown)


src/components/img.ts
```diff
@@ -38,7 +38,7 @@ const Img: FC<Props> = ({ image, sizes, className }) =>
         styledH('img', {
             src: image.src,
             alt: image.alt.withDefault(''),
-            className: 'js-launch-slideshow',
+            className: image.launchSlideshow ? 'js-launch-slideshow' : null,
```
JamieB-gu, line 41: I have a vague memory that this might appear in the DOM with the attribute `"null"` or `""` or something, but I may be completely wrong in which case please ignore 🤔

PR #368, opened 2020-06-05
more csp fixes

webb04:
## Changes

- Dark mode cutout image fix
- theguardian.com, TikTok, Spotify and YouTube csp fixes

## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |
|  |  |
|  |  |

### DIFF
--- a/src/server/csp.ts
+++ b/src/server/csp.ts
@@ -54,7 +54,7 @@ const buildCsp = ({ styles, scripts }: Assets, twitter: boolean): string => `
     style-src ${assetHashes(styles)} https://interactive.guim.co.uk ${twitter ? 'https://platform.twitter.com' : ''};
     img-src 'self' https://*.guim.co.uk ${twitter ? 'https://platform.twitter.com https://syndication.twitter.com https://pbs.twimg.com data:' : ''};
     script-src 'self' ${assetHashes(scripts)} https://interactive.guim.co.uk https://s16.tiktokcdn.com https://www.tiktok.com/embed.js ${twitter ? 'https://platform.twitter.com https://cdn.syndication.twimg.com' : ''};
-    frame-src https://www.tiktok.com https://interactive.guim.co.uk https://open.spotify.com https://www.youtube-nocookie.com ${twitter ? 'https://platform.twitter.com https://syndication.twitter.com https://twitter.com' : ''};
+    frame-src https://www.theguardian.com https://www.tiktok.com https://interactive.guim.co.uk https://open.spotify.com https://www.youtube-nocookie.com ${twitter ? 'https://platform.twitter.com https://syndication.twitter.com https://twitter.com' : ''};
     font-src 'self' https://interactive.guim.co.uk;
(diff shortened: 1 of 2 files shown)


webb04:
Adding the multiple TikTok urls seem a little brittle. We also need to add Facebook and Instagram and there's probably a few more.

I've left the csp used on dotcom as a comment `const cspString`.

What are your thoughts @faresite @alexduf ?

webb04:
This could be interesting, we could measure what we need to support first https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only

src/server/csp.ts
```diff
@@ -45,14 +45,18 @@ const interactiveAssets = compose(extractInteractiveAssets, getValidElements);
...
     img-src 'self' https://*.guim.co.uk ${twitter ? 'https://platform.twitter.com https://syndication.twitter.com https://pbs.twimg.com data:' : ''};
-    script-src 'self' ${assetHashes(scripts)} https://interactive.guim.co.uk ${twitter ? 'https://platform.twitter.com https://cdn.syndication.twimg.com' : ''};
-    frame-src https://interactive.guim.co.uk ${twitter ? 'https://platform.twitter.com https://syndication.twitter.com https://twitter.com' : ''};
+    script-src 'self' ${assetHashes(scripts)} https://interactive.guim.co.uk https://s16.tiktokcdn.com https://www.tiktok.com/embed.js ${twitter ? 'https://platform.twitter.com https://cdn.syndication.twimg.com' : ''};
+    frame-src https://www.theguardian.com https://www.tiktok.com https://interactive.guim.co.uk https://open.spotify.com https://www.youtube-nocookie.com ${twitter ? 'https://platform.twitter.com https://syndication.twitter.com https://twitter.com' : ''};
     font-src 'self' https://interactive.guim.co.uk;
-    connect-src 'self' https://interactive.guim.co.uk
+    connect-src 'self' https://interactive.guim.co.uk https://sf-hs-sg.ibytedtos.com/
```
alexduf, line 59: What's this one?

webb04: That's required by TikTok to render an embed. It just feels like it would be easy to break if they start depending on another url we don't support.

gu.com/commentisfree/2020/may/16/tiktok-perfect-medium-splintered-attention-spans-coronavirus-lockdown

alexduf: Agreed it's a shame but there's not much we can do. If they want their content to be embedded they have to provide stable URLs. That domain name looks really dodgy...

JamieB-gu: Does the embed actually break if we don't allow this URL?

webb04: yes it was required for the embed on this article :( gu.com/commentisfree/2020/may/16/tiktok-perfect-medium-splintered-attention-spans-coronavirus-lockdown

It returned a two field JSON blob with an API version...

PR #370, opened 2020-06-08
Explainer atom

webb04:
## Changes

- Add `atoms-rendering` dependency
- Implement `ExplainerAtom`
- Refactored some emotion cache to be able to use `atoms-rendering`. Now matches dotcom

## Screenshots

Screen Shot 2020-06-08 at 11 02 40


src/bodyElement.ts
```diff
@@ -307,20 +313,50 @@ const parse = (context: Context, atoms?: Atoms) =>
                 return new Err('No atom data returned by capi')
             }
 
-            const id = element.contentAtomTypeData?.atomId
-            const atom = atoms.interactives?.find(interactive => interactive.id === id);
+            switch (element.contentAtomTypeData?.atomType) {
+                case "interactive": {
```
alexduf, line 317: That feels like a lot to put in that switch statement. Would extracting this into its own function make it more readable?

PR #374, opened 2020-06-09
liveblog style fixes

webb04:
## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/client/liveblog.ts
+++ b/src/client/liveblog.ts
@@ -42,7 +42,7 @@ fetch(`${window.location}/live-blocks`)
     .then(({ newBlocks }) => {
         const blocks = deserialise(newBlocks);
 
-        ReactDOM.render(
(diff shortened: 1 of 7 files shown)


alexduf:
(There's a bit of linting to fix though)

src/client/liveblog.ts
```diff
@@ -42,7 +42,7 @@ fetch(`${window.location}/live-blocks`)
     .then(({ newBlocks }) => {
         const blocks = deserialise(newBlocks);
 
-        ReactDOM.render(
+        ReactDOM.hydrate(
```
JamieB-gu, line 45: Heads up that, as discussed [here](https://github.com/guardian/apps-rendering/pull/345#discussion_r430994045), there's no guarantee that the data matches what was available on the server, which may result in rendering bugs when using `hydrate`.

PR #375, opened 2020-06-09
Use generated capi models

alexduf:
## Why are you doing this?
So we can use properly generated models that don't have to sit in the same repository and that can be shared by the business.

## Changes

- Use the generated models
- Add support for getting the stack trace of an exception in production (change in the webpack config + new library)
- The thrift decoders had to be made async when using the official thrift library

The difference between the models generated by credit karma and ours is minimal, it was mainly:
 - imports
 - union types
 - different int64 type


src/server/decoders.ts
```diff
@@ -20,16 +15,29 @@ interface ThriftDecoder<A> {
...
+        return decoder.read(protocol);
+    } else {
+        return Promise.reject("Invalid request")
+    }
 }
 
-const capiDecoder = decodeContent(CapiResponse);
-const errorDecoder = decodeContent(ErrorResponse);
-const mapiDecoder = decodeContent(MapiContent);
+const capiDecoder = decodeContent(ItemResponseSerde);
```
webb04, line 38: Serde?

src/server/server.ts
```diff
@@ -1,5 +1,7 @@
 // ----- Imports ----- //
 
+import 'source-map-support/register'; // activating the source map suport
```
webb04, line 3: What does the source map do?

src/server/decoders.ts
alexduf: It stands for "SERialisation DEserialisation"

Like Modem: MOdulator DEModulator
Or Codec: COder DECoder

src/server/server.ts
alexduf: This intercepts exceptions and "beautify" them with the source map so that in our logs we can get the actual source code line rather than the bundle line.

alexduf:
Works in CODE.

_happy dance_

src/server/server.ts
webb04: ok 👍

webb04:
```suggestion
import 'source-map-support/register'; // activating the source map support
```

alexduf:
Thanks :)

webpack.config.js
```diff
@@ -91,6 +91,10 @@ const serverConfig = env => {
                 }
             ]
         },
+        optimization: {
+            minimize: false
```
JamieB-gu, line 95: Is this to improve source-mapping/stack traces?

PR #378, opened 2020-06-09
disable increasing scale

webb04:
## Why are you doing this?

This is recommended by lighthouse for accessibility but we have other features like text size increasing and galleries (pinch and zoom)

At the moment you can pinch and zoom, this will misalign adverts and other native content.


JamieB-gu:
I wonder if we should change the "text size increase" icon (`Aa`) to a "zoom" icon (🔎). Perhaps that more accurately describes its behaviour now, and will make it more obvious to users who want to zoom? @benwuersching @zeek01 🤔

PR #380, opened 2020-06-10
use touchstart

webb04:
## Changes

- Change 1
- Change 2

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/client/nativeCommunication.ts
+++ b/src/client/nativeCommunication.ts
@@ -57,9 +57,9 @@ function ads(): void {
             Array.from(document.querySelectorAll('.ad-placeholder'))
                 .forEach(placeholder => placeholder.classList.remove('hidden'))
             insertAds();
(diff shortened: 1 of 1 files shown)


src/client/article.ts
```diff
@@ -71,7 +71,7 @@ function topics(): void {
     const topic = getTopic(follow);
 
     if (topic) {
-        follow?.addEventListener('click', topicClick);
+        follow?.addEventListener('touchstart', topicClick);
```
JamieB-gu, line 74: Is this [supported by Safari](https://developer.mozilla.org/en-US/docs/Web/API/Element/touchstart_event#Browser_compatibility)?

webb04: Tested on my iOS device and it's working. Will make it harder to test on web browsers as there's no touch events, but we're not mocking Bridget anyway at the moment so maybe that's not important.

PR #381, opened 2020-06-10
use the current src for image

webb04:
## Why are you doing this?

We currently pass in low quality images for the gallery view.

This should pass the selected src from the Image srcset.

### DIFF
--- a/src/client/nativeCommunication.ts
+++ b/src/client/nativeCommunication.ts
@@ -67,11 +67,17 @@ function ads(): void {
     })
 }
 
+function getImageUrl(image: Element): string | null {
+    return (image instanceof HTMLImageElement)
+        ? image.currentSrc
+        : image.getAttribute('src')
+}
+
 function launchSlideshow(src: string | null): void {
     const images = Array.from(document.querySelectorAll('.js-launch-slideshow'));
     const title = document.title;
     const imagesWithCaptions: Image[] = images.flatMap((image: Element) => {
-        const url = image.getAttribute('src');
+        const url = getImageUrl(image);
         const caption =  image.getAttribute('data-caption') ?? undefined;
         const credit = image.getAttribute('data-credit') ?? undefined;
         return url ? new Image({ url, caption, credit }) : [];
@@ -85,8 +91,8 @@ function launchSlideshow(src: string | null): void {
 function slideshow(): void {
     const images = document.querySelectorAll('.js-launch-slideshow');
     Array.from(images)
-        .forEach((image: Element) => image.addEventListener('click', (e: Event) => {
-            launchSlideshow((image as HTMLImageElement)?.currentSrc ?? image.getAttribute('src'));
+        .forEach((image: Element) => image.addEventListener('touchstart', (e: Event) => {
+            launchSlideshow(getImageUrl(image))
         }));
 }
 
--- a/src/client/nativeCommunication.ts
+++ b/src/client/nativeCommunication.ts
@@ -67,20 +67,18 @@ function ads(): void {
     })
 }
 
-function getImageUrl(image: Element): string | null {
-    return (image instanceof HTMLImageElement)
-        ? image.currentSrc
-        : image.getAttribute('src')
-}
-
 function launchSlideshow(src: string | null): void {
     const images = Array.from(document.querySelectorAll('.js-launch-slideshow'));
     const title = document.title;
     const imagesWithCaptions: Image[] = images.flatMap((image: Element) => {
-        const url = getImageUrl(image);
-        const caption =  image.getAttribute('data-caption') ?? undefined;
-        const credit = image.getAttribute('data-credit') ?? undefined;
-        return url ? new Image({ url, caption, credit }) : [];
+        if (image instanceof HTMLImageElement) {
+            const url = image?.currentSrc ?? image.src;
+            const caption =  image.getAttribute('data-caption') ?? undefined;
+            const credit = image.getAttribute('data-credit') ?? undefined;
+            return new Image({ url, caption, credit });
+        } else {
+            return [];
+        }
     });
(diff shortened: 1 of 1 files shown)


webb04:
I think we should be using a higher quality image for the gallery view (can move to horizontal and pinch/zoom) so maybe it's worth adding 

`image.currentSrc.replace(`quality=45`, `quality=85`)`
or 
`image.currentSrc.replace(`quality=45`, `quality=100`)`

JamieB-gu:
I think property you're going to want to change to achieve that is the width rather than the `quality`. If the gallery view is going to require images of different dimensions then we may want to pass through the URL without the width, then allow the native code to apply a width that best suits its requirements.

webb04:
```
function updateUrl(url: string): string {
    const regex = /width=[0-9]+/
    const width = parseInt(url.match(regex)?.pop()?.split('width=')[1] ?? "0")
    const deviceWidth = Math.max(screen.height * 2, screen.width * 2, width);
    return url
        .replace(regex, `width=${deviceWidth}`)
        .replace('quality=45', 'quality=85')
}
```

JamieB-gu:
That's an option, although if we're going to go down that path can I recommend something like this over the use of regex?

```ts
function nativeImage(src: string, dpr: 1 | 2): string {
  const url = new Url(src);
  const width = parseInt(url.searchParams.get('width') ?? '0');

  url.searchParams.set('width', Math.max(screen.height * dpr, screen.width * dpr, width));
  return url.href;
}
```

Again, I don't think we want to alter the `quality`, those values were chosen for the best balance of performance and presentation.

webb04:
Happy to use `searchParams` over regex 👍 This would be done client side so `dpr` wouldn't be available, but maybe it should be for the multiplier or we use `window.devicePixelRatio`. Also this is only altering the image urls passed to Android/iOS for the gallery view, so the webview would still be using `quality=45`.

I think users of the gallery would accept a larger image to be able to zoom in on details.

JamieB-gu:
> This would be done client side so `dpr` wouldn't be available, but maybe it should be for the multiplier or we use `window.devicePixelRatio`

Yeah the method by which we derive the DPR value is up for grabs, but `window.devicePixelRatio > 1.25` is probably a good bet ([1.25 is derived from this](https://github.com/guardian/apps-rendering/blob/888e53976a23fe00f97a871dc0b9ab78c739607a/src/components/img.ts#L35)).

> Also this is only altering the image urls passed to Android/iOS for the gallery view

Yup I'm aware, but I still think we should stick with `45` for DPR 2 and `85` for DPR 1.

> users of the gallery would accept a larger image in the gallery view

What do you mean by a larger image?

webb04:
I can notice the difference between quality `45` and `85` here 

https://i.guim.co.uk/img/media/ed1c3c08bc27b6eb2ea3c16f41196d2bb09f4578/0_97_3500_2100/master/3500.jpg?width=1792&quality=45&fit=bounds&sig-ignores-params=true&s=0c71140c42722a06f82ce9e060dac456

larger image size*

JamieB-gu:
> I can notice the difference between quality `45` and `85` here

I'd hope so, otherwise that `quality` param wouldn't be doing its job 🙂


> larger image size

Ah I see. Perhaps, it's a reasonable hypothesis. How much of an edge are we talking about here? Instances of opening the gallery and then pinching to zoom in?

@paperboyo what do you think?

paperboyo:
Some loose thoughts:

> so maybe it's worth adding `"quality=100"`

Never under any circumstances! This is ☠️ . 😄 

Indeed, if Lightbox in galleries allows pinch-to-zoom, it may make sense to load a higher resolution asset (makes no sense to load same res, but higher `quality`). But:
- it will take time, so Lightbox should either cache it before (?) or first show an existing image and then replace it with higher res when available (or only on pinch in).
- these higher resolutions would better be compressed even more than `45`. We never established a good quality values beyond dpr `1` & `2` (remember, different for jpeg and WebP via VCL, IIRC), so that would require extra work.

So this sounds like a lot of extra logic to me.

If this is about images in gallery and not about them in Lightbox, I wouldn’t bother and would leave them the same quality/size as in normal articles.

In theory, we could make dynamic requests based on changing zoom ratio with viariable `dpr`, but we would need to dynamically establish separate quality curves for jpeg and WebPs and would need to set some ceiling to avoid absurdities and would need to figure out how to pre-cache or load tiles or… you know. Let‘s forget about what I just wrote ;-).

JamieB-gu:
> If this is about images in gallery and not about them in Lightbox

Sorry, I think we have the same word for too many things here 🙃. I believe "gallery" in this context is the name for the app's version of Lightbox, rather than Gallery articles. So we're talking about what happens when you tap on an image to bring up the lightbox/slideshow view in *all* articles on the app.

JamieB-gu:
> these higher resolutions would better be compressed even more than `45`

I think there are two situations we have here:

1) Opening images fullscreen in the slideshow/lightbox view. We should use widths for these images that correspond to the size of the lightbox, so "screen width" for DPR 1, "2x screen width" for DPR 2.
2) Zooming in on images that are fullscreen in the slideshow/lightbox. I suppose we could experiment with providing even higher resolutions with lower qualities for this, but I think it might bring too little benefit for a) the amount of work and b) the cache fragmentation if we start to vary it arbitrarily.

paperboyo:
Wot @JamieB-gu said. FYI, for dotcom, anything that reports dpr higher than 1,25 is treated as dpr=2. Even if it reports dpr 1978 :-).

I would go for: build and see how it looks and behaves. If ppl (us!) are unhappy and pinch to zoom a lot and see pixels and artifacts, we may reconsider.

As an aside, the latest Daily betas have their own straight from the oven implementation of Lightbox behind the switch in Settings. Have a look if you like. Some behaviours should be common between products, in my humble opinion. If not all.

PR #383, opened 2020-06-10
Use click over touchstart

webb04:
## Why are you doing this?

The majority of the page is image, which makes it very hard to scroll down and see the content. Trying to launches the gallery slideshow https://gu.com/us-news/2020/jun/10/george-floyd-protests-what-happens-next


webb04:
I’ll check but I think it could be, click might be ok for images.

webb04:
Changing to `click` solves the problem. A scroll action doesn't open the lightbox

PR #384, opened 2020-06-11
Live Event link component

webb04:
## Why are you doing this?

You can see a membership link on this article (with the old Guardian logo?) https://www.theguardian.com/world/2020/may/29/best-of-a-bad-situation-covid-19-silver-linings

I've used the current related links style but maybe we want a different design for membership events? "Read more" might not be right? We also get an image from Capi that we could display @benwuersching 

I've made a PR to source to add the Guardian Live logo. If that gets approved I can update the logos. https://github.com/guardian/source/pull/394

## Screenshots
Screen Shot 2020-06-22 at 09 45 17


src/bodyElement.ts
```diff
@@ -278,6 +283,16 @@ const parse = (context: Context, atoms?: Atoms) =>
             return new Ok({ kind: ElementKind.Embed, html: embedHtml, alt: fromNullable(alt) });
         }
 
+        case ElementType.MEMBERSHIP: {
+            const { linkText, originalUrl: url } = element.membershipTypeData ?? {};
+
+            if (!linkText || !url) {
+                return new Err('No title or href field on membershipTypeData')
```
JamieB-gu, line 290:
```suggestion
                return new Err('No linkText or originalUrl field on membershipTypeData');
```

src/bodyElement.ts
```diff
@@ -27,6 +27,7 @@ const enum ElementKind {
     Instagram,
     Audio,
     Embed,
+    Membership,
```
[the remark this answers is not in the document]
webb04: This is also called `Membership` in the capi models

webb04:
What they currently look like in the templates:
IMG_5412

src/bodyElement.ts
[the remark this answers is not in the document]
JamieB-gu: Yup, I'm not surprised, but I would argue that's out-of-date 🙂

JamieB-gu:
> What they currently look like in the templates

Ah ok, so pretty similar. I think we probably don't want the word "Membership" in there, but maybe we want to keep "Event". Any thoughts @benwuersching @zeek01?

webb04:
If we go for a new design it might be useful for dcr? @gtrufitt @oliverlloyd

benwuersching:
@webb04 @JamieB-gu Hello Jamies, I tried something out for both dotcom and app: https://www.figma.com/file/3ux4EJLTKAO6fCUywr8eVs/Guardian-Live-rich-link?node-id=0%3A1. To be discussed in more detail (e.g., apart from the design itself, how to label this, how much info to show on the app etc.).

benwuersching:
One question regarding the above screenshots: Can the "Guardian" logo in the pink header area be replaced by a "Guardian Live" logo?
image

webb04:
> One question regarding the above screenshots: Can the "Guardian" logo in the pink header area be replaced by a "Guardian Live" logo?
> image

We've just merged in a PR to add the Guardian Live logo to Source, so I will swap to using that one when the next Source version is available.

https://github.com/guardian/source/pull/394#pullrequestreview-432420278

webb04:
> Can we use the term `LiveEvent` instead of `Membership`? I think we should be reluctant to introduce _new_ references to the membership product going forward 😬.

If we use`LiveEvent` we will have two references for the same thing `Membership` from capi and then `LiveEvent`?

src/components/membershipLink.tsx
```diff
@@ -0,0 +1,108 @@
...
+            a, h1, button {
+                color: ${neutral[60]};
+            }
+        `}
+    `;
+
+const MembershipLink = (props: MembershipLinkProps): ReactElement => {
+    const { url, image, linkText, start, price } = props;
+    const headerImage = image ? <img src={image} alt="Membership event"/> : null;
+    const date = start ? formatDate(new Date(start?.iso8601)) : null;
```
JamieB-gu, line 92: Should this happen during the parsing step?

src/bodyElement.ts
```diff
@@ -136,6 +145,8 @@ function toSerialisable(elem: BodyElement): JsonSerialisable {
             return { ...elem };
         case ElementKind.Embed:
             return { ...elem, alt: optionToSerialisable(elem.alt) };
+        case ElementKind.LiveEvent:
+                return { ...elem };
```
JamieB-gu, line 149:
```suggestion
            return { ...elem };
```

JamieB-gu: Also, if it doesn't cause you a type error, you may be able to leave this to the `default` case.

src/bodyElement.ts
```diff
@@ -278,6 +289,31 @@ const parse = (context: Context, atoms?: Atoms) =>
...
+                price,
+                start,
+                image
+            } = element.membershipTypeData ?? {};
+
+            if (!linkText || !url) {
+                return new Err('No linkText or originalUrl field on membershipTypeData');
+            }
+
+            const formattedDate = start ? formatDate(new Date(start?.iso8601)) : undefined;
```
JamieB-gu, line 305: I think this may sidestep the possibility of an invalid date instance, either from `start` being `undefined` or if the `iso8601` field is mangled somehow:

```suggestion
            const date = new Date(start?.iso8601);
            const formattedDate = !isNaN(date) ? formatDate(date) : undefined;
```

src/components/liveEventLink.tsx
```diff
@@ -0,0 +1,105 @@
...
+    css`
+        background: ${neutral[97]};
+
+        a {
+            display: inline-block;
+            text-decoration: none;
+            color: ${neutral[7]};
+
+            div {
+                background: #B84376;
```
JamieB-gu, line 28: Does this exist in any of the Source colour palettes?

src/components/liveEventLink.tsx
```diff
@@ -0,0 +1,105 @@
...
+                    fill: currentColor;
+                }
+            }
+
+            img {
+                width: 100%;
+            }
+    
+            section {
+                padding: ${basePx(1)};
```
JamieB-gu, line 42:
```suggestion
                padding: ${remSpace[2]};
```

src/components/liveEventLink.tsx
```diff
@@ -0,0 +1,105 @@
...
+                ${textSans.xsmall()};
+
+                h1 {
+                    margin: ${basePx(0, 0, 1, 0)};
```
JamieB-gu, line 46:
```suggestion
                    margin: 0 0 ${remSpace[2]} 0;
```

src/components/liveEventLink.tsx
```diff
@@ -0,0 +1,105 @@
...
+                    margin-top: 0;
+                }
+    
+                button {
+                    background: none;
+                    border: none;
+                    ${textSans.small()};
+                    padding: 0;
+                    margin: 0;
+                    background: #B84376;
```
JamieB-gu, line 61: Same question as above about the Source palette 🤔

src/components/liveEventLink.tsx
```diff
@@ -0,0 +1,105 @@
...
+                    color: ${neutral[100]};
+                    border-radius: 24px;
+                    padding: 0 12px;
```
JamieB-gu, line 64:
```suggestion
                    padding: 0 ${remSpace[3]};
```

src/components/liveEventLink.tsx
```diff
@@ -0,0 +1,105 @@
...
+    
+                button {
+                    background: none;
+                    border: none;
+                    ${textSans.small()};
+                    padding: 0;
+                    margin: 0;
+                    background: #B84376;
+                    color: ${neutral[100]};
+                    border-radius: 24px;
```
JamieB-gu, line 63:
```suggestion
                    border-radius: 1.5rem;
```

src/components/liveEventLink.tsx
```diff
@@ -0,0 +1,105 @@
...
+                    padding: 0 12px;
+                }
+            }
+        }
+
+        float: left;
+        clear: left;
+        margin: ${basePx(1, 2, 1, 0)};
```
JamieB-gu, line 71:
```suggestion
        margin: ${basePx(1, 2, 1, 0)};
        margin: ${remSpace[2]} ${remSpace[4]} ${remSpace[2]} 0;
```

src/components/liveEventLink.tsx
```diff
@@ -0,0 +1,105 @@
...
+    const headerImage = image ? <img src={image} alt="Live event"/> : null;
+    return <aside css={liveEventLinkStyles}>
+        <a href={url}>
+            <div><SvgGuardianLiveLogo /></div>
+            { headerImage }
+            <section>
+                <h1>{ linkText }</h1>
+                <time>{ start }</time>
+                <p>{ price }</p>
+                <button>Book now</button>
```
JamieB-gu, line 98: Is this available to us from `src-button`? As in the `LoadMore` component?

src/components/liveEventLink.tsx
webb04: I don't think it does. The closest I can find is lifestyle.400 but I don't think there's a connection there.

I wonder if this colour is used elsewhere and should be added to the design system? @benwuersching @SiAdcock

src/components/liveEventLink.tsx
webb04: We could use that but it might not be worth it as we'd need to define a new theme for the colours. Plus we're not making use of the event listener or icons etc, everything is wrapped inside of an `<a />` tag.

src/components/liveEventLink.tsx
JamieB-gu: If it's the live events brand colour does it exist in a RR palette somewhere?

PR #385, opened 2020-06-11
add dpr to image url

webb04:
## Why are you doing this?

I noticed blurry images even when using `currentSrc` https://github.com/guardian/apps-rendering/pull/381

It looks like we use a lower quality but not a higher dpr. Dotcom use a dpr of 2 and a quality of 85 so should we use that for all of our images?

### DIFF
--- a/src/image.ts
+++ b/src/image.ts
@@ -78,6 +78,7 @@ function src(salt: string, input: string, width: number, dpr: Dpr): string {
     const params = new URLSearchParams({
         width: width.toString(),
         quality: dpr === Dpr.Two ? lowerQuality.toString() : defaultQuality.toString(),
+        dpr: dpr === Dpr.Two ? "2" : "1",
         fit: 'bounds',
         'sig-ignores-params': 'true',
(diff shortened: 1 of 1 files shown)


JamieB-gu:
This is intentional, our images are doubled in width already when we're working with DPR 2. We currently use a slightly different mechanism to dotcom, but it achieves the same result.

Adding the `dpr` parameter would have the net result of *quadrupling* our image widths, which is not something we want for bandwidth reasons. Are you sure dotcom use quality `85` in conjunction with `dpr` 2? It should be `45` for DPR 2 and `85` for DPR 1.

webb04:
ah ok, yep the first image here uses dpr 2 and quality 85 https://www.theguardian.com/commentisfree/2020/jun/11/pandemic-scientists-second-wave-coronavirus

JamieB-gu:
> yep the first image here uses dpr 2 and quality 85

Really? For me that's using `quality=45&dpr=2` 🤔

webb04:
Ok I can't find where that was. closing this.

PR #386, opened 2020-06-11
handle another case of linking to author profiles

webb04:
## Why are you doing this?

Profile links in immersive standfirsts (bylines can be inside) go to https://mobile.guardianapis.com/profile/steve. If we rewrite the relative links to be absolute dotcom links, the apps can handle them and display the author pages.

This might be a bit brittle, so we could:
- Redirect to dotcom in MAPI
- Add logic on iOS/Android to recognise `/profile/<NAME>` links
- Clean all the html from capi (what happened to sharing the cleaner with dotcom?)

http://localhost:8080/society/2020/jun/11/we-locked-down-a-week-early-the-scottish-care-home-with-no-coronavirus-cases

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -242,7 +242,8 @@ const standfirstTextElement = (format: Format) => (node: Node, key: number): Rea
         case 'A': {
             const colour = format.design === Design.Media ? inverted : kicker;
             const styles = css` color: ${colour}; text-decoration: none`;
-            const href = getHref(node).withDefault('');
+            const url = getHref(node).withDefault('')
+            const href = url.startsWith('profile/') ? `https://www.theguardian.com/${url}` : url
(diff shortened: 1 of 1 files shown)


JamieB-gu:
> Clean all the html from capi (what happened to sharing the cleaner with dotcom?)

Yeah we could do this in two steps. First introduce it in our parsing code for testing purposes, and second push that logic upstream to the CAPI cleaner so that we don't have to.

webb04:
> > Clean all the html from capi (what happened to sharing the cleaner with dotcom?)
> 
> Yeah we could do this in two steps. First introduce it in our parsing code for testing purposes, and second push that logic upstream to the CAPI cleaner so that we don't have to.

I've added it to the content wishlist

webb04:
@JamieB-gu is this ok to be merged?

PR #388, opened 2020-06-15
remove background colour on thumbnail images

webb04:
## Changes

- Thumbnail images can be round avatars so we shouldn't add a background colour

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/img.ts
+++ b/src/components/img.ts
@@ -4,9 +4,10 @@ import { createElement as h, FC } from 'react';
 import { css, jsx as styledH, SerializedStyles } from '@emotion/core';
 import { neutral } from '@guardian/src-foundations/palette';
 
-import { Image } from 'image';
+import { Image, Role } from 'image';
 import { darkModeCss } from 'styles';
 import { Format, Design } from '@guardian/types/Format';
+import { Option } from 'types/option';
 
 
 // ----- Component ----- //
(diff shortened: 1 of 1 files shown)


src/components/img.ts
```diff
@@ -18,14 +19,19 @@ interface Props {
...
-`;
+    `
+
+    return role.fmap(imageRole => imageRole).withDefault(Role.HalfWidth) === Role.Thumbnail
+        ? darkModeStyles
+        : css`
+            background-color: ${format?.design === Design.Media ? neutral[20] : neutral[97]};
+            ${darkModeStyles}
+        `
+}
```
JamieB-gu, line 34: I'm finding it a little bit difficult to follow the logic here. Is it worth splitting it into a couple of functions? Something like:

```ts
const backgroundColour = (role: Role, format: Format): Option<SerializedStyles> => {
    if (imageRole === Role.Thumbnail) {
        const colour = format.design === Design.Media ? neutral[20] : neutral[97];
        return new Some(css`background-color: ${colour};`);
    }

    return new None();
}

const darkStyles = darkModeCss`
    color: ${neutral[60]};
    background-color: ${neutral[20]};
`;

const styles = (role: Option<Role>, format: Format): SerializedStyles => css`
    ${role.andThen(backgroundColour).withDefault('')}
    ${darkStyles}
`;
```

PR #389, opened 2020-06-15
use click rather than touchstart

webb04:
## Why are you doing this?

Use click instead of touchstart. Touchstart is more performant and fires quicker but it's easy to trigger when scrolling down the article.

If we detect any slowness related to the Bridget APIs we should probably optimise the thrift message handling code on iOS/Android and apps-rendering.

### DIFF
--- a/src/styles.ts
+++ b/src/styles.ts
@@ -105,7 +105,7 @@ export const adStyles = css`
                 line-height: 24px;
 
                 span {
-                    margin-right: 24px;
+                    margin-right: 30px;
                 }
 
                 &::focus, &::hover, &::active {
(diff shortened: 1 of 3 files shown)


src/styles.ts
```diff
@@ -105,7 +105,7 @@ export const adStyles = css`
                 line-height: 24px;
 
                 span {
-                    margin-right: 24px;
+                    margin-right: 30px;
```
JamieB-gu, line 108: `remSpace`?

src/styles.ts
```diff
@@ -116,7 +116,7 @@ export const adStyles = css`
                     padding-left: ${basePx(1)};
                     ${icons}
                     content: "\\e04F";
-                    font-size: 16px;
+                    font-size: 20px;
```
JamieB-gu, line 119: `rem` for scaling purposes?

src/styles.ts
webb04: The whole advert placeholder is set in pixels and I think the native layers only request ads at a certain pixel. We also have a ticket to use the icons in src instead of fonts (had a quick look into that and it requires a bit of refactoring).

Will merge this in now as it's just changing the alignment slightly.

PR #390, opened 2020-06-15
Performance Metrics

JamieB-gu:
## Why are you doing this?

Making use of some of the [`Performance` APIs](https://developer.mozilla.org/en-US/docs/Web/API/Performance) available in the browser to check how our pages perform. For now this just logs to the console so that we can use it when developing.

**Note:** You will only see the "paint" metrics in Chrome, but you should see font metrics in all modern browsers.

## Changes

- Added a `metrics` module
- Included metrics within `setup` module
- Added tests for `metrics` module

### DIFF
--- a/src/client/setup.ts
+++ b/src/client/setup.ts
@@ -43,7 +43,7 @@ function twitter(): void {
     }
 }
 
-function performanceMetrics() {
+function performanceMetrics(): void {
     window.addEventListener(
         'load',
         () => console.log('Metrics: ', metrics(performance.getEntries())),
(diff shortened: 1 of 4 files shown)


src/client/setup.ts
```diff
@@ -39,7 +43,16 @@ function twitter(): void {
     }
 }
 
+function performanceMetrics(): void {
+    window.addEventListener(
+        'load',
+        () => console.log('Metrics: ', metrics(performance.getEntries())),
```
webb04, line 49: can we use the `logger` module Alex made? it's currently also using `console.log()` but it would be easy for us to swap that to a sentry log or something else

JamieB-gu: Interesting question. Currently we plan to pass this information to the native layers, for them to bundle with the other metric information before sending off.

Do we plan to introduce any client-side logging frameworks?

PR #392, opened 2020-06-16
Filter Lightbox/Slideshow By Width

JamieB-gu:
## Why are you doing this?

Only show images in the lightbox/slideshow if their width is greater than `620`. Brings AR into line with the [logic on dotcom](https://github.com/guardian/frontend/blob/2f8b53c7bac3a9c1933837d93a54809661499064/common/app/model/content.scala#L514).

## Changes

- Applied the slideshow class for `width > 620`
- Removed the `launchSlideshow` property from `Image`

### DIFF
--- a/src/components/img.ts
+++ b/src/components/img.ts
@@ -44,7 +44,7 @@ const Img: FC<Props> = ({ image, sizes, className, format }) =>
         styledH('img', {
             src: image.src,
             alt: image.alt.withDefault(''),
-            className: image.launchSlideshow ? 'js-launch-slideshow' : '',
+            className: image.width > 620 ? 'js-launch-slideshow' : '',
             css: [styles(image.role, format), className],
             'data-caption': image.nativeCaption.withDefault(''),
             'data-credit': image.credit.withDefault(''),
--- a/src/components/liveblog/avatar.test.tsx
+++ b/src/components/liveblog/avatar.test.tsx
@@ -23,7 +23,6 @@ describe('Avatar component renders as expected', () => {
             alt: new None(),
             role: new None(),
             nativeCaption: new None(),
-            launchSlideshow: false
         }),
     }]
     it('Adds correct alt attribute', () => {
--- a/src/contributor.ts
+++ b/src/contributor.ts
@@ -39,7 +39,6 @@ const parseContributors = (salt: string, content: Content): Contributor[] =>
             alt: new None(),
             role: new None(),
             nativeCaption: new None(),
-            launchSlideshow: false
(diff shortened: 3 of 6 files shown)


src/components/img.ts
```diff
@@ -44,7 +44,7 @@ const Img: FC<Props> = ({ image, sizes, className, format }) =>
         styledH('img', {
             src: image.src,
             alt: image.alt.withDefault(''),
-            className: image.launchSlideshow ? 'js-launch-slideshow' : '',
+            className: image.width > 620 ? 'js-launch-slideshow' : '',
```
webb04, line 47: Does this filter out Avatars and Cutouts? We probably also don't want the slideshow for thumbnail images https://www.theguardian.com/media/2020/jun/12/capturing-the-cry-for-change-photographers-on-the-blm-protests

webb04: Also don't we use a width of `500` for the default image?

dskamiotis: Seems to filter out unnecessary objects, layout and positioning seems perfect on my screen

dskamiotis: Layout seems perfect on my end, cutouts seems to filtered and there doesn't seem to be an author

JamieB-gu: This uses the width of the image as specified by the master asset from CAPI. [We specify](https://github.com/guardian/apps-rendering/blob/a01b50bd30993c6a61bf6ef550bfec0a9d1d49f0/src/contributor.ts#L36) a width of `192` for contributor images, so this will filter those out.

Regarding thumbnails, it will include them if their CAPI master asset width is greater than `620`, otherwise it won't. This matches the current behaviour on dotcom and is known to design (cc @paperboyo).

paperboyo: Yep. Thumbnails should be included (unless smaller than 620). Contributor images,badges etc from furniture obviously shouldn't (and their masters can be over 620, so they better be excluded based on something else).

JamieB-gu: @paperboyo we don't derive the widths from master assets for contributor images because we don't have them 🙂. We get `bylineLargeImageUrl` from the tag.

paperboyo: Perfetto!

[Photo essay](https://www.theguardian.com/media/2020/jun/12/capturing-the-cry-for-change-photographers-on-the-blm-protests) is a dud. And someone made a mistake of creating these circular portraits bigger than 620, so they appear in Lightbox. Nothing we can (or should) bo doing in code to prevent that.

PR #393, opened 2020-06-16
Create dependabot.yml

gtrufitt:
## Why are you doing this?

I am adding dependabot to platform and module repos as it should help us handle updates to things like the types modules and source foundations more effectively.


JamieB-gu:
Does this check for *any* outdated dependencies and automatically create PRs?

gtrufitt:
> Does this check for *any* outdated dependencies and automatically create PRs?

Think so, though docs are here. Looks like you can ignore https://help.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#ignore

PR #396, opened 2020-06-18
Pass more data from MAPI

alexduf:
## Why are you doing this?
This makes use of the apps-rendering models, particularly the `RenderingRequest` type.

## Changes

- Import the TS library
- Make use of the rendering request when parsing the client request. This changes the type received by the server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ In order to release the models you'll need to:
 
 In the SBT repl:
 ```sbtshell
+clean
 project scalaApiModels
 release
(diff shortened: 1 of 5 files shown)


alexduf:
To be deployed at the same time as https://github.com/guardian/mobile-apps-api/pull/1609

It will cause a short interruption of service during deployment, we don't mind the service isn't live yet.

alexduf:
Tested on CODE ✔️

PR #397, opened 2020-06-18
Media atoms and HeaderMedia component

webb04:
## Changes

- Render media atoms in article body
- Render media atoms as main media at the top of articles (standard, video, gallery, comment, liveblog)
- Both are placeholders that Bridget will communicate with iOS/Android to populate with YouTube native SDK videos


src/components/headerVideo.tsx
```diff
@@ -0,0 +1,46 @@
...
+        duration
+    }) =>
+        <div
+            css={styles}
+            data-posterUrl={posterUrl}
+            data-videoId={videoId}
+            data-duration={duration}
+        >
+        </div>
+    ).withDefault(<></>)
```
alexduf, line 42: what does this do? `.withDefault(<></>)`

src/video.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    const id = element.contentAtomTypeData?.atomId
+    const atom = atoms.media?.find(media => media.id === id);
+
+    if (atom?.data?.kind !== "media") {
+        return new None();
+    }
+
+    const { posterUrl, duration, assets, activeVersion } = atom?.data?.media;
+    const videoId = assets
```
alexduf, line 24: This logic looks very similar to the one in `atoms.ts`, is this on purpose?

src/components/headerVideo.tsx
webb04: It's almost the same as `.withDefault(null)` but it's to comply with the `JSX.Element` type 

https://reactjs.org/docs/fragments.html

src/video.ts
webb04: Yes it is just parsing the atom data from the `Content` type. Maybe it could be shared, there's two places where we use the atom, in the article body and the main media section at the top.

alexduf: Oh I see, nah it's probably not worth trying to make it generic then, to different use cases that happen to be similar

src/components/headerVideo.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { Video } from 'video';
+
+// ----- Component ----- //
+
+const styles = css`
+    margin: 0 0 ${remSpace[2]} 0;
+    position: relative;
+    display: block;
+    width: 100%;
+    padding-bottom: 56.25%;
```
JamieB-gu, line 18: Is this because all videos will be the same aspect ratio?

src/components/headerVideo.tsx
JamieB-gu: An alternative would be to make the return type `ReactElement` and use `null`.

src/item.ts
```diff
@@ -128,6 +129,8 @@ const itemFields = (context: Context, content: Content): ItemFields =>
         bylineHtml: fromNullable(content?.fields?.bylineHtml).fmap(context.docParser),
         publishDate: maybeCapiDate(content.webPublicationDate),
         mainImage: articleMainImage(content).andThen(parseImage(context)),
+        mainVideo: articleMainVideo(content)
+            .andThen(blockElement => parseVideo(blockElement, content.atoms)),
```
JamieB-gu, line 133: 👍  nice use of `andThen`

src/item.ts
```diff
@@ -26,6 +26,7 @@ interface Fields extends Format {
     bylineHtml: Option<DocumentFragment>;
     publishDate: Option<Date>;
     mainImage: Option<ImageData>;
+    mainVideo: Option<VideoData>;
```
JamieB-gu, line 29: Is it possible for an article to have *both* a main image and a main video? Is it possible for it to have neither? I'm guessing the second may happen (interactive header?) but the first won't?

To prevent invalid states, can we make use of a new type here?

```ts
const enum MainMediaKind {
    Image,
    Video,
}

type MainMedia = {
    kind: MainMediaKind.Image;
    ...videoData;
} | {
    kind: MainMediaKind.Video;
    ...imageData;
}

interface Fields extends Format {
    ...
    mainMedia: Option<MainMedia>;
    ...
}
```

src/components/headerVideo.tsx
webb04: yes they are in the templates.

src/item.ts
webb04: I think it's rare but it can be more than Images and videos. This series uses an Instagram embed 🙃 https://www.theguardian.com/books/booksblog/2020/may/18/tips-links-and-suggestions-what-are-you-reading-this-week.

webb04: I've created this type and extended it to other article types standard, liveblog, video.

src/capi.ts
```diff
@@ -48,8 +51,27 @@ const articleContributors = (content: Content): Tag[] =>
 const isImage = (elem: BlockElement): boolean =>
     elem.type === ElementType.IMAGE;
 
+const isVideo = (elem: BlockElement): boolean =>
+    elem.type === ElementType.CONTENTATOM &&
+    elem.contentAtomTypeData?.atomType === "media";
+
 const articleMainImage = (content: Content): Option<BlockElement> =>
-    fromNullable(content?.blocks?.main?.elements.filter(isImage)[0]);
+    fromNullable(content?.blocks?.main?.elements.filter(isImage)[0])
```
JamieB-gu, line 59: You may want to apply the same logic from `articleMainVideo` here to avoid the possibility of it crashing when doing an index lookup on `undefined`.

src/capi.ts
```diff
@@ -48,8 +51,27 @@ const articleContributors = (content: Content): Tag[] =>
...
+
+const articleMainVideo = (content: Content): Option<BlockElement> =>
+    fromNullable((content?.blocks?.main?.elements.filter(isVideo) ?? [])[0]);
+
+const articleMainMedia = (content: Content, context: Context): MainMedia =>
+    content?.blocks?.main?.elements.filter(isImage)[0]
```
JamieB-gu, line 65: As above regarding index lookups.

src/fixtures/item.ts
```diff
@@ -36,32 +34,38 @@ const fields = {
 const article: Item = {
     design: Design.Article,
     ...fields,
+    mainMedia: { kind: MainMediaKind.Image, image: new None() }
```
JamieB-gu, line 37: Are you able to put this line in the `fields` object above to avoid repeating it each time?

src/capi.ts
```diff
@@ -48,8 +51,27 @@ const articleContributors = (content: Content): Tag[] =>
...
+    elem.contentAtomTypeData?.atomType === "media";
+
 const articleMainImage = (content: Content): Option<BlockElement> =>
-    fromNullable(content?.blocks?.main?.elements.filter(isImage)[0]);
+    fromNullable(content?.blocks?.main?.elements.filter(isImage)[0])
+
+const articleMainVideo = (content: Content): Option<BlockElement> =>
+    fromNullable((content?.blocks?.main?.elements.filter(isVideo) ?? [])[0]);
+
+const articleMainMedia = (content: Content, context: Context): MainMedia =>
```
JamieB-gu, line 64: Would it make sense to have this return `Option<MainMedia>`? So the type becomes:

```ts
type MainMedia =
     { kind: MainMediaKind.Image; image: ImageData } |
     { kind: MainMediaKind.Video; video: VideoData }
```

Either we have an image main media with all the image data to render it, or we don't (likewise with video). That way you can simplify some of the logic in `headerMedia.tsx`, because once you unwrap the `Option` you *know* you have the `VideoData` or `ImageData`.

PR #398, opened 2020-06-19
Dump the ad targeting params into a script tag

alexduf:
## Why are you doing this?
To enable the native side of the apps to access to the targeting data.

## Changes

- Use the RenderingRequest type in the `page` function
- Dump the targeting params in a script tag

## Screenshots
No change (promise!)

### DIFF
--- a/src/server/page.tsx
+++ b/src/server/page.tsx
@@ -14,7 +14,7 @@ import LiveblogArticle from 'components/liveblog/article';
 import Opinion from 'components/opinion/article';
 import Media from 'components/media/article';
 import Interactive from 'components/interactive/article';
-import { Content } from '@guardian/content-api-models/v1/content';
+import { RenderingRequest } from '@guardian/apps-rendering-api-models/renderingRequest';
 import { includesTweets } from 'capi';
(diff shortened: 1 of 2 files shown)


src/server/page.tsx
```diff
@@ -175,13 +175,16 @@ function page(
...
             <head>
-                ${renderToString(<Head webTitle={content.webTitle} />)}
+                ${renderToString(<Head webTitle={renderingRequest.content.webTitle} />)}
                 <meta
                     http-equiv="Content-Security-Policy"
                     content="${cspString}"
                 />
                 <style>${styles}</style>
                 <style data-emotion-css="${ids.join(' ')}">${css}</style>
+                <script id="targetingParams" type="application/json">
```
alexduf, line 185: Should I put that in the body?

JamieB-gu: Semantically I would say the `head` is the best place, as this is metadata about the document. Performance-wise I'm not aware of any major downsides; the browser shouldn't need to do much with this since it's treated as a [data block](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-type).

alexduf: Thanks! 👍

PR #399, opened 2020-06-19
Pass the targeting data to the native side

alexduf:
## Why are you doing this?

To pass all the required ad targeting data to render the ads

## Changes

- Bump Bridget (no one got hurt)
- Parse the targeting data and send it to the native layer

I've "memoised" the function that loads the ad targeting data such that if it's called a second time it doesn't need to parse the JSON again.

## Screenshots
Hasn't changed!

### DIFF
--- a/src/client/nativeCommunication.ts
+++ b/src/client/nativeCommunication.ts
@@ -9,7 +9,8 @@ const getTargetingParams: () => Map<string, string> = memoise(() => {
     const parsed = JSON.parse(content);
     const map: Map<string, string> = new Map();
     for (const key in parsed) {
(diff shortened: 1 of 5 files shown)


src/client/nativeCommunication.ts
```diff
@@ -2,9 +2,24 @@ import { AdSlot } from "@guardian/bridget/AdSlot";
 import { Image } from "@guardian/bridget/Image";
 import { commercialClient, galleryClient, userClient, acquisitionsClient } from "../native/nativeApi";
 import { logger } from "../logger";
+import {memoise} from "../lib";
```
webb04, line 5:
```suggestion
import { memoise } from "../lib";
```

PR #400, opened 2020-06-22
use ellipsis for overflowing tags

webb04:
## Why are you doing this?
The tags on this article were overflowing, causing a white space to the right of the article.
This PR uses ellipsis for long tags.

https://www.theguardian.com/environment/2020/jun/22/uk-arts-leading-figures-join-call-for-green-recovery-from-coronavirus-crisis

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/shared/tags.tsx
+++ b/src/components/shared/tags.tsx
@@ -1,8 +1,9 @@
 import React from 'react';
 import { css, SerializedStyles } from '@emotion/core'
-import { darkModeCss, basePx } from '../../styles';
+import { darkModeCss } from '../../styles';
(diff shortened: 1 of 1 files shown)


src/components/shared/tags.tsx
```diff
@@ -14,9 +14,8 @@ const tagsStyles = (background: string = neutral[97]): SerializedStyles => css`
     ${textSans.medium()}
 
     li {
-        margin: ${basePx(1, 1, .5, 0)};
+        margin: ${basePx(1, 1, 0, 0)};
```
JamieB-gu, line 17: Opportunity to migrate this to `remSpace`?

PR #404, opened 2020-06-23
Functional Option

JamieB-gu:
## Why are you doing this?

`Option` moves from an [ES6 class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes) to a simpler [discriminated union](https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions).

I'm hoping this will provide a few benefits for us. Firstly, we should not longer need to write serialisation/deserialisation code for `Option`, as it will now be comprised of plain JS objects that JSON serialisers understand out of the box. Secondly, it should allow us to make use of `Option` in places like the new [liveblog-rendering](https://github.com/guardian/liveblog-rendering) repo *without* placing a requirement on dotcom to migrate to the framework that comes with it. In practice, this means that they can treat it like a normal JS object if they choose, without worrying about `map`, `andThen` et al. Thirdly, it simplifies the implementation of `Option` itself, which is now made up of a couple of types and a handful of functions, rather than the class-based inheritance model it was using before.

I've also included a new set of `pipe` functions to make working with the new `Option` format easier. These are inspired in part by the elegant [fp-ts](https://github.com/gcanti/fp-ts) library, as well as similar operators that exist in languages like [Elm](https://package.elm-lang.org/packages/elm/core/latest/Basics#|%3E), [Haskell](https://hackage.haskell.org/package/base-4.14.0.0/docs/Prelude.html#v:-36-), [F#](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/symbol-and-operator-reference/) etc.

**Note:** If this is successful I will look to migrating over `Result` next.

FYI @SiAdcock @gtrufitt @dskamiotis @oliverlloyd 

## Changes

- Removed the use of ES6 classes for `Option`; now simple objects
- Migrated `Option` methods to plain functions
- Refactored codebase to use new `Option`
- Removed serialisation code for `Option`

### DIFF
--- a/src/lib.ts
+++ b/src/lib.ts
@@ -1,6 +1,10 @@
 // ----- Functions ----- //
 
 const compose = <A, B, C>(f: (_b: B) => C, g: (_a: A) => B) => (a: A): C => f(g(a));
+const pipe = <A, B>(a: A, f: (_a: A) => B): B => f(a);
+const pipe2 = <A, B, C>(a: A, f: (_a: A) => B, g: (_b: B) => C): C => g(f(a));
+const pipe3 = <A, B, C, D>(a: A, f: (_a: A) => B, g: (_b: B) => C, h: (_c: C) => D): D =>
+    h(g(f(a)));
 
 const identity = <A>(a: A): A => a;
 
@@ -26,6 +30,9 @@ function memoise<A>(fn: () => A): () => A {
 
 export {
     compose,
+    pipe,
+    pipe2,
+    pipe3,
     identity,
     isElement,
     toArray,
--- a/src/bodyElement.ts
+++ b/src/bodyElement.ts
@@ -6,12 +6,13 @@ import { ElementType } from '@guardian/content-api-models/v1/elementType';
 import {
     Option,
     fromNullable,
-    toSerialisable as optionToSerialisable,
+    map,
+    withDefault,
 } from 'types/option';
 import { Result, Err, Ok } from 'types/result';
 import { Context, DocParser } from 'types/parserContext';
 import { Image as ImageData, parseImage } from 'image';
-import { isElement } from 'lib';
+import { isElement, pipe2 } from 'lib';
 import JsonSerialisable from 'types/jsonSerialisable';
 import { parseAtom } from 'atoms';
 import { formatDate } from 'date';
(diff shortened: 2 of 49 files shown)


alexduf:
>as it will now be comprised of plain JS objects that JSON serialisers understand out of the box

So is it serialised like that? 
```json
{
  "kind": "Some",
  "value": "a"
}
```

src/lib.ts
```diff
@@ -1,6 +1,10 @@
 // ----- Functions ----- //
 
 const compose = <A, B, C>(f: (_b: B) => C, g: (_a: A) => B) => (a: A): C => f(g(a));
+const pipe = <A, B>(a: A, f: (_a: A) => B): B => f(a);
+const pipe2 = <A, B, C>(a: A, f: (_a: A) => B, g: (_b: B) => C): C => g(f(a));
```
webb04, line 5: Any chance of making `pipe` variadic?

JamieB-gu: Yeah that's an option (no pun intended), [fp-ts](https://github.com/gcanti/fp-ts/blob/1b0ee81826e93ddd86ce20a3c353d349a6994bb6/src/function.ts#L325) does this using the `arguments` length.

I thought about it, but I wasn't sure about the implementation and if it would cause confusion. What do you think?

JamieB-gu:
> So is it serialised like that?

Almost. It's serialised like this:

```ts
{
  "kind": 0,
  "value": "a"
}

{
  "kind": 1
}
```

due to the way that `const enum`s work (they're stripped out during compilation and replaced with their underlying number equivalent).

> can't we achieve something similar by using `type Option<A> = A | null`?

I'm not sure you know 🤔. *Maybe*. I'd have to think about some of the ramifications. For example, it precludes us ever being able to assign `null` to `A`. Also I'm a little bit wary of blending it with the built-in `null` type for fear of them being assigned to one another implicitly and unexpectedly.

src/lib.ts
alexduf: I was expecting the same as well so that might be a good idea

JamieB-gu:
This PR has a large surface area and I'm starting to hit merge conflicts, so I've merged and will pick up the variadic `pipe` in a separate PR.

PR #416, opened 2020-06-24
GLA-2118: update adverts when font resize

faresite:
## Why are you doing this?

[JIRA ticket](https://theguardian.atlassian.net/browse/GLA-2117?atlOrigin=eyJpIjoiNjMyMzVlNzY3YWFmNDBhMjliMjYwMjczNDBmYTQwOTIiLCJwIjoiaiJ9)

## Changes

- observe `html` mutations instead of `body`
- fliter mutations to `style` only.

## Screenshots

| Before | After |
| --- | --- |
|  | |

### DIFF
--- a/src/client/nativeCommunication.ts
+++ b/src/client/nativeCommunication.ts
@@ -48,7 +48,6 @@ function insertAds(): void {
         commercialClient.insertAdverts(adSlots);
         const targetNode = document.querySelector('html') as Node;
         const config: MutationObserverInit = {
-            attributes: true,
             childList: true,
             subtree: true,
             attributeFilter: ["style"]
--- a/src/client/nativeCommunication.ts
+++ b/src/client/nativeCommunication.ts
@@ -46,8 +46,13 @@ function insertAds(): void {
     let adSlots = getAdSlots();
     if (adSlots.length > 0) {
         commercialClient.insertAdverts(adSlots);
-        const targetNode = document.querySelector('body') as Node;
-        const config = { attributes: true, childList: true, subtree: true };
+        const targetNode = document.querySelector('html') as Node;
+        const config: MutationObserverInit = {
+            attributes: true,
+            childList: true,
+            subtree: true,
+            attributeFilter: ["style"]
+        };
         const callback = function(): void {
             const currentAdSlots = getAdSlots();
             if (JSON.stringify(adSlots) !== JSON.stringify(currentAdSlots)) {
--- a/src/client/nativeCommunication.ts
+++ b/src/client/nativeCommunication.ts
@@ -60,7 +60,7 @@ function insertAds(): void {
             }
         };
 
-        window.addEventListener('orientationchange', callback);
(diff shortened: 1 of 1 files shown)


src/client/nativeCommunication.ts
```diff
@@ -46,8 +46,13 @@ function insertAds(): void {
     let adSlots = getAdSlots();
     if (adSlots.length > 0) {
         commercialClient.insertAdverts(adSlots);
-        const targetNode = document.querySelector('body') as Node;
-        const config = { attributes: true, childList: true, subtree: true };
+        const targetNode = document.querySelector('html') as Node;
+        const config: MutationObserverInit = {
+            attributes: true,
```
webb04, line 51: I'm not sure this is needed with attributeFilter? https://developer.mozilla.org/en-US/docs/Web/API/MutationObserverInit/attributeFilter

faresite: It's not needed for this to work, but if we only care about layout changes this could be a more performant observer than to observe every other attribute? Trying to meet you both in the middle 😅

faresite:
> Nice 🎉
> We might have to test changing orientation on iPads to see if adverts are still ok. I'm not sure this will be caught anymore
> 
> https://github.com/guardian/apps-rendering/blob/9730055830c7d66719bfb6f1f22aae2954f3afe9/src/renderer.ts#L531

Will test this and see 👍

src/client/nativeCommunication.ts
JamieB-gu: @faresite I may be misunderstanding this, but I *think* @webb04 is referring to the fact that the `attributes: true` property is superseded by the `attributeFilter` property? If this is correct then I think we can drop the `attributes` field here because it's applied implicitly by the `attributeFilter` field below. Or am I missing something? 🤔

faresite: ah of course 🤦‍♀️  sorry @webb04 misunderstood you! you're both right.

src/client/nativeCommunication.ts
```diff
@@ -56,6 +60,7 @@ function insertAds(): void {
             }
         };
 
+        window.addEventListener('orientationchange', callback);
```
faresite, line 63: don't love this but that's the only way I managed to get the ads to update on orientation change .. also this won't work until I make a PR iOS side as right now the update adverts code in iOS only updates the `y` position 🤷‍♂️

faresite: https://github.com/guardian/ios-live/pull/5218

webb04: Would this still work?

We'll need to update ad positions when users multi-task multiple apps https://support.apple.com/en-gb/HT207582

```suggestion
        window.addEventListener('resize', callback);
```

faresite: This actually even works better as it repositions the ad immediately instead of after the rotation animation finishes 👍

PR #422, opened 2020-06-25
LiveBlog dark mode list items

dskamiotis:
## Why are you doing this?
Live blog in dark mode was not displaying listItem bullet points clearly, changed value to make lighter in colour and clearer to see.

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -109,10 +109,8 @@ const listStyles: SerializedStyles = css`
     clear: both;
 `;
 
-const listItemStyles = (format: Format): SerializedStyles => {
-    const backgroundColour = format.design === Design.Media
-        ? neutral[46] : neutral[86];
-    return css`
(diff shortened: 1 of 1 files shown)


src/renderer.ts
```diff
@@ -139,6 +137,29 @@ const listItemStyles = (format: Format): SerializedStyles => {
...
+        ${darkModeCss`
+            &::before {
+                background-color: ${neutral[86]};
+            }
+        `}
+    `
+
+    switch (format.design) {
+        case (Design.Media):
+            return [baseStyles, mediaStyles];
```
JamieB-gu, line 157: FYI, as an alternative to returning an array, you could use the `css` function to get `SerializedStyles` as the return type:

https://github.com/guardian/apps-rendering/blob/683fec79cb0be6a9755603757bc7b8148a32562f/src/components/figCaption.tsx#L148

PR #423, opened 2020-06-25
refactor keyline styles

webb04:
## Why are you doing this?
Use a lighter keyline in liveblog dark mode

## Screenshots

| Before | After |
| --- | --- |
|  |  |


webb04:
I just remembered we have a Keyline component in the design system. So it might be useful to migrate to that at some point.

PR #430, opened 2020-06-26
Functional Result

JamieB-gu:
## Why are you doing this?

`Result` moves from an [ES6 class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes) to a simpler [discriminated union](https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions).

This PR is the companion to #404; the reasons for doing this are described in full there.

FYI @SiAdcock @gtrufitt @dskamiotis

## Changes

- Removed the use of ES6 classes for `Result`; now simple objects
- Migrated `Result` methods to plain functions
- Refactored codebase to use new `Result`

### DIFF
--- a/src/bodyElement.ts
+++ b/src/bodyElement.ts
@@ -9,13 +9,14 @@ import {
     map,
     withDefault,
 } from 'types/option';
-import { Result, Err, Ok } from 'types/result';
+import { Result, err, ok, map as rmap } from 'types/result';
 import { Context, DocParser } from 'types/parserContext';
 import { Image as ImageData, parseImage } from 'image';
-import { isElement, pipe2 } from 'lib';
+import { isElement, pipe2, pipe } from 'lib';
 import JsonSerialisable from 'types/jsonSerialisable';
 import { parseAtom } from 'atoms';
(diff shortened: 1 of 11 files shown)


src/bodyElement.ts
```diff
@@ -224,45 +225,47 @@ const parse = (context: Context, atoms?: Atoms) =>
...
             } else if (!h) {
-                return new Err('No "html" field on tweetTypeData')
+                return err('No "html" field on tweetTypeData')
             }
 
-            return tweetContent(id, context.docParser(h))
-                .fmap(content => ({ kind: ElementKind.Tweet, content }));
+            return pipe(
+                tweetContent(id, context.docParser(h)),
+                rmap(content => ({ kind: ElementKind.Tweet, content })),
```
webb04, line 257: rmap?

webb04: could we just use map?

JamieB-gu: `rmap` was my way of specifying `Result.map`, as `map` was already taken in this scope by `Option.map`.

PR #435, opened 2020-06-29
sponsor logos

webb04:
## Why are you doing this?
Render sponsor logos on standard, immersive and opinon articles
https://gu.com/cities/2020/jan/13/the-case-for-never-demolishing-another-building

## Screenshots
The font is no longer working when running local MAPI and apps-rendering servers. This should be ok in production.

| Light | Dark |
| --- | --- |
|  |  |

### Immersive
Screen Shot 2020-06-30 at 10 48 01

### Opinion
Screen Shot 2020-06-30 at 10 52 20


src/components/shared/logo.tsx
```diff
@@ -0,0 +1,65 @@
...
+    branding: Branding;
+    format: Format;
+}
+
+const styles = (lightModeImage: string, darkModeImage?: string): SerializedStyles => css`
+    margin: ${remSpace[9]} 0;
+    ${textSans.small()}
+
+    img {
+        content: url("${lightModeImage}");
```
JamieB-gu, line 24: What was the reason behind using `content` in the CSS over `src` in the `` tag out of curiosity? 🧐

webb04: I wanted to make use of `darkModeCss` to use the alternative logo. I guess this could be done with client side JavaScript but I like having it in css!

They are `static.theguardian.com` urls so maybe? They don't seem to be the best of quality and they can contain random white space which puts our alignment off..

PR #444, opened 2020-06-30
Send Performance Metrics With Bridget

JamieB-gu:
## Why are you doing this?

This uses Bridget to send performance metrics ("first paint", "first contentful paint", "fonts") to the native layers. Based on the work done in [this Bridget PR](https://github.com/guardian/bridget/pull/49).

## Changes

- Refactored `metrics` module to use Bridget types
- Used metrics client in `setup`
- Created metrics client

### DIFF
--- a/src/client/setup.ts
+++ b/src/client/setup.ts
@@ -48,11 +48,13 @@ function twitter(): void {
 function performanceMetrics(): void {
     window.addEventListener(
         'load',
-        () => pipe2(
-            performance.getEntries(),
-            metrics,
-            metricsClient.sendMetrics,
-        ),
+        () => {
+            pipe2(
+                performance.getEntries(),
+                metrics,
+                metricsClient.sendMetrics.bind(null),
+            );
+        },
         { once: true },
     );
 }
--- a/src/client/setup.ts
+++ b/src/client/setup.ts
@@ -2,6 +2,8 @@
 
 import { logger } from 'logger';
 import { metrics } from 'client/metrics';
+import { metricsClient } from 'native/nativeApi';
+import { pipe2 } from 'lib';
 
 
 // ----- Procedures ----- //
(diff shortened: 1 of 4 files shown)


src/client/setup.ts
```diff
@@ -46,7 +48,13 @@ function twitter(): void {
 function performanceMetrics(): void {
     window.addEventListener(
         'load',
-        () => console.log('Metrics: ', metrics(performance.getEntries())),
+        () => {
+            pipe2(
+                performance.getEntries(),
+                metrics,
+                metricsClient.sendMetrics.bind(null),
```
webb04, line 55: why do you need to bind here?

src/client/setup.ts
```diff
@@ -46,7 +48,13 @@ function twitter(): void {
 function performanceMetrics(): void {
     window.addEventListener(
         'load',
-        () => console.log('Metrics: ', metrics(performance.getEntries())),
+        () => {
+            pipe2(
```
webb04, line 52: guess pipe is quite small so doesn't add much to the client side bundle?

src/client/setup.ts
JamieB-gu: ESLint complains about using unbound methods otherwise, due to [this rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/unbound-method.md). I believe the same warning prompted [this](https://github.com/guardian/apps-rendering/blob/27f8f65a54a36fa452dbc183761e7e0755978773/src/server/server.ts#L54).

src/client/setup.ts
JamieB-gu: Yup, I suspect this will pull in the `lib` module because I don't *think* we're tree shaking atm.

PR #445, opened 2020-06-30
implement insert videos and update videos

webb04:
## Changes

- Initialises videos (media atoms) by passing `id`, `duration` and `posterUrl`
- Report video positions to IOS/Android using the `Videos` Bridget client
- Similar to Advert positioning
- Shared `getRect` function
- MutationObserver decoupled from insertAdverts
- MutationObserver responsible for reporting back changes to advert or video position changes


src/client/nativeCommunication.ts
```diff
@@ -127,7 +110,73 @@ function slideshow(): void {
...
+
+    if (!videoSlots) {
+        return [];
+    }
+
+    return Array.from(videoSlots).map(videoSlot => {
+        const slotPosition = videoSlot.getBoundingClientRect();
+        return new VideoSlot({
+            rect: getRect(slotPosition),
+            videoId: videoSlot.getAttribute('data-videoId') ?? '',
```
alexduf, line 124: I think I'd rather have the whole VideoSlot be absent, than having the some properties be filled with empty strings.
It will make the job easier for the native side as they'll have to ignore that video slot anyway.

src/client/nativeCommunication.ts
```diff
@@ -127,7 +110,73 @@ function slideshow(): void {
...
+    const config: MutationObserverInit = {
+        childList: true,
+        subtree: true,
+        attributeFilter: ["style"]
+    };
+    const callback = function(): void {
+        const currentAdSlots = getAdSlots();
+        const currentVideoSlots = getVideoSlots();
+
+        if (JSON.stringify(adSlots) !== JSON.stringify(currentAdSlots)) {
```
alexduf, line 152: Is that a performance risk right here?
Is this executed each time there's a scroll, or each time there's a re-layout?

src/client/nativeCommunication.ts
webb04: yes we make sure we have these fields server side but the browser API can still return `undefined`. It might be worth handling undefined again though.

src/client/nativeCommunication.ts
webb04: At the moment it's only run when a `style` tag is changed (font resizing) but yeah we need to keep an eye on this and possibly debounce it when we extend the mutation observer to watch for other changes.

src/client/nativeCommunication.ts
webb04: Made this a reduce so only slots with `videoId` will be returned

src/client/nativeCommunication.ts
```diff
@@ -117,15 +117,17 @@ function getVideoSlots(): VideoSlot[] {
...
-            duration: parseInt(videoSlot.getAttribute('data-duration') ?? '0')
-        })
-    });
+    return Array.from(videoSlots).reduce((slots: VideoSlot[], elem) => {
+        const slotPosition = elem.getBoundingClientRect();
+        const videoId =  elem.getAttribute('data-videoId') ?? '';
+        const posterUrl = elem.getAttribute('data-posterUrl') ?? '';
+        const duration = parseInt(elem.getAttribute('data-duration') ?? '0');
+        const rect = getRect(slotPosition);
+        if (videoId) {
```
alexduf, line 126: I think you need to remove the default value of all of these, and have a test to check that videoId, posterUrl and duration are all non null.

webb04: We might need to update Bridget to make `duration` optional...

webb04: https://github.com/guardian/bridget/pull/53

webb04: Once in beta, these changes won't be optional though

src/client/nativeCommunication.ts
```diff
@@ -25,52 +41,19 @@ function getAdSlots(): AdSlot[] {
...
             targetingParams
         })
     });
 }
 
 function insertAds(): void {
-    let adSlots = getAdSlots();
-    if (adSlots.length > 0) {
+    const adSlots = getAdSlots();
+    if (adSlots.length) {
```
JamieB-gu, line 55:
```suggestion
    if (adSlots.length > 0) {
```

src/client/nativeCommunication.ts
```diff
@@ -135,7 +118,80 @@ function slideshow(): void {
...
+    return Array.from(videoSlots).reduce((slots: VideoSlot[], elem) => {
+        const slotPosition = elem.getBoundingClientRect();
+        const videoId =  elem.getAttribute('data-videoId');
+        const posterUrl = elem.getAttribute('data-posterUrl');
+        const durationString = elem.getAttribute('data-duration');
+        const rect = getRect(slotPosition);
+        if (videoId && posterUrl) {
+            if (durationString) {
+                const duration = parseInt(durationString);
+                slots.push(new VideoSlot({ rect, videoId, posterUrl, duration }))
```
JamieB-gu, line 137: What happens if `duration` is `NaN`?

src/client/nativeCommunication.ts
```diff
@@ -135,7 +118,80 @@ function slideshow(): void {
...
+                slots.push(new VideoSlot({ rect, videoId, posterUrl }))
+            }
+        }
+        return slots;
+    }, []);
+}
+
+function videos(): void {
+    const videoSlots = getVideoSlots();
+    if (videoSlots.length) {
```
JamieB-gu, line 148:
```suggestion
    if (videoSlots.length > 0) {
```

src/client/nativeCommunication.ts
```diff
@@ -135,7 +118,80 @@ function slideshow(): void {
...
+        const videoId =  elem.getAttribute('data-videoId');
+        const posterUrl = elem.getAttribute('data-posterUrl');
+        const durationString = elem.getAttribute('data-duration');
+        const rect = getRect(slotPosition);
+        if (videoId && posterUrl) {
+            if (durationString) {
+                const duration = parseInt(durationString);
+                slots.push(new VideoSlot({ rect, videoId, posterUrl, duration }))
+            } else {
+                slots.push(new VideoSlot({ rect, videoId, posterUrl }))
```
JamieB-gu, line 139:
```suggestion
                return [ ...slots, new VideoSlot({ rect, videoId, posterUrl, duration }) ];
            } else {
                return [ ...slots, new VideoSlot({ rect, videoId, posterUrl }) ];
```

src/client/nativeCommunication.ts
JamieB-gu: I may have missed something, but why `JSON.stringify` and compare the strings? Why not compare the arrays directly?

webb04:
> I may have missed something, but why JSON.stringify and compare the strings? Why not compare the arrays directly?

Maybe we could compare all the values now so stringily isn't necessary... but a direct comparison doesn't work

[1, 2, 3] === [1, 2, 3]
false

src/client/nativeCommunication.ts
webb04: Maybe we could compare all the values now so stringify isn't necessary... but a direct comparison doesn't work

[1, 2, 3] === [1, 2, 3]
false

JamieB-gu:
> a direct comparison doesn't work

Yeah, JavaScript is great... We'd have to write a quick comparison function to do this, and it would have to handle Object comparison too because that doesn't work either 🤦‍♂️

webb04:
How is this looking? is it worth adding a deep object comparison? or a `Rect` comparison?

JamieB-gu:
> is it worth adding a deep object comparison? or a `Rect` comparison?

Up to you. I think there are a couple of things to take into account:

1) Performance - is it faster to run a comparison function, or faster to run `stringify` and do a string comparison?
2) Robustness - is the `stringify` approach likely to be brittle? AFAIK JS doesn't guarantee the order of keys in an object, so you may get different strings back 🤔.

PR #447, opened 2020-06-30
Pass image height and width to client

webb04:
Fix for GLA-2121 where some images have an incorrect ratio in the slideshow/lightbox view.

### DIFF
--- a/src/client/nativeCommunication.ts
+++ b/src/client/nativeCommunication.ts
@@ -90,13 +90,16 @@ function ads(): void {
     })
 }
 
-function updateUrl(src: string): string {
+function getImageWidth(src: string): number {
     const url = new URL(src);
     const width = parseInt(url.searchParams.get('width') ?? '0');
     const dpr = window.devicePixelRatio >= 1.25 ? 2 : 1;
-    const newWidth = Math.max(screen.height * dpr, screen.width * dpr, width);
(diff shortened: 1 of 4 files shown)


src/client/nativeCommunication.ts
```diff
@@ -108,7 +111,9 @@ function launchSlideshow(src: string | null): void {
             const url = updateUrl(image?.currentSrc ?? image.src);
             const caption =  image.getAttribute('data-caption') ?? undefined;
             const credit = image.getAttribute('data-credit') ?? undefined;
-            return new Image({ url, caption, credit });
+            const width = getImageWidth(url);
+            const height = width * parseFloat(image.getAttribute('data-ratio') ?? '0.56');
```
JamieB-gu, line 115: It's possible height could be `NaN` here. What happens in the Thrift transport layer/on the native side if a number value is `NaN`?

webb04: Could it? `getImageWidth` returns a number and `image.width` `image.height` aren't optional

JamieB-gu: `parseFloat` can return a `NaN`, which could happen, for example, if `data-ratio` gets mangled.

webb04: ok we might have to avoid showing the image without having a height/width.

PR #466, opened 2020-07-03
Render comment count

webb04:
## Screenshots

Screen Shot 2020-07-03 at 15 11 08

### DIFF
--- a/src/components/commentCount.stories.tsx
+++ b/src/components/commentCount.stories.tsx
@@ -1,18 +1,19 @@
 // ----- Imports ----- //
 
 import React, { FC } from 'react';
-import { withKnobs, number, boolean } from '@storybook/addon-knobs';
+import { withKnobs, boolean } from '@storybook/addon-knobs';
 
 import CommentCount from './commentCount';
 import { Pillar, Design, Display } from 'format';
 import { selectPillar } from 'storybookHelpers';
+import { some } from 'types/option';
 
 
 // ----- Stories ----- //
 
 const Default: FC = () =>
     <CommentCount
-        count={number('Count', 1234, { min: 0 })}
+        count={some(1234)}
         pillar={selectPillar(Pillar.News)}
         design={Design.Article}
         display={Display.Standard}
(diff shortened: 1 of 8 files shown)


src/components/commentCount.stories.tsx
```diff
@@ -1,18 +1,19 @@
...
 import { selectPillar } from 'storybookHelpers';
+import { some } from 'types/option';
 
 
 // ----- Stories ----- //
 
 const Default: FC = () =>
     <CommentCount
-        count={number('Count', 1234, { min: 0 })}
+        count={some(1234)}
```
JamieB-gu, line 16: Does this drop the control for choosing the number of comments? Could that be useful for testing UIs against small or large comment counts?

src/components/commentCount.tsx
```diff
@@ -5,15 +5,16 @@ import { css, SerializedStyles } from '@emotion/core';
...
 import { getPillarStyles } from 'pillarStyles';
+import { pipe2 } from 'lib';
 
 
 // ----- Component ----- //
 
 interface Props extends Format {
-    count: number;
+    count: Option<number>;
     commentable: boolean;
```
JamieB-gu, line 18: Will `commentable: false` always correspond to `count: None`? Are there situations where there may be comments even when `commentable` is set to `true`? Or situations where the count is `None` (rather than `0`), even when `commentable` is set to `true`? 🤔

I'm asking because we've now got two variables that represent whether to show a comment count, and perhaps `count` is all we need?

src/components/commentCount.stories.tsx
webb04: Yeah it might be nice to keep, did you come across a way of changing optional values?

src/components/commentCount.tsx
webb04: That's a good point, we've used `commentable` in a few other places including whether to render discussion-rendering or not. I think it's also used to filter out articles in mapi.

It probably can be refactored though so that we just use one.

src/components/commentCount.stories.tsx
JamieB-gu: 🙂

```tsx
<CommentCount
  count={some(number('Count', 1234, { min: 0 }))}
/>
```

webb04: https://github.com/guardian/apps-rendering/pull/473

PR #472, opened 2020-07-06
discussion-rendering

webb04:
### Update 16/11/2020
All the callbacks have been added to make requests through the native layers to the discussion-rendering service.
- [X] recommend
- [x] comment
- [x] reply
- [x] preview

## Why are you doing this?
This is a first attempt to integrate discussion-rendering into apps-rendering. It doesn't require much code to integrate but there's probably some cases we might need to discuss/work on.

### Pros
- Shareable code (parity between platforms, reduced duplication)
- Lots of new functionality for apps (sorting of comments etc)
- Would be able to delete a lot of native code

### Cons
- Dark mode not supported, we might need to add a new prop to opt-in and dark styles for all components
- Quite a few links go to theguardian.com, is this acceptable in app?
- We lose top 3 server-side rendered comments which is the current behaviour on apps. Maybe this could be a placeholder before rendering `discussion-rendering`
- Requires `regenerator-runtime` polyfill for AR
- This polyfill applies inline styles, which causes problems for our csp
- More work needed to implement, posting, reporting comments, recommending
  - We either use the cookie approach currently set (would we need cookie banners in app?)
  - Or we abstract `discussion-rendering` so that we can use it with Bridget and the native layers handle authentication
- `onPermaLink` needs to be implemented
- Had to add `@guardian/src-link` and `@guardian/src-text-input` as dependencies to get `discussion-rendering` working

## Screenshots

Screen Shot 2020-07-06 at 12 29 57

Screen Shot 2020-07-06 at 12 30 18


webb04:
Could we add some abstractions to `discussion-rendering` that would avoid us setting cookies (as we don't have any cookie banners on apps-rendering)? 

That could be an `onInit` for setting cookies, and `onPost`, `onReport`, `onReply`, `onRecommend`. Then we could either use the cookies or use `Bridget` so that the native layers can handle the authentication?

@gtrufitt @JamieB-gu @alexduf

JamieB-gu:
This is amazing, very excited for this 🎉!

I haven't had a chance to look through the code yet, but a couple of questions did occur to me reading the pros and cons you laid out (very comprehensive btw 👍):

> Would be able to delete a lot of native code

Should we tag iOS and Android on this PR to start that conversation?

> Quite a few links go to theguardian.com, is this acceptable in app?

My guess is that we'll want these translated to MAPI URLs? Are these links that are posted by users, or are they part of the UI?

If part of the UI I think this is something we should handle (e.g. relative URLs or a domain flag).

If posted by users that's potentially trickier, as we'd probably have to re-write all the ones that point to dotcom. Would that be problematic in terms of accurately representing data they've submitted?

> Requires regenerator-runtime polyfill for AR

Why is this required? Is it due to the way `discussion-rendering` is built (i.e. does it transpile `async/await` before publishing)?

> Had to add @guardian/src-link and @guardian/src-text-input as dependencies

Are they required peer-dependencies?

webb04:
> My guess is that we'll want these translated to MAPI URLs? Are these links that are posted by users, or are they part of the UI?

I think some might be ok to be left as dotcom links, I don't think we support profile pages for examples so that could be a feature enhancement but we can go through these case by case.

As part of the abstraction work away from cookies, we would probably need to update some like clicking "reply" when signed out - we wouldn't want to go to dotcom in the app.

> Why is this required? Is it due to the way discussion-rendering is built (i.e. does it transpile async/await before publishing)?

I'm not too sure, it might be supported in dotcom-rendering babel targets but not ours?

> Are they required peer-dependencies?

They are peer dependencies in yarn but yarn install didn't pull them in for me

JamieB-gu:
> are they part of the UI

> we would probably need to update some like clicking "reply" when signed out

Ah, so part of the UI. Ok, this shouldn't be as difficult to change because it can be a property we pass.

> it might be supported in dotcom-rendering babel targets but not ours?

In the specific case of `async`/`await` I'd be surprised; we have a smaller and more recent browser support list that DCR. I believe `regenerator-runtime` is an optional peer dependency of `@babel/preset-env`, which uses it along with `core-js` to polyfill/transform missing features, based on what you use and your browser support list. Not 100% on this though, there are some contradictions in the Babel docs that have confused me (e.g. suggestions to use `babel-polyfill`, and then other places that say don't because it's deprecated).

> They are peer dependencies in yarn but yarn install didn't pull them in for me

Not sure about yarn, but more recent versions (3+) of npm don't automatically install peer dependencies. If these two packages are *required* for functionality to work should they instead be *required* dependencies?

gtrufitt:
## Requires regenerator-runtime polyfill for AR

`regenerator-runtime` is my fault - we don't (currently) do any bundling on discussion-rendering, instead opening up pure source to the npm module (this is probably a mistake and I was planning to move it to use microbundle). 

In dotcom-rendering we avoid excluding discussion-rendering from bundling: https://github.com/guardian/dotcom-rendering/blob/e9741e59c9120198fc1fc76ebefeeb3d5427c0de/scripts/webpack/browser.js#L76 and so the source gets bundled as any other component would in your application.

## Peer dependencies 

> Not sure about yarn, but more recent versions (3+) of npm don't automatically install peer dependencies. If these two packages are required for functionality to work should they instead be required dependencies?

It was assumed that these source packages would be mostly used in platforms, I think it's fair to assume in modules consumed by Gu platforms that source packages are available and therefore peer.

## Quite a few links go to theguardian.com, is this acceptable in app?

Yup lets config/extract these!

## Or we abstract discussion-rendering so that we can use it with Bridget and the native layers handle authentication

Lets abstract into methods that can be passed through props

gtrufitt:
Note: I bumped discussion rendering (2.6.0) https://github.com/guardian/discussion-rendering/releases

webb04:
Not too sure why I'm seeing a misaligned icon 🤔

Screen Shot 2020-08-03 at 18 10 15

dskamiotis (approved):

PR #479, opened 2020-07-07
AR-67: Add caption to media atoms 

dskamiotis:
## Why are you doing this?
Captions not showing up in media atoms

## Changes

- Captions now showing up

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/atoms.ts
+++ b/src/atoms.ts
@@ -3,8 +3,9 @@ import { BlockElement } from "@guardian/content-api-models/v1/blockElement";
 import { Result, err, ok } from 'types/result';
 import { BodyElement, ElementKind } from "bodyElement";
 import { fromNullable } from "types/option";
+import {DocParser} from "./types/parserContext";
 
-function parseAtom(element: BlockElement, atoms: Atoms): Result<string, BodyElement> {
+function parseAtom(element: BlockElement, atoms: Atoms, docParser: DocParser): Result<string, BodyElement> {
     const id = element.contentAtomTypeData?.atomId
     switch (element.contentAtomTypeData?.atomType) {
         case "interactive": {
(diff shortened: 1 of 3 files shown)


webb04:
I think the ticket should have had some more description, but the grey square will eventually be covered by a video so we would need the caption to be just underneath like we have image captions.

webb04:
The same as what dotcom have https://www.theguardian.com/world/2020/jul/07/jair-bolsonaro-coronavirus-positive-test-brazil-president

src/atoms.ts
```diff
@@ -3,8 +3,9 @@ import { BlockElement } from "@guardian/content-api-models/v1/blockElement";
 import { Result, err, ok } from 'types/result';
 import { BodyElement, ElementKind } from "bodyElement";
 import { fromNullable } from "types/option";
+import {DocParser} from "./types/parserContext";
```
webb04, line 6:
```suggestion
import { DocParser } from "types/parserContext";
```

src/renderer.ts
```diff
@@ -596,7 +596,8 @@ const render = (format: Format, excludeStyles = false) =>
                 'data-duration': duration,
                 css: styles
             }
-            return styledH('div', attributes)
+            const figcaption = h(FigCaption, { format, caption: caption, credit: none });
+            return styledH('div', attributes, figcaption)
```
webb04, line 600: I think `figcaption` should be a sibling rather than a child. Happy to go through this over chat.

webb04: When the key/value is the same name in an object you can do 
```
{ format, caption, credit: none }
```

JamieB-gu: > I think `figcaption` should be a sibling rather than a child so that it appears underneath the video

Yeah. In fact I think we should wrap both in a `<figure>` element, as that's the only allowed parent of a [`<figcaption>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption):

```suggestion
            return h('figure', null, [ styledH('div', attributes), figcaption ]);
```

This is nice because it attaches more semantic meaning to the video for screen readers and the like. Note that it may require a couple of style tweaks to make it look right.

webb04:
@dskamiotis can pair on this Monday if it's useful?

PR #484, opened 2020-07-08
use labs colours in standfirst and logo links

webb04:
## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |
|  |  |

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -260,7 +260,7 @@ const textElement = (format: Format) => (node: Node, key: number): ReactNode => 
     }
 };
 
-const colourFromFormat = (format: Format): string => {
+const linkColourFromFormat = (format: Format): string => {
     if (format.design === Design.AdvertisementFeature) {
         return palette.labs[300];
     }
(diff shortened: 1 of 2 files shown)


src/renderer.ts
```diff
@@ -260,9 +260,17 @@ const textElement = (format: Format) => (node: Node, key: number): ReactNode =>
     }
 };
 
+const colourFromFormat = (format: Format): string => {
```
JamieB-gu, line 263:
```suggestion
const linkColourFromFormat = (format: Format): string => {
```

src/renderer.ts
```diff
@@ -271,7 +279,7 @@ const standfirstTextElement = (format: Format) => (node: Node, key: number): Rea
         case 'LI':
             return styledH('li', { css: listItemStyles(format) }, children);
         case 'A': {
-            const colour = format.design === Design.Media ? inverted : kicker;
+            const colour = colourFromFormat(format);
```
JamieB-gu, line 282:
```suggestion
            const colour = linkColourFromFormat(format);
```

PR #485, opened 2020-07-08
remove unused font resize link

webb04:
We've got an open PR to change this for Android but it's causing a csp issue for beta users so maybe we remove it until it's needed?


JamieB-gu:
cc @maxspencer

PR #489, opened 2020-07-09
add formstack urls

webb04:


src/server/csp.ts
```diff
@@ -58,20 +56,20 @@ const assetHashes = (assets: string[]): string =>
 //     child-src https: blob:
 // `.trim()
 
-const buildCsp = ({ styles, scripts }: Assets, twitter: boolean): string => `
+const buildCsp = ({ scripts }: Assets, twitter: boolean): string => `
     default-src 'self';
-    style-src ${assetHashes(styles)} https://interactive.guim.co.uk ${twitter ? 'https://platform.twitter.com' : ''};
+    style-src https: 'unsafe-inline';
```
JamieB-gu, line 61: This is [problematic for security](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src#Unsafe_inline_styles). Which styles has this been introduced to allow?

webb04: The formstack embed uses a few inline styles `display: none`

PR #490, opened 2020-07-09
handle latest tags

webb04:
The first link in the standfirst on this article links to a tag page with `/latest` at the end of the path. The current behaviour in apps rendering is to open up theguardian.com as it's not recognised as a valid tag.

Removing the `/latest` part of the url will link to a valid tag page and keep the users in app.

https://www.gu.com/stage/2020/jul/08/77-per-cent-comedy-venues-in-uk-face-closure-survey-emergency-arts-funding-coronvirus

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -33,6 +33,19 @@ import LiveEventLink from 'components/liveEventLink';
 const getAttrs = (node: Node): Option<NamedNodeMap> =>
     isElement(node) ? some(node.attributes) : none;
 
+
+const transformHref = (href: string): string => {
+    if (href.startsWith('profile/')) {
+        return `https://www.theguardian.com/${href}`;
+    }
+
+    if (/https:\/\/www\.theguardian\.com\/[a-zA-Z0-9|/|-]+\/latest/.exec(href)?.length) {
+        return href.slice(0, -7);
+    }
+
+    return href
+}
+
 const getHref = (node: Node): Option<string> =>
     pipe(
         getAttrs(node),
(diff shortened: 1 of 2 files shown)


src/renderer.ts
```diff
@@ -33,19 +33,29 @@ import LiveEventLink from 'components/liveEventLink';
...
     isElement(node) ? some(node.attributes) : none;
 
-const getAttr = (attr: string) => (node: Node): Option<string> =>
+
+const transformHref = (href: string): string => {
+    if (href.startsWith('profile/')) {
+        return `https://www.theguardian.com/${href}`;
+    }
+
+    if (/https:\/\/www\.theguardian\.com\/[a-zA-Z0-9|/|-]+\/latest/.exec(href)?.length) {
```
JamieB-gu, line 42: An alternative if you wanted to avoid the regex. Don't have a strong opinion either way, it's just regexes can be a bit brittle sometimes.

```ts
const url = new URL(href);
const path = url.pathname.split('/');
const isLatest = url.hostname === 'www.theguardian.com' && path[path.length - 1] === 'latest';
```

webb04: This is about the third time you've suggested `URL` but I continue to think of regex... I think that's more readable

JamieB-gu: Just a warning, the `URL` constructor can throw if it's not given a valid URL 🙄, so wrapping it in a `Result.fromUnsafe` might be helpful.

webb04: yeah it probably ends up with more code, but it might be easier to handle future URLs that need transforming.

PR #493, opened 2020-07-10
Babel Polyfill -> Corejs 3

JamieB-gu:
## Why are you doing this?

For a while we've had build warnings relating to `core-js` and Babel.

Babel [makes use of `core-js`](https://babeljs.io/docs/en/babel-preset-env#how-does-it-work) to polyfill missing features in your target runtime, based on your browserslist config. We're using the [`useBuiltIns: 'usage'`](https://babeljs.io/docs/en/babel-preset-env#usebuiltins-usage) functionality of `@babel/preset-env` to apply these polyfills only when the corresponding feature appears in our code.

I believe that the `core-js` polyfills were managed previously [via `@babel/polyfill`](https://babeljs.io/docs/en/babel-preset-env#usebuiltins-entry). However, `@babel/polyfill` was deprecated as of Babel 7.4.0 and the recommendation is now to [make `core-js` a direct dependency](https://babeljs.io/docs/en/babel-preset-env#usebuiltins) and specify the version in your Babel config. This recommendation change was responsible for the warning shown in our build logs.

To fix this I have removed `@babel/polyfill` and replaced it with a direct dependency on `core-js` 3.

FYI @gtrufitt @SiAdcock @sndrs 

## Changes

- Removed `@babel/polyfill`, deprecated as of Babel 7.4.0
- Added core-js 3 as dependency
- Specified core-js 3 in `babel-loader` config


sndrs:
the only issue i've had with v3 of corejs is some other dependencies having a peer dep on v2. if you don't get that then i'm pretty sure v3 is the way to go 👍

webb04:
Does this remove the need to import `regenerator-runtime`? https://github.com/guardian/apps-rendering/pull/472

gtrufitt:
> Does this remove the need to import `regenerator-runtime`? #472

See my reply, I hope you don't need it anyway :) https://github.com/guardian/apps-rendering/pull/472#issuecomment-656194199

JamieB-gu:
> Does this remove the need to import `regenerator-runtime`

No sadly not. That doesn't come under the scope of `core-js` AFAIK.

PR #494, opened 2020-07-10
render strong tags in standfirsts

webb04:
## Why are you doing this?
This ensures we stay inside the recursive `renderStandfirst`

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -274,13 +274,15 @@ const standfirstTextElement = (format: Format) => (node: Node, key: number): Rea
     switch (node.nodeName) {
         case 'P':
             return h('p', { key }, children);
+        case 'STRONG':
+            return h('strong', { key }, children);
         case 'UL':
             return styledH('ul', { css: listStyles }, children);
         case 'LI':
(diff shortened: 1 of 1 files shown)


JamieB-gu:
The only difference I can see between the two screenshots is that the underline has disappeared from the link, which I presume is the `;` applied after `text-decoration`? Is there something visual that handling `<strong>` tags here has changed?

webb04:
> The only difference I can see between the two screenshots is that the underline has disappeared from the link, which I presume is the `;` applied after `text-decoration`? Is there something visual that handling `<strong>` tags here has changed?

Yes by having a `strong` case it doesn't reach the `default` case which uses the standard body element renderer. So now it stays inside the `renderStandfirst` recursive loop and uses the standfirst `A` case

JamieB-gu:
> `strong` case it doesn't reach the `default`

Ah, you mean there are anchor tags *inside* a `<strong>` tag. I see.

JamieB-gu:
> `strong` case it doesn't reach the `default`

Ah, you mean there are anchor tags *inside* `<strong>` tags, I see.

PR #502, opened 2020-07-14
replace reserved keyword with mandatory

webb04:
## Why are you doing this?
When attempting to generate the Scala and TypeScript packages I get an error about using a reserved keyword "required".

It's a shame to lose the name "required" as it's commonly used to describe required fields in callouts throughout the targeting tool, dotcom and the callout React components that we hope to share with dotcom. For example: https://github.com/guardian/frontend/blob/master/common/app/model/dotcomrendering/pageElements/CalloutExtraction.scala

Would using French thrift allow this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,8 @@ In order to release the models you'll need to:
  - have a bintray account with access to the guardian organisation
  - have an NPM account, part of the [@guardian](https://www.npmjs.com/org/guardian) org with a [configured token](https://docs.npmjs.com/creating-and-viewing-authentication-tokens)
 
+Make sure you set upstream `git push --set-upstream origin <BRANCH_NAME>`
+
 In the SBT repl:
 ```sbtshell
 clean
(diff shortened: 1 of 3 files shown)


davidfurey:
French thrift would allow this.  Possibly also moving to Alex's new Scrooge Thrift generator might also allow this.

JamieB-gu:
Is this something that's blocking you and you'd like to merge now, or would you prefer to look into the solutions David mentioned above?

webb04:
This is using Alex's scrooge generation but it looks like it's using the official thrift repo rather than french-thrift. Do you know if we've published the french-thrift packages? https://github.com/guardian/apps-rendering/blob/master/build.sbt#L36

webb04:
@JamieB-gu This can be released how it is, I've released the new packages and merged into master https://github.com/guardian/mobile-apps-api/pull/1627

PR #503, opened 2020-07-14
debounce resize event listener for mutation observer

webb04:
## Why are you doing this?

There's a heavy `stringify` operation inside the `callback` function that should be debounced inside the event listener.


src/client/nativeCommunication.ts
```diff
@@ -62,7 +62,15 @@ function insertAds(): void {
...
 
-        window.addEventListener('resize', callback);
+
+        let timeout: number;
+        window.addEventListener('resize', () => {
+            if (timeout) {
+                window.cancelAnimationFrame(timeout);
+            }
+            timeout = window.requestAnimationFrame(callback);
+        }, false);
```
JamieB-gu, line 72: Would you be able to give a brief explanation of how this works?

webb04: https://gomakethings.com/debouncing-your-javascript-events/#requestanimationframe

JamieB-gu: Under what circumstances will the `timeout` in `if (timeout)` be falsy?

webb04: The first time when there is no previous animation frame to cancel. I've renamed it as it's not really a "timeout"

src/client/nativeCommunication.ts
```diff
@@ -183,7 +183,13 @@ function reportNativeElementPositionChanges(): void {
         }
     };
 
-    window.addEventListener('resize', callback);
+    let currentAnimationFrame: number;
+    window.addEventListener('resize', () => {
+        if (currentAnimationFrame) {
```
JamieB-gu, line 188:
```suggestion
    let currentAnimationFrame: number | null = null;
    window.addEventListener('resize', () => {
        if (currentAnimationFrame !== null) {
```

PR #511, opened 2020-07-16
Fix Lightbox Images

JamieB-gu:
## Why are you doing this?

Beta feedback has flagged that lightbox images are no longer working. This is a short-term fix for that problem.

This was a side-effect of the changes in #506, in which we started signing the query params of images, rather than just the path. Because we were altering the `width` parameter of lightbox images on the fly, depending on the device size, the URL passed to the lightbox no longer matched its signature.

**Note:** This may result in slightly lower resolutions for images in the lightbox. I'm going to do some additional work to remedy that in a later PR.

## Changes

- Stopped altering lightbox image sources

### DIFF
--- a/src/client/nativeCommunication.ts
+++ b/src/client/nativeCommunication.ts
@@ -97,18 +97,12 @@ function getImageWidth(src: string): number {
     return Math.max(screen.height * dpr, screen.width * dpr, width);
 }
 
-function updateUrl(src: string): string {
-    const url = new URL(src);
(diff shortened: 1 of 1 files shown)


webb04:
What was the reasons for removing the `sig-ignore` param?

JamieB-gu:
Just testing on CODE.

JamieB-gu:
> What were the reasons for removing the `sig-ignore` param?

Discussed offline.

PR #521, opened 2020-07-17
Refactor Page

JamieB-gu:
## Why are you doing this?

When walking through this code recently with @dskamiotis I realised that it's become a little tangled, and thus hard to understand for those coming to it fresh. This PR is an attempt to tidy it up a bit by splitting into smaller functions and components that do fewer things at once.

I'd like to do some further work to provide better error logging now that we're in Beta, which I hope will build on this.

## Changes

- Pulled the body rendering into a component
- Pulled some of the head rendering into a component
- Put the scripts rendering into a standalone component file
- Broke down into smaller functions

### DIFF
--- a/src/server/page.tsx
+++ b/src/server/page.tsx
@@ -100,7 +100,7 @@ function renderHead(
             ${JSON.stringify(request.targetingParams)}
         </script>
     `;
-};
+}
 
 const renderBody = (item: Item, request: RenderingRequest): EmotionCritical =>
     compose(extractCritical, renderToString)(
--- a/src/server/page.tsx
+++ b/src/server/page.tsx
@@ -1,32 +1,34 @@
 // ----- Imports ----- //
 
-import React, { ReactNode, FC, ReactElement } from 'react';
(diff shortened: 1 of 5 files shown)


src/server/page.tsx
```diff
@@ -164,53 +78,67 @@ const styles = (format: Format): string => `
...
-    imageSalt: string,
-    renderingRequest: RenderingRequest,
-    getAssetLocation: (assetName: string) => string,
-): Page {
-    const item = fromCapi({ docParser, salt: imageSalt })(renderingRequest);
-    const shouldHideAds = renderingRequest.content.fields?.shouldHideAdverts ?? false;
-    const hasTweets = includesTweets(renderingRequest.content);
-    const clientScript = map(getAssetLocation)(scriptName(item));
-    const { html: body, css, ids } = compose(extractCritical, renderToString)(
+function renderHead(
```
webb04, line 81: Maybe another one to abstract into a `head` file?

JamieB-gu: I actually did that at one point and then undid it again 🙈

PR #522, opened 2020-07-17
Use guardian/types For Option And Result

JamieB-gu:
## Why are you doing this?

`Option` and `Result` are now available in the [guardian/types](https://github.com/guardian/types) repo, so we no longer need to maintain them here. This PR updates the `guardian/types` dependency and migrates all imports across. It also deletes the now-unused types files in this repo.

This PR shifts to using a direct GitHub dependency on the types repo, instead of depending on the library in npm's registry. This is something we've discussed as an option for internal packages in the @guardian/client-side-infra group, so would be keen to hear what people think about this (cc @sndrs @SiAdcock). We can always revert to releasing a new version of `guardian/types` and updating this PR to use that instead, if needed.

**Note:** The `monad` and `functor` files have actually been unused since #404 and #430.

## Changes

- Started using `guardian/types` for `Option` and `Result`
- Migrated from npm dependency to GitHub dependency for `guardian/types`
- Deleted `functor`, `monad`, `option` and `result` (and tests) files


package.json
```diff
@@ -71,7 +71,7 @@
     "@guardian/src-button": "^1.8.0",
     "@guardian/src-foundations": "^1.9.0",
     "@guardian/src-svgs": "^1.8.0",
-    "@guardian/types": "^0.3.0",
```
webb04, line 74: Is this a new syntax for GitHub packages? Does it always fetch master or the latest tag release? might we want to pin to a specific version?

ripecosta: By default I believe it points to the latest version of master. You can further specific which commit/tag to use by using an `#`, which you can use for versioning. See more here: https://docs.npmjs.com/files/package.json#git-urls-as-dependencies

JamieB-gu: The version pinning is handled in [`package-lock.json`](https://github.com/guardian/apps-rendering/pull/522/files#diff-32607347f8126e6534ebc7ebaec4853dR4571); it specifies the repo and specific commit on `master` as the version.

webb04: Would an `npm i` overwrite the lock file to a later version?

JamieB-gu: I don't think so... if it did it would defeat the point of a lockfile wouldn't it?

webb04: hmm yeah, but how would the lockfile know what version to use? It might be more explicit in the package.json too

JamieB-gu: If you don't specify a version when you first add it to the project it picks the latest `master` commit and locks the package to that. Alternatively you can specify a specific commit when you first add it as @ripecosta mentioned above.

PR #529, opened 2020-07-20
Callout embeds

webb04:
Test article: https://www.theguardian.com/world/2020/jul/16/uk-government-orders-halt-randox-covid-19-tests-over-safety-issues

TODO:
- Fix CORS issues by enabling cross origin submissions from apps-rendering urls. I've spoken to the CAPI team about this to update the submission lambda.

Screen Shot 2020-07-22 at 14 42 43
Screen Shot 2020-07-22 at 14 44 31
Screen Shot 2020-07-22 at 14 42 56
Screen Shot 2020-07-22 at 14 43 07


webb04:
@paperboyo These screenshots were running apps-rendering through MAPI locally, so it didn't have the correct font assets

src/components/calloutForm.tsx
```diff
@@ -0,0 +1,225 @@
...
+        }
+    
+        .campaign--kicker {
+            display: flex;
+            flex-direction: row;
+    
+            > .campaign--snippet__heading-logo {
+                flex: initial;
+    
+                .speech-bubble {
```
JamieB-gu, line 62: With Emotion are we able to avoid these types of deeply nested selectors? Can we apply styles as needed directly to the components? For example, below in `CalloutForm`:

```tsx
const speechBubbleStyles =(kicker: string) => css`
  padding: ${remSpace[3]};
  position: relative;
  color: ${neutral[100]};
  background-color: ${kicker};
  margin-bottom: ${remSpace[9]};
  min-width: 5.5rem;

  &::after {
    content: '';
    width: 1.25rem;
    height: 1.375rem;
    border-radius: 0 0 1.125rem;
    position: absolute;
    bottom: -0.75rem;
    left: 0.625rem;
    color: ${neutral[100]};
    background-color: ${kicker};
  }
`;

const headingStyles = css`
  ${headline.xxxsmall({ fontWeight: 'bold' })};
  margin: 0;
`;

const Logo = () =>
  <div css={css`flex: initial;`}>
    <div css={speechBubbleStyles(kicker)}>
      <h4 css={headingStyles}>Take part</h4>
    </div>
  </div>
```

I think this approach will allow us to split the UI into smaller components. Also, applying styles directly like this without relying on tag and class selectors will make it easier to scope the styles to specific elements, in addition to altering the structure if needed.

src/components/calloutForm.tsx
```diff
@@ -0,0 +1,225 @@
...
+`;
+
+const labelStyles = css`
+    ${textSans.medium({ fontWeight: 'bold' })};
+`;
+
+const renderField = ({ type, label, mandatory, options, id }: FormField): JSX.Element => {
+    switch (type) {
+        case 'text': {
+            return <TextInput name={`field_${id}`} label={label} optional={!mandatory}/>
```
JamieB-gu, line 143: Could we store this field id in a `const` so we don't need to repeat the template string?

```tsx
const fieldId = `field_${id}`;
```

src/components/calloutForm.tsx
```diff
@@ -0,0 +1,225 @@
...
+    ${textSans.small({ fontStyle: "italic" })};
+    color: ${neutral[46]};
+`;
+
+const labelStyles = css`
+    ${textSans.medium({ fontWeight: 'bold' })};
+`;
+
+const renderField = ({ type, label, mandatory, options, id }: FormField): JSX.Element => {
+    switch (type) {
```
JamieB-gu, line 141: This is nice 🎉

src/components/calloutForm.tsx
```diff
@@ -0,0 +1,225 @@
...
+            return <>
+                <label css={labelStyles} htmlFor={`field_${id}`}>{label} {optionalLabel}</label>
+                <input
+                    name={`field_${id}`}
+                    type="file"
+                    accept="image/*, .pdf"
+                    required={mandatory}
+                />
+                <p>We accept images and pdfs. Maximum total file size: 6MB</p>
+            </>
```
JamieB-gu, line 159: Worth extracting this into a sub-component to reduce the length of the `renderField` function? Could also make its way upstream to Source at some point: `FileInput`.

src/components/calloutForm.tsx
```diff
@@ -0,0 +1,225 @@
...
+                                        key={value}
+                                        value={value}
+                                        label={label}
+                                    />
+                                )
+                            })
+                        }
+                    </RadioGroup>
+                </>
+            )
```
JamieB-gu, line 182: Similar comment as above, worth extracting into a sub-component?

src/components/calloutForm.tsx
```diff
@@ -0,0 +1,225 @@
...
+const optionalLabelStyles = css`
+    ${textSans.small({ fontStyle: "italic" })};
+    color: ${neutral[46]};
+`;
+
+const labelStyles = css`
+    ${textSans.medium({ fontWeight: 'bold' })};
+`;
+
+const renderField = ({ type, label, mandatory, options, id }: FormField): JSX.Element => {
```
JamieB-gu, line 140:
```suggestion
const renderField = ({ type, label, mandatory, options, id }: FormField): ReactElement | null => {
```

src/components/calloutForm.tsx
```diff
@@ -0,0 +1,225 @@
...
+                                )
+                            })
+                        }
+                    </RadioGroup>
+                </>
+            )
+        }
+
+        default:
+            return <></>;
```
JamieB-gu, line 186:
```suggestion
            return null;
```

src/components/calloutForm.tsx
```diff
@@ -0,0 +1,225 @@
...
+            <summary>
+                <div className="campaign--kicker">
+                    <div className="campaign--snippet__heading-logo">
+                        <div className="speech-bubble">
+                            <h4>Take part</h4>
+                        </div>
+                    </div>
+                    <div className="description">
+                        <h4>{campaign.fields.callout}</h4>
+                        <div dangerouslySetInnerHTML={{__html: campaign.fields.description ?? ''}}></div>
```
JamieB-gu, line 204: Can we render this with React components?

src/components/calloutForm.tsx
```diff
@@ -0,0 +1,225 @@
...
+                </div>
+                <Button size="xsmall" className="is-off callout-expand" iconSide="left" icon={<SvgPlus />}>Tell us</Button>
+                <Button size="xsmall" className="is-on callout-expand" iconSide="left" icon={<SvgMinus />}>Hide</Button>
+            </summary>
+
+
+            <form action="#" method="post">
+                <div>
+                    <input name="formId" type="hidden" value={campaign.id} />
+                    {campaign.fields.formFields.map(renderField)}
```
JamieB-gu, line 215: Nice!

src/item.ts
```diff
@@ -35,6 +36,7 @@ interface Fields extends Format {
     shouldHideReaderRevenue: boolean;
     branding: Option<Branding>;
     commentCount: Option<number>;
+    campaigns: Option<Campaign[]>;
```
JamieB-gu, line 39: Is this property accessed anywhere?

src/server/csp.ts
```diff
@@ -65,7 +65,7 @@ const buildCsp = ({ styles, scripts }: Assets, twitter: boolean): string => `
     script-src 'self' ${assetHashes(scripts)} http://www.instagram.com/embed.js https://interactive.guim.co.uk https://s16.tiktokcdn.com https://www.tiktok.com/embed.js ${twitter ? 'https://platform.twitter.com https://cdn.syndication.twimg.com' : ''};
     frame-src https://www.theguardian.com https://www.scribd.com https://www.instagram.com https://www.tiktok.com https://interactive.guim.co.uk https://open.spotify.com https://www.youtube-nocookie.com ${twitter ? 'https://platform.twitter.com https://syndication.twitter.com https://twitter.com' : ''};
     font-src 'self' https://interactive.guim.co.uk;
-    connect-src 'self' https://interactive.guim.co.uk https://sf-hs-sg.ibytedtos.com/ https://gdn-cdn.s3.amazonaws.com/
+    connect-src 'self' https://callouts.code.dev-guardianapis.com/formstack-campaign/submit https://interactive.guim.co.uk https://sf-hs-sg.ibytedtos.com/ https://gdn-cdn.s3.amazonaws.com/
```
JamieB-gu, line 68: Hmm, this is an interesting point. Perhaps we need to start passing the `Stage` to the CSP module to swap these URLs between CODE and PROD. An example of switching CSPs like this can be seen in the support-frontend confs: [CODE.public.conf](https://github.com/guardian/support-frontend/blob/f7ba7175bd75368a33ee61296ef87903670841a4/support-frontend/conf/CODE.public.conf#L20) and [PROD.public.conf](https://github.com/guardian/support-frontend/blob/f7ba7175bd75368a33ee61296ef87903670841a4/support-frontend/conf/PROD.public.conf#L19).

webb04: That could be a good idea, in this case (confusingly) both PROD and CODE submit to that url. It's the same case on dotcom @JustinPinner

JamieB-gu: That's... odd. Why would PROD submit to a CODE url?

src/bodyElement.ts
```diff
@@ -143,6 +152,8 @@ function toSerialisable(elem: BodyElement): JsonSerialisable {
             return { ...elem, content: serialiseNodes(elem.content) };
         case ElementKind.MediaAtom:
             return { ...elem, caption: map(serialiseFragment)(elem.caption) };
+        case ElementKind.Callout:
+            return { ...elem, campaign: serialiseCampaign(elem.campaign) };
```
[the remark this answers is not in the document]
webb04: I don't think it is, but I don't think Callouts are used in lilveblogs so it worth adding? Would it just be stringifying the Campaign object?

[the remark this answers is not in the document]
JamieB-gu: Everything [here](https://github.com/guardian/apps-rendering/blob/30290357ca9b93f6bf49e342a16c19c474dca2ba/api-models/thrift/src/main/thrift/appsRendering.thrift#L40) is a `string`, `bool`, `number`, `Array` or `Object` isn't it? If so, won't JSON.stringify automatically be able to convert it?

Essentially what I'm asking is: if you added `case ElementKind.Callout:` to the line after `case ElementKind.Embed:`, would it "just work"?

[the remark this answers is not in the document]
webb04: That doesn't work, there's a few nested types that would need serialising https://github.com/guardian/apps-rendering/blob/30290357ca9b93f6bf49e342a16c19c474dca2ba/api-models/thrift/src/main/thrift/appsRendering.thrift#L47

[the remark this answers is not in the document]
JamieB-gu: Ok, think this is because `Int64` gets represented as a `class` once it's been converted to TypeScript. One solution to this would be to add this case to the `JsonSerialisable` type:

```ts
| { toString(): string }
```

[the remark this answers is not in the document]
JamieB-gu: Following a discussion offline, this is probably fine as-is, although if Callouts start to appear in liveblogs we may have to re-visit.

src/client/article.ts
```diff
@@ -110,6 +110,92 @@ function insertEpic(): void {
...
+        });
+
+        reader.readAsDataURL(file);
+    })
+}
+
+function callouts(): void {
+    const callouts = Array.from(document.querySelectorAll('details.callout'));
+    callouts.forEach(callout => {
+        const buttons = Array.from(callout.querySelectorAll('.callout-expand'));
```
JamieB-gu, line 159:
```suggestion
        const buttons = Array.from(callout.querySelectorAll('.js-callout-expand'));
```

src/client/article.ts
```diff
@@ -110,6 +110,92 @@ function insertEpic(): void {
...
+            })
+        })
+
+        const form = callout.querySelector('form');
+        if (!form) return;
+        // eslint-disable-next-line @typescript-eslint/no-misused-promises
+        form.addEventListener('submit', async (e): Promise<void> => {
+            try {
+                e.preventDefault();
+                const elements = form.elements as HTMLCollectionOf<HTMLInputElement>;
```
JamieB-gu, line 172: Can you avoid the cast like this?

```suggestion
                const elements = form.getElementsByTagName('input');
```

src/client/article.ts
```diff
@@ -110,6 +110,92 @@ function insertEpic(): void {
...
+                        }
+                    } else if (type === 'file' && files?.length) {
+                        acc[name] = await readFile(files[0]);
+                    } else if (value) {
+                        acc[name] = value;
+                    }
+                    return Promise.resolve(acc);
+                }, Promise.resolve({}));
+
+                submit(await data, form);
```
JamieB-gu, line 188: I think this can be simplified slightly (and immutability retained up to a point):

```suggestion
                const data = Array.from(elements).reduce<Promise<FormData>>(async (formData, elem) => {
                    const { type, checked, name, value, files } = elem;

                    if (type === 'radio' && checked) {
                        return { ...(await formData), [name]: value };
                    } else if (type === 'file' && files?.length) {
                        return { ...(await formData), [name]: await readFile(files[0]) };
                    } else if (value) {
                        return { ...(await formData), [name]: value };
                    }
                    
                    return formData;
                }, Promise.resolve({}));
```

**Note** I haven't tested this, but it does type-check 🙂

src/components/FileInput.tsx
```diff
@@ -0,0 +1,36 @@
...
+    ${textSans.small({ fontStyle: "italic" })};
+    color: ${neutral[46]};
+`;
+
+const labelStyles = css`
+    ${textSans.medium({ fontWeight: 'bold' })};
+`;
+
+const optionalLabel = (required: boolean): ReactElement =>
+    required ? <></> : <span css={optionalLabelStyles}>Optional</span>
```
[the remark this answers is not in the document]
webb04: Is there a performance reason for this? or is it for readability?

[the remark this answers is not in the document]
JamieB-gu: The [recommendation from the docs](https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering) to prevent a something from rendering is to return `null`. One possible difference is that `Fragment` may create an object in the tree that React builds up - so there might be a performance difference.

src/components/FileInput.tsx
```diff
@@ -0,0 +1,37 @@
...
+const labelStyles = css`
+    ${textSans.medium({ fontWeight: 'bold' })};
+`;
+
+const optionalLabel = (required: boolean): ReactElement | null =>
+    required ? null : <span css={optionalLabelStyles}>Optional</span>
+
+const FileInput = (props: FileInputProps): ReactElement =>
+    <>
+        <label css={labelStyles} htmlFor={props.name}>{props.label} {optionalLabel}</label>
```
JamieB-gu, line 26: Isn't `optionalLabel` a function? Do you need to call it with a value for `required`?

src/components/calloutForm.tsx
```diff
@@ -0,0 +1,186 @@
...
+    const fieldId = `field_${id}`;
+    switch (type) {
+        case 'text':
+            return <TextInput name={fieldId} label={label} optional={!mandatory} />
+        case 'textarea':
+            return <TextArea name={fieldId} label={label} optional={!mandatory} />
+        case 'file':
+            return <FileInput required={mandatory} name={fieldId} label={label} />
+        case 'radio':
+            return <RadioInput options={options} name={fieldId } label={label} />
```
JamieB-gu, line 144:
```suggestion
            return <RadioInput options={options} name={fieldId} label={label} />
```

src/components/calloutForm.tsx
```diff
@@ -0,0 +1,186 @@
...
+        default:
+            return null;
+    }
+}
+
+const CalloutForm: FC<CalloutProps> = (props: CalloutProps): ReactElement => {
```
JamieB-gu, line 150: Just an FYI, I *think* `FC` provides a return type of `ReactElement | null` for you, so you could choose not to include the `: ReactElement` here.

src/fixtures/item.ts
```diff
@@ -31,6 +31,7 @@ const fields = {
     shouldHideReaderRevenue: false,
     branding: none,
     commentCount: none,
+    campaigns: none,
```
JamieB-gu, line 34: I think this is unused?

```suggestion
```

src/client/article.ts
```diff
@@ -110,6 +110,96 @@ function insertEpic(): void {
     }
 }
 
+interface FormData {
+    [key: string]: string;
+}
+
+function submit(body: FormData, form: Element): void {
+    fetch('https://callouts.code.dev-guardianapis.com/formstack-campaign/submit', {
```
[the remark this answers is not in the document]
webb04: Messaged on chat.

PR #547, opened 2020-07-28
Related content cards

webb04:
## Why are you doing this?
This will allow us to server side render related content in apps-rendering. If we want to make it async, we probably need a new MAPI endpoint as there's a lot of redundant data in the current `related-items` endpoint.

The link could be a `rendered-item` url or an old `relatedUri` in MAPI for old templates - would this be confusing @faresite? maybe we need two separate optional fields?

We might need to pass through `pillar` and `feature` if we want to match current features. We could also update some of the card designs e.g. related content on dotcom uses opinion styles for opinion cards. The example below would also require author data being passed through.

Screen Shot 2020-07-28 at 16 18 16


api-models/thrift/src/main/thrift/appsRendering.thrift
```diff
@@ -54,4 +67,5 @@ struct RenderingRequest {
     4: optional map<string,string> targetingParams
     5: optional Branding branding
     6: optional list<Campaign> campaigns
+    7: optional list<RelatedContent> relatedContent
```
JamieB-gu, line 70: Would this mean that there are two situations in which we wouldn't want to render the related content component? One when this field is missing/`undefined`, and one when the list is empty?

api-models/thrift/src/main/thrift/appsRendering.thrift
```diff
@@ -3,6 +3,19 @@ include "content/v1.thrift"
...
 
+struct Image {
+    1: required string url
+    2: required i32 height
+    3: required i32 width
+}
+
+struct RelatedContent {
+    1: required string title
+    2: required string lastModified
```
JamieB-gu, line 14: Do we have a date type/format available for this by any chance?

api-models/thrift/src/main/thrift/appsRendering.thrift
```diff
@@ -3,6 +3,19 @@ include "content/v1.thrift"
 namespace scala com.gu.mobile.ar.models
 namespace typescript _at_guardian.apps_rendering_api_models
 
+struct Image {
+    1: required string url
```
JamieB-gu, line 7: What type of url is provided for us here? Will it be a master asset that we can derive a `srcset` from?

JamieB-gu: Found it, the answer is "yes", it is a master asset.

api-models/thrift/src/main/thrift/appsRendering.thrift
```diff
@@ -3,6 +3,19 @@ include "content/v1.thrift"
...
+    2: required i32 height
+    3: required i32 width
```
JamieB-gu, line 9: What do we use these for in Apps-Rendering?

webb04: Nothing at the moment, I thought they might be useful to know the aspect ratio but I'm not sure if they are required in the image resizing code?

JamieB-gu: > I thought they might be useful to know the aspect ratio

Sounds sensible 👍

src/components/shared/relatedContent.tsx
```diff
@@ -0,0 +1,62 @@
...
+    }
+`
+
+const RelatedContent = ({ content }: Props): JSX.Element => {
+    return pipe2(
+        content,
+        map(({ title, relatedItems, resizedImages }) => {
+            return <section css={styles}>
+                <h1>{title}</h1>
+                <ul>
```
JamieB-gu, line 48: Similar comment about applying styles directly:

```suggestion
                <h1 css={headingStyles}>{title}</h1>
                <ul css={listStyles}>
```

src/components/shared/relatedContent.tsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import { RelatedContent } from '@guardian/apps-rendering-api-models/relatedContent';
```
JamieB-gu, line 2: Is this used?

src/components/standard/article.tsx
```diff
@@ -103,10 +104,13 @@ const Standard = ({ item, children }: Props): JSX.Element => {
...
             </Body>
             {epicContainer}
-            <footer css={articleWidthStyles}>
+            <section id="tags" css={articleWidthStyles}>
                 <Tags tags={item.tags} format={item}/>
-            </footer>
+            </section>
         </article>
+        <footer css={[articleWidthStyles, footerWidthStyles]}>
+            <RelatedContent css={articleWidthStyles} content={item.relatedContent}/>
```
JamieB-gu, line 112: I'm not sure this `css` prop is obeyed for non-built-in components 🤔. I can't see the styles being applied in the DOM.

```suggestion
            <RelatedContent content={item.relatedContent}/>
```

src/components/standard/article.tsx
```diff
@@ -103,10 +104,13 @@ const Standard = ({ item, children }: Props): JSX.Element => {
                 {children}
             </Body>
             {epicContainer}
-            <footer css={articleWidthStyles}>
+            <section id="tags" css={articleWidthStyles}>
                 <Tags tags={item.tags} format={item}/>
-            </footer>
+            </section>
         </article>
+        <footer css={[articleWidthStyles, footerWidthStyles]}>
```
JamieB-gu, line 111: I think we may want to avoid using `articleWidthStyles` here; I think we probably want this to be full-width between `phablet` and `wide` instead?

tablet

src/components/standard/article.tsx
```diff
@@ -103,10 +104,13 @@ const Standard = ({ item, children }: Props): JSX.Element => {
                 {children}
             </Body>
             {epicContainer}
-            <footer css={articleWidthStyles}>
+            <section id="tags" css={articleWidthStyles}>
```
JamieB-gu, line 107: What does this `id` get used for?

src/image.ts
```diff
@@ -159,6 +161,24 @@ const parseImage = ({ docParser, salt }: Context) =>
...
+        ...srcsets(image.url, salt),
+        alt: none,
+        width: image.width,
+        height: image.height,
+        caption: none,
+        credit: none,
+        nativeCaption: none,
+        role: some(Role.Card),
+    });
+}
```
JamieB-gu, line 180: This is great!

src/item.ts
```diff
@@ -149,7 +156,12 @@ const itemFields = (context: Context, request: RenderingRequest): ItemFields =>
         tags: content.tags,
         shouldHideReaderRevenue: content.fields?.shouldHideReaderRevenue ?? false,
         branding: fromNullable(branding),
-        commentCount: fromNullable(commentCount)
+        commentCount: fromNullable(commentCount),
+        relatedContent: pipe2(relatedContent, fromNullable, map(relatedContent => ({
```
JamieB-gu, line 160: Nice!

api-models/thrift/src/main/thrift/appsRendering.thrift
```diff
@@ -3,6 +3,42 @@ include "content/v1.thrift"
...
+    SPECIAL,
+    AUDIO,
+    LIVE,
+    VIDEO,
+    REVIEW,
+    ADVERTISEMENT_FEATURE,
+    GUARDIAN_VIEW
+}
+
+struct Image {
```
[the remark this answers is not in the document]
webb04: We do have an optional altText so I'll pass that through too

api-models/thrift/src/main/thrift/appsRendering.thrift
```diff
@@ -3,6 +3,42 @@ include "content/v1.thrift"
...
+}
+
+struct RelatedItem {
+    1: required string title
+    2: optional v1.CapiDateTime lastModified
+    3: optional Image headerImage
+    4: required string link
+    5: required RelatedItemType type
+    6: required v1.Pillar pillar
+    7: optional string length
```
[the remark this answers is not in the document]
webb04: Audio and Video articles show how long they last on the card. This could be made more descriptive like `mediaLength` or `duration`?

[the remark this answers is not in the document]
JamieB-gu: > This could be made more descriptive like `mediaLength` or `duration`?

👍 `mediaLength` or `mediaDuration` sound good.

src/components/shared/card.tsx
```diff
@@ -0,0 +1,106 @@
...
+        position: absolute;
+        top: 0;
+        left: 0;
+    }
+
+    h2 {
+        ${headline.xxxsmall()};
+        margin: 0 0 0;
+        padding: ${remSpace[2]};
+        min-height: 150px;
```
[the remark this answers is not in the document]
webb04: This is currently needed to keep cards the same height, I think this can be achieved with grid now..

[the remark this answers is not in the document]
JamieB-gu: Really? It looks like this for me with that property turned off:

card-height

src/components/standard/article.tsx
webb04: It's used to position epics above I think, so I'll change this to js-tags classname

src/components/standard/article.tsx
webb04: Takes up more space now
Screen Shot 2020-08-07 at 11 35 15

src/components/shared/card.tsx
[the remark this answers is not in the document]
webb04: Does a longer headline change it? environment/2020/aug/07/covid-19-lockdown-will-have-negligible-impact-on-climate-crisis-study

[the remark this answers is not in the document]
JamieB-gu: This is what I'm getting for the cards on that piece:

environment-related

src/components/standard/article.tsx
JamieB-gu: Is that locally? I'm not seeing that on CODE.

webb04: Yes that's been updated, let me push those changes

src/components/shared/card.tsx
```diff
@@ -0,0 +1,166 @@
...
+    position: relative;
+`;
+
+const relativeFirstPublished = (date: Option<Date>): JSX.Element | null => pipe2(
+    date,
+    map(date => <time>{makeRelativeDate(date)}</time>),
+    withDefault<JSX.Element | null>(null),
+);
+
+const cardStyles = (type: RelatedItemType, pillar: contentPillar): SerializedStyles => {
```
JamieB-gu, line 78: Doesn't seem to have caused any problems here, but is there a concern over using a TS reserved word (`type`) for a variable name?

src/components/shared/card.tsx
```diff
@@ -0,0 +1,166 @@
...
+import { Option, withDefault, map, fromNullable } from '@guardian/types/option';
+import { makeRelativeDate } from 'date';
+import { pipe2 } from 'lib';
+import { text, neutral } from '@guardian/src-foundations/palette';
+import BodyImage from 'components/bodyImage';
+import { Pillar, Design, Display } from '@guardian/types/Format';
+import { Image } from 'image';
+import { darkModeCss } from 'styles';
+import { RelatedItemType } from '@guardian/apps-rendering-api-models/relatedItemType';
+import { Pillar as contentPillar } from '@guardian/content-api-models/v1/pillar';
```
JamieB-gu, line 15:
```suggestion
import { Pillar as ContentPillar } from '@guardian/content-api-models/v1/pillar';
```

src/components/shared/card.tsx
JamieB-gu, line 78:
```suggestion
const cardStyles = (type: RelatedItemType, pillar: ContentPillar): SerializedStyles => {
```

src/components/shared/card.tsx
```diff
@@ -0,0 +1,166 @@
...
+            return css``;
+        }
+    }
+
+    return css``;
+}
+
+const Card = ({ relatedItem, image }: Props): JSX.Element => {
+
+    const format = {
```
JamieB-gu, line 137: Should this be a `prop` passed to `Card`?

src/components/shared/card.tsx
```diff
@@ -0,0 +1,166 @@
...
+        pillar: Pillar.News,
+        design: Design.Article,
+        display: Display.Standard
+    }
+
+    const img = pipe2(
+        image,
+        map(img => {
+            return <BodyImage image={img} format={format}/>
```
JamieB-gu, line 146: Would using `Img` directly cause problems?

src/components/shared/card.tsx
```diff
@@ -0,0 +1,166 @@
...
+        }),
+        withDefault<ReactElement | null>(null)
+    )
+
+    const lastModified = relatedItem.lastModified?.iso8601;
+    const date = lastModified ? relativeFirstPublished(fromNullable(new Date(lastModified))) : null;
+
+    return <li css={[listStyles, cardStyles(relatedItem.type, relatedItem.pillar)]}>
+            <a css={anchorStyles}href={relatedItem.link}>
```
JamieB-gu, line 155:
```suggestion
            <a css={anchorStyles} href={relatedItem.link}>
```

src/components/shared/relatedContent.tsx
```diff
@@ -0,0 +1,65 @@
...
+        display: none;
+    }
+`;
+
+const RelatedContent = ({ content }: Props): JSX.Element | null => {
+    return pipe2(
+        content,
+        map(({ title, relatedItems, resizedImages }) => {
+            return <section css={styles}>
+                <h1 css={headingStyles}>{title}</h1>
```
JamieB-gu, line 46: Should this be an `h2` and the card titles `h3`s?

src/components/shared/card.tsx
webb04: Maybe this can be done in a following PR, when we generate a format from the data we get from MAPI

src/components/shared/card.tsx
```diff
@@ -0,0 +1,172 @@
...
+    const sizes = `(min-width: ${breakpoints.phablet}px) 620px, 100%`;
+
+    const img = pipe2(
+        image,
+        map(img => {
+            return <Img
+                image={img}
+                sizes={sizes}
+                format={format}
+            />
```
JamieB-gu, line 152: Nice! Does this mean we no longer need the changes to `BodyImage`?

PR #548, opened 2020-07-28
Stop running Chromatic on unnecessary PRs

webb04:
## Why are you doing this?

We're currently running Chromatic on all branches, regardless of what has changed. This should filter out branches that do not directly impact components or version updates e.g. from Source.

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

cc @gtrufitt @SiAdcock

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -2,7 +2,12 @@ name: "Chromatic"
 on:
     push:
         paths:
-        - 'src/components'
+        - '!api-models/'
+        - '!project/'
+        - '!*.sbt'
+        - '!*.md'
+        - '!*.yaml'
+        - '!.gitignore'
 
 jobs:
     test:
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,5 +1,8 @@
 name: "Chromatic"
-on: push
+on:
+    push:
+        paths:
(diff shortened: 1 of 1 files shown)


gtrufitt:
Nice, will look at DCR for this too, *however* I do like having that confidence that a dependency bump hasn't messed something up in the UI somehow (for example, if src-foundations makes a change)

JamieB-gu:
This seems sensible, but I share the concern that Gareth raised.

SiAdcock:
Seriously, I'd reconsider this. While Source's Chromatic run may catch obvious visual bugs, it can't predict how the components and foundations will be used by the wider department. Also Source's expectations of how a component ought to look may not meet apps-rendering's expectations.

webb04:
Good point @SiAdcock, I've made it more specific to not run on PRs like https://github.com/guardian/apps-rendering/pull/547

SiAdcock:
Okay thanks @webb04. I'm softening my stance on this slightly as Source upgrades are likely to be the edge case here. The vast majority of dependabot PRs are likely to be build tooling or things unrelated to the UI. As long as you have a visual regression system in place for checking, for instance, React, Emotion and Source upgrades before merging, maybe this will be fine?

I'm definitely stealing this idea for the Source repo! 👍

PR #577, opened 2020-08-05
LIVE-343-Prevent Image Src Generation Crashing On Invalid URL

frankie297:
## Why are you doing this?
At the moment the generation of image srcs uses the URL constructor.

This will crash if `input` is an invalid URL.

This PR fixes the crash issue. This has been tested locally and it no longer crashes with an invalid input string, it causes a broken image but still loads the article. 

paired with @dskamiotis


frankie297:
Would screenshots be helpful?

src/image.ts
```diff
@@ -72,19 +73,24 @@ const sign = (salt: string, path: string): string =>
...
+                fit: 'bounds',
+            });
 
-    return `${imageResizer}/${service}${path}&s=${sig}`;
+            const path = `${url.pathname}?${params.toString()}`;
+            const sig = sign(salt, path);
+            return `${imageResizer}/${service}${path}&s=${sig}`;
+    }),
+        withDefault(input)
+    )
```
JamieB-gu, line 93: Perhaps it doesn't make too much difference in this case, but worth noting that you don't need to go via `Option` to achieve this. `Result` also has `map` and a way of extracting the value via `either`.

Using `pipe2`:

```ts
return pipe2(
  maybeUrl,
  resultMap(url => {
    const service = getSubdomain(url.hostname);

    const params = new URLSearchParams({
      width: width.toString(),
      quality: dpr === Dpr.Two ? lowerQuality.toString() : defaultQuality.toString(),
      fit: 'bounds',
    });

    const path = `${url.pathname}?${params.toString()}`;
    const sig = sign(salt, path);

    return `${imageResizer}/${service}${path}&s=${sig}`;
  }),
  either(_ => input, url => url),
);
```

Alternatively, using `switch` (we should expose `ResultKind` as we did for `OptionKind`):

```ts
switch (maybeUrl.kind) {
  case ResultKind.Ok:
    const url = maybeUrl.value;
    const service = getSubdomain(url.hostname);

    const params = new URLSearchParams({
      width: width.toString(),
      quality: dpr === Dpr.Two ? lowerQuality.toString() : defaultQuality.toString(),
      fit: 'bounds',
    });

    const path = `${url.pathname}?${params.toString()}`;
    const sig = sign(salt, path);

    return `${imageResizer}/${service}${path}&s=${sig}`;

  case ResultKind.Err:
  default:
    return input;
);
```

src/image.ts
```diff
@@ -72,19 +73,24 @@ const sign = (salt: string, path: string): string =>
...
-        quality: dpr === Dpr.Two ? lowerQuality.toString() : defaultQuality.toString(),
-        fit: 'bounds',
-    });
-
-    const path = `${url.pathname}?${params.toString()}`;
-    const sig = sign(salt, path);
+    const url: Result<string, URL> = fromUnsafe(() => new URL(input), 'invalid url')
+    return pipe2(
+        toOption(url),
+        map(url => {
```
JamieB-gu, line 79: We should probably use a different name for the parameter here to avoid shadowing the `url` in the enclosing scope.

src/image.ts
frankie297: I think both suggestions look sound to me, as it was a pairing task with @dskamiotis I might ask him what makes more sense for his understanding as to what direction we take.

src/image.ts
frankie297: 👍

PR #582, opened 2020-08-06
LIVE-398-Image Height Using Percentage

dskamiotis:
## Why are you doing this?

Ensuring image height is rendered as a percentage.
Padding percentages are always based on the width of the containing block, but height is not.


dskamiotis:
Please feel free to check with body images, i believe i did check what i thought to be body images.

webb04:
I think for non-phablet:

height: calc((100vw - ${remSpace[4]}) * ${height / width});

and then I think we can remove `object-fit: cover`

JamieB-gu:
I'm probably missing something here, but I've been looking at the diff for a little while now and I'm not sure what's changed 🤔? The number is calculated using JS maths instead of a `calc`... but it's still the same number isn't it?

webb04:
> I'm probably missing something here, but I've been looking at the diff for a little while now and I'm not sure what's changed 🤔? The number is calculated using JS maths instead of a `calc`... but it's still the same number isn't it?

I think it might be a little different but I'm not sure we want to set height to be a value of vw

dskamiotis:
Yes I replaced the calc() as I thought (research) the calculation can be done server side? Thus saving a bit of time?
Also as the calc is usually used when calculating 2 different types of units it wasn't needed, the width and height are taken directly from the image data.

src/components/bodyImage.tsx
```diff
@@ -22,13 +22,12 @@ const styles = css`
...
 const imgStyles = (width: number, height: number): SerializedStyles => css`
-    height: calc(100vw * ${height / width});
+    height: ${100 * height / width}%;
     display: block;
     width: 100%;
-    object-fit: cover;
 
     ${from.phablet} {
-        height: calc(620px * ${height / width});
+        height: ${620 * height / width}%;
```
JamieB-gu, line 30:
```suggestion
        height: ${620 * height / width}px;
```

src/components/bodyImage.tsx
```diff
@@ -22,13 +22,12 @@ const styles = css`
 `;
 
 const imgStyles = (width: number, height: number): SerializedStyles => css`
-    height: calc(100vw * ${height / width});
+    height: ${100 * height / width}%;
```
JamieB-gu, line 25: This percentage will be relative to the parent if said parent has a `height` set. Because here the parent _doesn't_ have a height set explicitly, this actually looks ok because I believe it uses the aspect ratio of the image itself to set the height.

Since this is the case you might wonder _why_ we set the `height` at all, since it's set automatically for us by the image. We do it because we're worried about what happens _before_ the image loads. Images typically load in later than the text of the articles, particularly on slow connections. This causes two problems:

1. A caption that doesn't refer to anything visible floating in the middle of a block of text

2. Reflow (the article jumping up and down) as the images load in

We solve this by setting the height of the image tag explicitly, so that it retains a placeholder slot in the page while we wait for the image to load in. That way it makes it more obvious that a caption refers to an image that just hasn't loaded yet, _and_ the images loading in don't cause the page to jump about.

| No Height | Height Set |
| - | - |
| caption | placeholder |

In summary, percentages are problematic (something we also need to address in `image-rendering`). I think you may have to do something like this here:

```suggestion
    height: calc((100vw - ${remSpace[4]}) * ${height / width});
```

It's a bit brittle. I think we may need something more robust longer term - again something for the `image-rendering` project to solve.

dskamiotis: thanks for that explanation Jamie, makes it much clearer, will update 👍

JamieB-gu (approved):

PR #606, opened 2020-08-14
Live-418: audio and video cards

dskamiotis:
## Why are you doing this?

Appropriate formatting needed for audio and video cards need in related content, as well as svg and duration of audio/video for cards not displaying.

## Changes

- Added audio, video cards, the duration and also fixed the appearance on media pages (gallery, video, audio)

video-card

### DIFF
--- a/src/components/shared/card.tsx
+++ b/src/components/shared/card.tsx
@@ -15,7 +15,7 @@ import { getPillarStyles, pillarFromString } from 'pillarStyles';
 import Img from 'components/img';
 import { border } from 'editorialPalette';
 import { SvgCamera } from '@guardian/src-icons';
-
+import { SvgAudio } from '@guardian/src-icons';
 
 
 interface Props {
@@ -80,7 +80,7 @@ const relativeFirstPublished = (date: Option<Date>): JSX.Element | null => pipe2
 );
 
 const cardStyles = (itemType: RelatedItemType, format: Format): SerializedStyles => {
-    switch(itemType) {
+    switch (itemType) {
         case RelatedItemType.FEATURE: {
             const { kicker } = getPillarStyles(format.pillar);
 
@@ -121,7 +121,12 @@ const cardStyles = (itemType: RelatedItemType, format: Format): SerializedStyles
         }
 
         case RelatedItemType.AUDIO: {
-            return css``;
+            return css`
+                background: ${background.inverse};
+                h3 {
+                    color: ${text.ctaPrimary};
+                }
+            `;
         }
 
         case RelatedItemType.LIVE: {
@@ -193,10 +198,14 @@ const Card = ({ relatedItem, image }: Props): JSX.Element => {
     `;
 
     const icon = (itemType: RelatedItemType, format: Format): JSX.Element => {
-        if (itemType === RelatedItemType.GALLERY){
+        if (itemType === RelatedItemType.GALLERY) {
             return <section css={parentIconStyles}>
-                    <span css={iconStyles}>< SvgCamera /></span>
-                   </section>;
+                <span css={iconStyles}>< SvgCamera /></span>
+            </section>;
+        } else if (itemType === RelatedItemType.AUDIO) {
+            return <section css={parentIconStyles}>
+                <span css={iconStyles}>< SvgAudio /></span>
+            </section>;
         } else {
             return <section css={parentIconStyles} ></section>;
         }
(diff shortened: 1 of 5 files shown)


src/components/shared/card.tsx
```diff
@@ -150,10 +152,67 @@ const cardStyles = (itemType: RelatedItemType, format: Format): SerializedStyles
...
+        margin-top: 0.300rem;
+        display: block;
+    }
+`;
+
+const iconStyles = (format: Format): SerializedStyles => {
+    const { inverted } = getPillarStyles(format.pillar);
+    return css`
+        width: 1.5rem;
+        height: 1.4375rem;
```
webb04, line 171: We could simplify this to 1.5rem?

src/components/shared/card.tsx
```diff
@@ -150,10 +152,67 @@ const cardStyles = (itemType: RelatedItemType, format: Format): SerializedStyles
...
+            <span css={iconStyles(format)}>< SvgVideo /></span>
+        </section>;
+    } else {
+        return <section css={parentIconStyles} ></section>;
+    }
+}
+
+const metadataStyles: SerializedStyles = css`
+    padding: 0 ${remSpace[2]};
+    min-height: 1.4375rem
```
webb04, line 198: This could be 1.5rem?

src/components/shared/card.tsx
```diff
@@ -150,10 +152,67 @@ const cardStyles = (itemType: RelatedItemType, format: Format): SerializedStyles
     }
 }
 
-const Card = ({ relatedItem, image }: Props): JSX.Element => {
+const parentIconStyles: SerializedStyles = css`
+    display:inline-block;
+    svg {
+        width: 0.875rem;
```
webb04, line 158: We could make the width 1rem to be a little larger?

src/components/shared/card.tsx
```diff
@@ -150,10 +152,67 @@ const cardStyles = (itemType: RelatedItemType, format: Format): SerializedStyles
...
+        height: auto;
+        margin-left: auto;
+        margin-right: auto;
+        margin-top: 0.300rem;
```
webb04, line 162: If we make the svg 1rem and the height of the parent 1.5rem, I think a margin-top of 0.25rem will centre it nicely.

src/components/shared/card.tsx
```diff
@@ -150,10 +152,67 @@ const cardStyles = (itemType: RelatedItemType, format: Format): SerializedStyles
...
+        width: 1.5rem;
+        height: 1.4375rem;
+        display: inline-block;
+        background-color: ${inverted};
+        border-radius: 50%;
+    `;
+}
 
+const icon = (itemType: RelatedItemType, format: Format): JSX.Element => {
+    if (itemType === RelatedItemType.GALLERY) {
```
frankie297, line 179: Could this be a switch statement rather than an if/else? To make it easier to read

src/components/shared/card.tsx
```diff
@@ -172,58 +231,24 @@ const Card = ({ relatedItem, image }: Props): JSX.Element => {
...
-                </section>
-            </a>
-        </li>
+        <a css={anchorStyles} href={relatedItem.link}>
+            <section css={headingWrapperStyles}>
+                <h3 css={headingStyles}>{relatedItem.title}</h3>
+            </section>
+            <section>
+                <div css={metadataStyles}>
+                    {icon(relatedItem.type, format)}
```
JamieB-gu, line 244: Since the icon parent doesn't appear to change, would it make sense to do:

```suggestion
                    <section css={parentIconStyles}>
                        {icon(relatedItem.type, format)}
                    </section>
```

and just have `icon` return the `span` (`null` in the default case)?

On a related note, what's the thinking behind this being a `section`? Not suggesting that it's wrong btw, just curious about the semantics you're conveying 🙂

src/date.ts
```diff
@@ -98,6 +98,28 @@ function fromString(date: string): Option<Date> {
     }
 }
 
+function formatSeconds(seconds: string): string | null {
+    try {
```
JamieB-gu, line 102: Which statement here can throw an error?

src/date.ts
```diff
@@ -98,6 +98,28 @@ function fromString(date: string): Option<Date> {
...
+        if (hrs > 0) {
+            ret += `${hrs}:${mins < 10 ? "0" : ""}`
+        }
+
+        ret += `${mins}:${secs < 10 ? "0" : ""}${secs}`;
+        return ret;
+    } catch(e) {
+        return null;
+    }
+}
```
JamieB-gu, line 121: Is it worth using our `Option` type here?

```suggestion
function formatSeconds(seconds: string): Option<string> {
    try {
        const secondsInt = parseInt(seconds);
        if (isNaN(secondsInt)) {
            return none;
        }
        const hrs = Math.floor((secondsInt / 3600));
        const mins = Math.floor(((secondsInt % 3600) / 60));
        const secs = Math.floor(secondsInt) % 60;
        let ret = "";

        if (hrs > 0) {
            ret += `${hrs}:${mins < 10 ? "0" : ""}`
        }

        ret += `${mins}:${secs < 10 ? "0" : ""}${secs}`;
        return some(ret);
    } catch(e) {
        return none;
    }
}
```

src/components/shared/card.tsx
```diff
@@ -63,10 +71,19 @@ const headingWrapperStyles = css`
     min-height: 150px;
 `
 
-const headingStyles = css`
-    ${headline.xxxsmall()};
-    margin: 0;
+const headingStyles = (itemType: RelatedItemType) => {
+    if (itemType === RelatedItemType.ADVERTISEMENT_FEATURE){
```
frankie297, line 75: I think you might have committed to this PR by accident, isn't this for labs content cards?

src/components/shared/card.tsx
```diff
@@ -120,24 +140,19 @@ const cardStyles = (itemType: RelatedItemType, format: Format): SerializedStyles
...
-            return css``;
-        }
-
         case RelatedItemType.REVIEW: {
             return css``;
         }
 
         case RelatedItemType.ADVERTISEMENT_FEATURE: {
-            return css``;
+            return css`
```
frankie297, line 152: Again, I think you might want this in a separate PR

frankie297: Or alternatively, update this PR to include a description of all the changes. Ideally it would be better separated but it's up to you 😄

src/components/shared/card.tsx
```diff
@@ -14,39 +14,47 @@ import { RelatedItemType } from '@guardian/apps-rendering-api-models/relatedItem
...
-import { SvgCamera } from '@guardian/src-icons';
-
+import { SvgCamera, SvgVideo, SvgAudio } from '@guardian/src-icons';
 
 
 interface Props {
     relatedItem: RelatedItem;
     image: Option<Image>;
+
 }
```
frankie297, line 24:
```suggestion
interface Props {
    relatedItem: RelatedItem;
    image: Option<Image>;
}
```

src/components/shared/card.tsx
```diff
@@ -75,12 +92,13 @@ const imageWrapperStyles = css`
...
     date,
-    map(date => <time>{makeRelativeDate(date)}</time>),
+    map(date => <time css={[timeStyles, dateStyles]}>{makeRelativeDate(date)}</time>),
     withDefault<JSX.Element | null>(null),
 );
 
 const cardStyles = (itemType: RelatedItemType, format: Format): SerializedStyles => {
-    switch(itemType) {
+    
+    switch (itemType) {
```
frankie297, line 101:
```suggestion
    switch (itemType) {
```

PR #611, opened 2020-08-17
stop linting node_modules

webb04:
## Why are you doing this?
`node_modules` are being linted unnecessarily when running `npm run lint`. This adds about a minute to the running time.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 *.test.ts
+node_modules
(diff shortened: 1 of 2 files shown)


JamieB-gu:
Was this caused by #601 or has it been the case for a while?

webb04:
This has been around a while, maybe a new package included `.ts` files at some stage?

PR #618, opened 2020-08-18
mock related content

webb04:
## Why are you doing this?
This fetches related content from capi and renders it when running apps-rendering on port `8080`.
The logic is similar, so should make developing related content cards easier but doesn't match MAPI 100%. To get the same cards shown in production you will still need to run mapi and apps-rendering locally, accessing the article from mapi.


src/relatedContent.ts
```diff
@@ -0,0 +1,87 @@
...
+                },
+                starRating: content.fields?.starRating?.toString()
+            }
+        }).slice(0, 4)
+    }
+}
+
+export {
+    parseRelatedContent
+};
```
JamieB-gu, line 87: I think there are some tabs and spaces mixed together in this file - that's why the formatting looks erratic on GitHub.

src/relatedContent.ts
```diff
@@ -0,0 +1,87 @@
...
+	const optionalImage = pipe(
+        articleMainImage(content),
+        map(element => {
+            const masterAsset = element.assets.find(asset => asset?.typeData?.isMaster);
+            const data = element.imageTypeData;
+            return pipe2(
+				masterAsset,
+				fromNullable,
+				map(asset => ({
+					url: asset.file ?? '#',
```
JamieB-gu, line 45: What does the hash URL do here?

src/relatedContent.ts
```diff
@@ -0,0 +1,87 @@
...
+		return RelatedItemType.ADVERTISEMENT_FEATURE
+	} else {
+		return RelatedItemType.ARTICLE
+	}
+}
+
+const parseHeaderImage = (content: Content): Image | undefined => {
+	const optionalImage = pipe(
+        articleMainImage(content),
+        map(element => {
```
JamieB-gu, line 38: I think if you change this to an `andThen` you can avoid the nested `Option` below on line 54:

```suggestion
        andThen(element => {
```

src/server/server.ts
```diff
@@ -93,7 +100,9 @@ const parseCapiResponse = (articleId: string) =>
     }
 }
 
-const askCapiFor = (articleId: string): Promise<Result<number, Content>> =>
+type CapiReturn = Promise<Result<number, [Content, RelatedContent]>>;
```
JamieB-gu, line 103: Can you re-use this type for the return of `parseCapiResponse` up on line 68?

JamieB-gu (approved):

PR #642, opened 2020-08-25
Live:416 - added styling for comment cards and added capiendpoint for date

dskamiotis:
## Why are you doing this?

Styling needed for comment related cards to be consistent with headline and svgquote.
Date added is not rendering within the card also. 

## Changes

- Added background colour, font headline small, added co;our and size for quotation svg.
- Adde pillar styles for border-top to be active
- Rendered a new component tag `quotationComment` to be generated only when its a comment card
- Added specific styling for commentIconStyle to make quotation inline with headline text
- added endpoint for lastModified date in capi.ts to get date from capi (lastModified)

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/capi.ts
+++ b/src/capi.ts
@@ -128,7 +128,8 @@ const capiEndpoint = (articleId: string, key: string): string => {
         'displayHint',
         'starRating',
         'commentable',
-        'liveBloggingNow'
+        'liveBloggingNow',
+        'lastModified'
     ];
 
     const params = new URLSearchParams({
--- a/src/components/shared/card.tsx
+++ b/src/components/shared/card.tsx
@@ -212,11 +212,13 @@ const iconStyles = (format: Format): SerializedStyles => {
 
 const commentIconStyle = (): SerializedStyles => {
     return css`
-        width: 1.9rem;
+        width: 2.0rem;
         height: 1.4375rem;
         display: inline-block;
-        fill: ${opinion[400]};
-        vertical-align : text-top;
+        fill: #E05E00;
+        vertical-align: text-top;
+        margin-top: -3px;
+        margin-right: -2px;
     `;
 }
 
--- a/src/components/shared/card.tsx
+++ b/src/components/shared/card.tsx
@@ -6,7 +6,7 @@ import { remSpace, breakpoints, palette } from '@guardian/src-foundations';
 import { Option, withDefault, map, fromNullable, OptionKind } from '@guardian/types/option';
 import { makeRelativeDate, formatSeconds } from 'date';
 import { pipe2 } from 'lib';
-import { text, neutral, background } from '@guardian/src-foundations/palette';
+import { text, neutral, background, opinion } from '@guardian/src-foundations/palette';
 import { Design, Display, Format } from '@guardian/types/Format';
 import { Image } from 'image';
 import { darkModeCss } from 'styles';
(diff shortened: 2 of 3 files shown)


src/components/shared/card.tsx
```diff
@@ -220,6 +233,14 @@ const icon = (itemType: RelatedItemType, format: Format): ReactElement | null =>
     }
 }
 
+const quotationComment = (itemType: RelatedItemType, format: Format): ReactElement | null => {
```
webb04, line 236: Nice 🎉

src/capi.ts
```diff
@@ -128,7 +128,8 @@ const capiEndpoint = (articleId: string, key: string): string => {
         'displayHint',
         'starRating',
         'commentable',
-        'liveBloggingNow'
+        'liveBloggingNow',
+        'lastModified'
```
webb04, line 132: You'll need to update `capi.test.ts` with this field

src/components/shared/card.tsx
dskamiotis: thanks 👍

src/capi.ts
dskamiotis: Not sure if i have to create a new field as the date created is dynamic

webb04: It will always be "lastModified" in the request and then the response will be different. The test is expecting this string:

https://content.guardianapis.com/cities/2019/sep/13/reclaimed-lakes-and-giant-airports-how-mexico-city-might-have-looked?format=thrift&api-key=TEST_KEY&show-atoms=all&show-fields=headline%2Cstandfirst%2CbylineHtml%2CfirstPublicationDate%2CshouldHideAdverts%2CshouldHideReaderRevenue%2CdisplayHint%2CstarRating%2Ccommentable%2CliveBloggingNow%2ClastModified&show-tags=all&show-blocks=all&show-elements=all&show-related=true

dskamiotis: Thanks Jamie, how did you come up with that string?

src/components/shared/card.tsx
```diff
@@ -207,6 +210,18 @@ const iconStyles = (format: Format): SerializedStyles => {
     `;
 }
 
+const commentIconStyle = (): SerializedStyles => {
+    return css`
+        width: 2.0rem;
+        height: 1.4375rem;
+        display: inline-block;
+        fill: #E05E00;
```
webb04, line 218: Can we use the opinion colour from the design system? you had it as `fill: ${opinion[400]};` before

dskamiotis: Ah yes your right, i took from storybook when adjusting for size of speech mark and must've pasted raw clor

webb04 (approved):
Good job 👍

src/components/shared/card.tsx
```diff
@@ -93,7 +93,7 @@ const headingStyles = (itemType: RelatedItemType): SerializedStyles => {
         `;
     } else {
         return css`
-            ${headline.xxxsmall()};
+            ${headline.xxsmall()}
```
mxdvl, line 96: Why did the semi-colon `;` got removed?

src/components/shared/card.tsx
```diff
@@ -207,6 +210,18 @@ const iconStyles = (format: Format): SerializedStyles => {
     `;
 }
 
+const commentIconStyle = (): SerializedStyles => {
+    return css`
+        width: 2.0rem;
+        height: 1.4375rem;
+        display: inline-block;
+        fill: ${opinion[400]};;
```
mxdvl, line 218: Double semi-colon there : `;;`!

src/components/shared/card.tsx
```diff
@@ -175,7 +175,10 @@ const cardStyles = (itemType: RelatedItemType, format: Format): SerializedStyles
         }
 
         case RelatedItemType.COMMENT: {
-            return css``;
+            return css`
+                background-color : ${opinion[800]};
+                ${headline.xxsmall()}
```
mxdvl, line 180: Another missing semi-colon `;`.

mxdvl (commented):
This looks good! I would get a designer's feedback as the icon feels like it could be pushed another pixel or two to the left.

I've captured a few odd semi-colons, but they'll need to be captured in a subsequent PR.

👍

PR #644, opened 2020-08-26
Live-125: Changed dateline colour neutral.20

dskamiotis:
## Why are you doing this?

Contrast of dateline colour needs to be clearer upon rendering to stand out on background.

## Changes

- Changed dateline colour to neutral.20
- Omitted text from import - deprecated 

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/dateline.stories.tsx
+++ b/src/components/dateline.stories.tsx
@@ -5,15 +5,16 @@ import { date, withKnobs } from '@storybook/addon-knobs';
 import { some } from '@guardian/types/option';
 import Dateline from './dateline';
 import { Pillar } from '@guardian/types/Format';
+import { selectPillar } from 'storybookHelpers';
 
-type DefaultProps = {
-    pillar: Pillar
-}
 
 // ----- Stories ----- //
 
-const Default: FC<DefaultProps> = ({pillar}:DefaultProps) =>
-    <Dateline pillar={pillar} date={some(new Date(date('Publish Date', new Date('2019-12-17T03:24:00'))))} />
+const Default: FC = () =>
+    <Dateline 
+        pillar={selectPillar(Pillar.Opinion)} 
+        date={some(new Date(date('Publish Date', new Date('2019-12-17T03:24:00'))))} 
+    />
 
 
 // ----- Exports ----- //
--- a/src/components/dateline.stories.tsx
+++ b/src/components/dateline.stories.tsx
@@ -2,7 +2,6 @@
 
 import React, { FC } from 'react';
 import { date, withKnobs } from '@storybook/addon-knobs';
-
 import { some } from '@guardian/types/option';
 import Dateline from './dateline';
 import { Pillar } from '@guardian/types/Format';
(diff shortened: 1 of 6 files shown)


src/components/dateline.tsx
```diff
@@ -23,7 +23,7 @@ const darkStyles = darkMode`
 
 const styles = css`
     ${textSans.xsmall()}
-    color: ${text.supporting};
+    color: ${neutral[20]};
```
webb04, line 26: I think we just want this darker colour for opinion types and `text.supporting` for the others

src/components/shared/card.tsx
```diff
@@ -208,6 +208,9 @@ const iconStyles = (format: Format): SerializedStyles => {
 }
 
 const icon = (itemType: RelatedItemType, format: Format): ReactElement | null => {
+    // console.log(itemType)
```
webb04, line 211: Can remove these ㏒s

webb04 (changes requested):

src/components/dateline.tsx
dskamiotis: 👍 will change - also UI tests seem to be stalled

webb04: I think that's Chromatic picking up a UI change in the dateline component! can show you how to approve these once you've made the changes

JamieB-gu:
I think the change for DCR will be to [this component](https://github.com/guardian/dotcom-rendering/blob/master/src/web/components/Dateline.tsx).

webb04:
Think you'll have to update the stories to use the new `pillar` prop for `dateline` components https://github.com/guardian/apps-rendering/blob/bf62d6577909881f39abc60675bac51e5f03868e/src/components/dateline.stories.tsx

src/components/dateline.stories.tsx
```diff
@@ -2,15 +2,18 @@
...
 
+type DefaultProps = {
+    pillar: Pillar
+}
 
 // ----- Stories ----- //
 
-const Default: FC = () =>
-    <Dateline date={some(new Date(date('Publish Date', new Date('2019-12-17T03:24:00'))))} />
+const Default: FC<DefaultProps> = ({pillar}:DefaultProps) =>
```
webb04, line 15: I think you can pass pillar directly into Dateline like they pass the props in here? 

https://github.com/guardian/apps-rendering/blob/a35bbe2cc861099b157888a2692e2ecc8a81f665/src/components/anchor.stories.tsx#L19

dskamiotis: Trying to replica the model with little success:
const Default: FC = () =>
    <Dateline 
        pillar={{
            pillar: Pillar
        }}

webb04: I think what you have now is ok, you just might not need type and argument to the default component?

<Dateline pillar={pillar} date={some(new Date(date('Publish Date', new Date('2019-12-17T03:24:00'))))} />

dskamiotis: coming up with error: Cannot find name pillar when i do that, thats why i initially used the prop DefaultProps to feed it in.

webb04 (approved):

PR #657, opened 2020-09-01
Live:353- Sponsor offline links colour as kicker

dskamiotis:
## Why are you doing this?

Currently when offline the alt text defaults to the HTML blue.
We would like the alt text to be consistent with the Pillar colour, as well as inverted when in dark mode.
## Changes

- Added kicker colour from Pillar, for text to render when offline/missing logo
- Included inverted colour for dark mode
- Included format prop in Logo component to read Pillar type

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/shared/logo.tsx
+++ b/src/components/shared/logo.tsx
@@ -25,7 +25,7 @@ const styles = (format: Format,
         ${textSans.small()}
 
         img {
-            content: url("${darkModeImage ?? lightModeImage}");
+            content: url("${lightModeImage}");
             display: block;
             margin: ${remSpace[2]} 0;
             max-height: 60px;
@@ -49,7 +49,7 @@ const styles = (format: Format,
             }
 
             a {
-                color: ${inverted}
+                color: ${inverted};
             }
         `}
     `;
@@ -66,7 +66,7 @@ export const cleanImageUrl = (url: string): string =>
         .replace(/\(/g, '%28')
         .replace(/\)/g, '%29')
 
-const Logo: FC<Props> = ({ branding, format }: Props) => {    
+const Logo: FC<Props> = ({ branding, format }: Props) => {
     const lightLogo = cleanImageUrl(branding.logo);
     const darkLogo = cleanImageUrl(branding.altLogo ?? branding.logo);
 
--- a/src/components/shared/logo.tsx
+++ b/src/components/shared/logo.tsx
@@ -23,6 +23,7 @@ const styles = (format: Format, lightModeImage: string, darkModeImage?: string):
         ${textSans.small()}
 
         img {
+            content: url("${darkModeImage ?? lightModeImage}");
             display: block;
             margin: ${remSpace[2]} 0;
             max-height: 60px;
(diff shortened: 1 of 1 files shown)


src/components/shared/logo.tsx
```diff
@@ -9,37 +9,51 @@ import { Item, getFormat } from 'item';
...
+            img {
+                content: url("${darkModeImage ?? lightModeImage}");
+            }
+
+            label {
+                color: ${neutral[86]};
+            }
+
+            a {
+                color: ${inverted}
```
JamieB-gu, line 52:
```suggestion
                color: ${inverted};
```

src/components/shared/logo.tsx
```diff
@@ -9,37 +9,51 @@ import { Item, getFormat } from 'item';
...
-        max-height: 60px;
-    }
-
-    label {
-        color: ${text.supporting};
-    }
-
-    ${darkModeCss`
         img {
             content: url("${darkModeImage ?? lightModeImage}");
```
JamieB-gu, line 28: I believe this is the image URL in light mode?

```suggestion
            content: url("${lightModeImage}");
```

src/components/shared/logo.tsx
```diff
@@ -52,12 +66,12 @@ export const cleanImageUrl = (url: string): string =>
         .replace(/\(/g, '%28')
         .replace(/\)/g, '%29')
 
-const Logo: FC<Props> = ({ branding, format }: Props) => {
+const Logo: FC<Props> = ({ branding, format }: Props) => {
```
JamieB-gu, line 69:
```suggestion
const Logo: FC<Props> = ({ branding, format }: Props) => {
```

JamieB-gu (commented):
Couple of small comments, otherwise looks good!

src/components/shared/logo.tsx
dskamiotis: Not 100% sure to be honest, I assumed `darkModeImage ??` meant it would load a different image in dark mode, and if one was not found it would back up to the original dark mode.

JamieB-gu: `??` is the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) - it attempts to use the first value, but if that value is `null` or `undefined` it falls back to the second. In this case it attempts to use the value of `darkModeImage`, and if that value is missing it falls back to using the `lightModeImage`. This is desirable in situations where we've only been provided with a light mode version of a logo: the code will use that for both light mode and dark mode.

However, we only want to make an attempt to use a dark mode variant of the logo *when in dark mode*. Our `darkModeCss` function is what allows us to set styles that are only applied when the user is in dark mode. Therefore, any CSS relevant to dark mode should *only be applied via this function*. In summary, on this line we're not using the `darkModeCss` function, so we only want to provide the `lightModeImage`. Down on line 44 we *are* using the `darkModeCss` function so that's where we provide the `darkModeImage`.

dskamiotis: Ah ok, thanks for that, to be honest i was looking at the dark mode css, so it is redundant where it is as we are not in dark mode at that part of the css.
Thanks for the explanation by the way I do really appreciate the time detail these help me a great deal (I copy them into my notes for later reference when stuck!)
Also i was looking at the line 44 instance of it 🤦‍♂️

JamieB-gu (approved):

PR #658, opened 2020-09-01
LIVE-557: Build rendered items assets html

faresite:
## Why are you doing this?

We need a html document containing all possible assets and fonts to prewarm the cache on iOS. Using webpack to create the html as a build step seems like the best place to do it, in case there would be more js chunks in the future there will be no need to update the template.

@webb04 and I paired on [this PR](https://github.com/guardian/mobile-apps-api/pull/1644), however I'm not sure we need this anymore as looks like the MAPI `/assets` endpoint is already setup to route to AR express server's `/assets` endpoint (for example [https://mobile.guardianapis.com/assets/manifest.json](https://mobile.guardianapis.com/assets/manifest.json)). Webpack puts the generated html in the assets folder which hopefully means we can benefit from the fastly config out of the box since it's in the same `/assets` route as the JS and `manifest.json` assets. Correct me if I'm wrong?!

In the template itself, you will notice I created non-empty DOM elements to use the fonts, have I not added any text, the browser wouldn't request the font, so this was my work around to force request all fonts.

## Changes

- Install the html webpack plugin
- Create html template for webpack to use that includes fonts


config/rendered-items-assets-template.html
```diff
@@ -0,0 +1,144 @@
...
+    <div style="font-family: GH Guardian Headline; font-weight: 300;">.</div>
+    <div style="font-family: GH Guardian Headline; font-weight: 300; font-weight: bold;">.</div>
+    <div style="font-family: GH Guardian Headline; font-weight: 400;">.</div>
+    <div style="font-family: GH Guardian Headline; font-weight: 400; font-weight: bold;">.</div>
+    <div style="font-family: GH Guardian Headline; font-weight: 500;">.</div>
+    <div style="font-family: GH Guardian Headline; font-weight: 500; font-weight: bold;">.</div>
+    <div style="font-family: GH Guardian Headline; font-weight: 600;">.</div>
+    <div style="font-family: GH Guardian Headline; font-weight: 600; font-weight: bold;">.</div>
+    <div style="font-family: GH Guardian Headline; font-weight: 700;">.</div>
+    <div style="font-family: GH Guardian Headline; font-weight: 700; font-weight: bold;">.</div>
```
JamieB-gu, line 141: A few duplicate `font-weight`s in here?

webpack.config.js
```diff
@@ -167,6 +168,11 @@ const clientConfigProduction = {
     devtool: false,
     plugins: [
         new ManifestPlugin(),
+        new HtmlWebpackPlugin({
+            filename: 'rendered-items-assets.html',
+            template: path.resolve(__dirname, `config/rendered-items-assets-template.html`),
```
JamieB-gu, line 173:
```suggestion
            template: path.resolve(__dirname, 'config/rendered-items-assets-template.html'),
```

config/rendered-items-assets-template.html
faresite: the art of copy pasting .. will fix that later 👍

JamieB-gu:
It would be good to set a CSP for this template. Looks like the plugin [provides functionality](https://github.com/jantimon/html-webpack-plugin#simulate-http-response-headers) for that. If we combine that with [template parameters](https://github.com/jantimon/html-webpack-plugin/tree/master/examples/template-parameters) for the inline styles we can use our [existing asset hashing function](https://github.com/guardian/apps-rendering/blob/f518b9ed4372a714c243d43f29ad09ad0c4e0793/src/server/csp.ts#L46) to explicitly allow these inline styles in said CSP.

webb04:
Looks good! did you want to request the `js` in `rendered-items-assets-template.html` too?

I think we can test the Fastly stuff by checking headers and maybe the Fastly dashboard?

faresite:
> Looks good! did you want to request the `js` in `rendered-items-assets-template.html` too?
> 
> I think we can test the Fastly stuff by checking headers and maybe the Fastly dashboard?

@webb04 you'll be pleased to know that all JS chunks are included by default :)

webb04 (approved):

webb04:
Nice! would the caching work.. because we don't have the hash in the filename?

faresite:
> Nice! would the caching work.. because we don't have the hash in the filename?

If it's all working nicely with fastly, then the js hashes inside the html would change and so the etag would change too? Definitely need to check on MAPI code 👍

Also this PR is not ready yet because I'll need to update it following the TS config PR and add CSP to the html doc.

webpack.config.js
```diff
@@ -3,7 +3,16 @@
 const { fork } = require('child_process');
 const webpack = require('webpack');
 const path = require('path');
+const { createHash } = require('crypto');
+const CleanCSS = require('clean-css');
```
JamieB-gu, line 7: We don't appear to have a dependency on this 🤔. If it currently exists only as an indirect dependency can we make our requirement for it explicit by adding it as a direct dependency?

faresite: yes 👍 was planning to but forgot!

webpack.config.ts
```diff
@@ -161,13 +165,28 @@ export const clientConfig: Configuration = {
     }
 };
 
+const assetsTemplateCss = new CleanCSS().minify(renederedItemsAssetsCss).styles.trim()
+const assetHash = (asset: string): string =>
+    createHash('sha256').update(asset).digest('base64');
+
```
faresite, line 171: Unfortunately couldn't get this to work from `server/csp.ts` and spent enough time trying to make it happen. Kind of means there was not much of a point to do this PR with TypeScript but at least we have a TS webpack config now.

config/rendered-items-assets-template.html
```diff
@@ -0,0 +1,32 @@
...
+    <div class="l">.</div>
+    <div class="m">.</div>
+    <div class="n">.</div>
+    <div class="o">.</div>
+    <div class="p">.</div>
+    <div class="q">.</div>
+    <div class="r">.</div>
+    <div class="s">.</div>
+    <div class="t">.</div>
+    <div class="u">.</div>
```
webb04, line 29: So close... I'm sure we can find some more in fontastique?

webb04: maybe we use font names for classes to avoid lookups?

webb04 (approved):

config/rendered-items-assets-template.html
faresite: @webb04 👍 I've renamed the classes. I've taken reference from the `styles.ts` file. There were some duplicate font urls so pruned those.

webb04 (approved):

PR #675, opened 2020-09-04
LIVE-575: Webpack config in TypeScript

faresite:
## Why are you doing this?
While working on [This PR](https://github.com/guardian/apps-rendering/pull/658) it would be helpful to use some of the code that's written in TS in our webpack config, however the config was plain JS. Instead of copy pasting the code, making our Webpack run off TS was straight forward enough and it's a small win on its own.


faresite:
Ignore the failed build, fixing the storybook config.

JamieB-gu (commented):
Pending the lint issues, looks great!

webpack.config.ts
```diff
@@ -36,11 +37,11 @@ function resolve(loggerName) {
             logger: path.resolve(__dirname, `src/logger/${loggerName}`)
         },
     }
-};
+}
 
 // ----- Configs ----- //
 
-const serverConfig = env => {
+const serverConfig = (env: { [key: string]: boolean | undefined }): Configuration => {
```
faresite, line 44: Not proud of this one, but turns out `NodeJS.ProcessEnv` was the wrong one here after adding the `Configuration` return type. It seems to me that Webpack does the magic of parsing the CLI flags but I couldn't find a helpful type for it https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/webpack/index.d.ts

webpack.config.ts
```diff
@@ -177,4 +178,4 @@ const clientConfigProduction = {
 
 // ----- Exports ----- //
 
-module.exports = [ serverConfig, clientConfig, clientConfigProduction ];
+export default [ serverConfig, clientConfig, clientConfigProduction ];
```
faresite, line 181: Not sure we need this export but kept it just in case I'm not aware of something.

.storybook/webpack.config.ts
```diff
@@ -0,0 +1,16 @@
+import { Configuration } from 'webpack';
+import { clientConfig } from '../webpack.config';
+
+export default async ({ config } : { config: Configuration }) : Promise<Configuration> => {
+	return {
```
faresite, line 5: The closest type I found was `ConfigurationFactory` in [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/77357b2c3de87fdaee7aa75487b5e767fa26dbec/types/webpack/index.d.ts#L154) but it's not it as the config there is a string.

webb04 (approved):

.storybook/webpack.config.ts
```diff
@@ -0,0 +1,16 @@
...
+		...config,
+		target: clientConfig?.target,
+		resolve: clientConfig?.resolve,
+		module: {
+			rules: clientConfig?.module?.rules ?? [],
+		},
+		plugins: [
+			...config.plugins ?? []
```
JamieB-gu, line 13: If `config.plugins` is undefined will this result in `[[]]`?

JamieB-gu (approved):

.storybook/webpack.config.ts
faresite: I was puzzled by this a bit too but what it actually does is something like `...(config.plugin ?? [])`

PR #684, opened 2020-09-07
Add PR deployment action

jacobwinch:
## Why are you doing this?

This PR provides an easy mechanism for deploying a branch to a public URL. In the short-term this can be used for manually testing changes (either your own, or someone else's, as part of the code review process). Longer term, exposing a public URL for every PR allows us to easily run additional tasks/checks (e.g. taking screenshots, running performance tests etc.). 

Thanks to @jorgeazevedo who did all of the hard work on this with Dotcom (see https://github.com/guardian/dotcom-rendering/pull/1841) - I've borrowed liberally from his work here!

Whilst under development, this check won't (automatically) run against every PR - see `How will it work?` section for more details.

## Changes

- Add new GitHub workflow
- Modify config fetching code to avoid call to SSM if running in the GH Actions environment

## How will it work?

Adding the `PR Deployment` label starts the new workflow*:

image

Once the workflow has started the application, another status check will appear. This provides a link to the running environment:

image

The check will only run if the `PR Deployment` label is added to a PR. Adding any other label will produce a 'skipped' check:

image

*Note that this workflow will keep running (and retain the yellow 'pending' dot) until the timeout is reached. Developers do not need to wait for this check to 'finish' before merging.


webb04 (approved):
This is great! thanks a 

Is this time the creation time rather than the expiry? "This PR is now live until Fri 08:57. Click details to access it ->"

Or is it open for 12 hours?

jacobwinch:
> Is this time the creation time rather than the expiry? "This PR is now live until Fri 08:57. Click details to access it ->"

Good catch, we're adding an hour to the current time (``date -d "+1 hour" "+%a %H:%M"``), but the timezone on the GitHub Containers is UTC (which is 1 hour behind BST), so we end up with the current time! This should fix it: https://github.com/guardian/apps-rendering/pull/684/commits/3d7476111e561bd6ca1edc41973fdef4f0f3b469

image

JamieB-gu:
FYI @alexduf

PR #692, opened 2020-09-08
Live:104 - converted star rating icon to svg

dskamiotis:
## Why are you doing this?

Currently Star Rating is using icons which cause issues when rendering in certain browser and ios/android versions, with strange fallback icons when not loading (square box).

Issues regarding Icons and how SVg are more desirable:

We could provide PNG icons as fallback, but since we have SVG's as part of source not necessary.

## Changes

- Customised styles for the new SVG element
- Included SvgStar instead of the icon as a tag

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/starRating.tsx
+++ b/src/components/starRating.tsx
@@ -22,7 +22,7 @@ const starStyles = css`
 ;
     svg {
         margin: 0 -0.125rem;
-        height: 2rem;
+        height: 1.75rem;
     }
 
     ${darkModeCss`
--- a/src/components/starRating.tsx
+++ b/src/components/starRating.tsx
@@ -1,32 +1,23 @@
 // ----- Imports ----- //
 
-import React, { ReactNode, ReactElement } from 'react';
+import React, { ReactNode, FC } from 'react';
 import { css } from '@emotion/core';
-import { brandAltBackground, text } from '@guardian/src-foundations/palette';
-import { remSpace } from '@guardian/src-foundations';
-
+import { brandAltBackground, brandAltLine } from '@guardian/src-foundations/palette';
+import { SvgStar } from '@guardian/src-icons';
 import { Item } from 'item';
-import { icons, darkModeCss } from 'styles';
+import { darkModeCss } from 'styles';
 import { Design } from '@guardian/types/Format';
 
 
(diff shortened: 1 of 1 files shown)


dskamiotis:
@benwuersching I have added you as we re factored the construction of the SvgStar for star rating and tried to keep as similar to DCR and code live.

benwuersching:
@dskamiotis This looks good, thank you. The new stars seem ever so slightly further apart. Could they move slightly closer to each other?

JamieB-gu:
> Could they move slightly closer to each other?

I think this is actually because the component in Source contains padding on the left and right. @SiAdcock do you think we should adjust this here in CSS or tweak the SVG in Source?

src/components/starRating.tsx
```diff
@@ -1,56 +1,53 @@
...
-
-    &:nth-of-type(1) {
-        padding-left: ${remSpace[1]};
-    }
+    line-height: 0;
 
-    &:nth-of-type(5) {
-        padding-right: ${remSpace[1]};
+    svg {
+        height: 2rem;
```
webb04, line 20: Do we lose some of that right padding?

src/components/starRating.tsx
```diff
@@ -16,7 +16,12 @@ const starStyles = css`
     display: inline-block;
     line-height: 0;
 
+    &:nth-of-type(5) {
+        padding-right: 2px;
+    }
+;
     svg {
+        margin: 0 -2px;
```
JamieB-gu, line 24: You may want to convert this to `rem` to handle scaling.

src/components/starRating.tsx
```diff
@@ -1,56 +1,58 @@
...
+    line-height: 0;
 
     &:nth-of-type(5) {
-        padding-right: ${remSpace[1]};
+        padding-right: 2px;
+    }
+;
+    svg {
+        margin: 0 -0.125rem;
+        height: 2rem;
```
webb04, line 25: This brings the size a bit closer to dotcom and the current apps-rendering articles:

```suggestion
        height: 1.75rem;
```

webb04 (approved):

PR #696, opened 2020-09-09
Live 603 icon svg headerimagecaption

dskamiotis:
## Why are you doing this?

Currently Header Image Caption renders an icon of a camera using icons which cause issues when rendering in certain browser and ios/android versions, with strange fallback icons when not loading (square box).

Issues regarding Icons and how SVG are more desirable:

We could provide PNG icons as fallback, but since we have SVG's as part of source not necessary.

## Changes

- Have eliminated use of icon and included the SvgCamera
- Modified the overall styling for header image captions
- Added a customised svg styling for the camera svg and added the styling to the surrounding span tag
- Have tried to keep the ration of image similar to mobile

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/headerImageCaption.tsx
+++ b/src/components/headerImageCaption.tsx
@@ -14,7 +14,7 @@ const captionId = 'header-image-caption';
 
 const HeaderImageCaptionStyles = css`
 	summary {
-		line-height: 30px;
+		/* line-height: 30px; */
         text-align: center;
         background-color: ${brandAlt[400]};
         color: ${neutral[7]};
@@ -30,11 +30,6 @@ const HeaderImageCaptionStyles = css`
 		
 		span {
 			font-size: 0;
-		}
-
-		svg {
-			position: absolute;
-			right: ${basePx(0.25)};
 		}
 
 		&::-webkit-details-marker {
@@ -61,6 +56,7 @@ const HeaderImageCaptionStyles = css`
 
 		${darkModeCss`
 			color: ${neutral[60]};
+			height: 5rem;
 		`}
 	}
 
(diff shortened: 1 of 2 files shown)


src/components/headerImageCaption.tsx
```diff
@@ -62,6 +48,7 @@ const HeaderImageCaptionStyles = css`
 
 		${darkModeCss`
 			color: ${neutral[60]};
+			height: 5rem;
```
JamieB-gu, line 51: The `height` changes when in dark mode?

src/components/headerImageCaption.tsx
```diff
@@ -75,18 +62,28 @@ const HeaderImageCaptionStyles = css`
 	}
 `;
 
+const svgStyle = css`
+	line-height: 55px;
```
JamieB-gu, line 66: Can we convert this to `rem` for scaling purposes? I think it will be approximately `3.44rem`.

src/components/headerImageCaption.tsx
```diff
@@ -75,18 +62,28 @@ const HeaderImageCaptionStyles = css`
...
+	font-size: 0;
+	svg {
+		width: ${basePx(2.7)};
+    	height: ${basePx(2.7)};
```
JamieB-gu, line 70: `basePx` is deprecated, please use `1.35rem` here instead:

```suggestion
        width: 1.35rem;
        height: 1.35rem;
```

src/renderer.test.ts
```diff
@@ -217,13 +218,13 @@ describe('Renders different types of elements', () => {
...
         const pullquote = nodes.flat()[0];
-        expect(getHtml(pullquote)).toContain('<blockquote><p>quote</p></blockquote>');
+        expect(getHtml(pullquote)).toContain('<aside><blockquote><p>quote</p></blockquote></aside>');
     })
 
     test('ElementKind.Pullquote with attribution', () => {
         const nodes = render(pullquoteWithAttributionElement())
         const pullquote = nodes.flat()[0];
-        expect(getHtml(pullquote)).toContain('<blockquote><p>quote</p><cite>attribution</cite></blockquote>');
+        expect(getHtml(pullquote)).toContain('<aside><blockquote><p>quote</p><cite>attribution</cite></blockquote></aside>');
```
JamieB-gu, line 227: Have these been pulled in from some other changes?

JamieB-gu (commented):
Nice! A few questions/comments.

src/renderer.test.ts
dskamiotis: Yes it was pulled in from prev ticket, ill update and check both as the test was written

src/components/headerImageCaption.tsx
dskamiotis: not sure how that got there!

JamieB-gu (approved):

JamieB-gu:
Think we might be seeing a couple visual bugs with this in Safari and Firefox:

| Safari | Firefox |
| - | - |
| safari | firefox |

PR #701, opened 2020-09-10
Profile, Timeline and Chart atoms

webb04:
## Why are you doing this?
LIVE-503, LIVE-504, LIVE-505

## Changes

- Bump `atoms-rendering` to version `1.12.0`
- Add images and credits to previously added guide and qandaatoms
- Support profile atoms
- Support timeline atoms
- Adds support for chart atoms but hides them until we find a solution to the inline styles

## Screenshots

| Guide Atom Before | Guide Atom After |
| --- | --- |
|  |  |

### Profile
Screen Shot 2020-09-10 at 10 51 20

### Timeline
Screen Shot 2020-09-10 at 12 01 16

### Chart
Screen Shot 2020-09-11 at 14 53 14


src/bodyElement.ts
```diff
@@ -82,18 +94,39 @@ interface MediaAtom {
...
+interface QandaAtom extends ExpandableAtom {
     kind: ElementKind.QandaAtom;
-    html: string;
+}
+
+interface ProfileAtom extends ExpandableAtom {
+    kind: ElementKind.ProfileAtom;
+}
+
+interface TimelineItem {
```
JamieB-gu, line 117: Do you reckon it would be possible to `import` this, and possibly some of the atom interfaces, from `atoms-rendering`?

JamieB-gu (approved):
Nice!

src/bodyElement.ts
webb04: They do export that type so I'm using it now.

src/atoms.ts
```diff
@@ -4,6 +4,20 @@ import { Result, err, ok } from '@guardian/types/result';
 import { BodyElement, ElementKind } from "bodyElement";
 import { fromNullable } from '@guardian/types/option';
 import { DocParser } from "types/parserContext";
+import Int64 from 'node-int64';
+import { isValidDate } from "date";
+import { atomScript } from "components/atoms/interactiveAtom";
+
+function formatDate(date: Int64): string {
+    return new Date(date.toNumber()).toDateString();
```
JamieB-gu, line 12: Is it worth running `isValidDate` on this too?

src/atoms.ts
```diff
@@ -78,7 +98,98 @@ function parseAtom(
...
+            const events = atom.data.timeline.events.map(event => ({
+                title: event.title,
+                date: formatDate(event.date),
+                body: event.body,
+                toDate: formatOptionalDate(event.toDate),
+            }));
+
+            const description = atom.data.timeline.description;
+
+            if (!title || !events.length) {
```
JamieB-gu, line 183:
```suggestion
            if (!title || events.length === 0) {
```

src/renderer.ts
```diff
@@ -517,6 +517,33 @@ const render = (format: Format, excludeStyles = false) =>
...
+                likeHandler: () => { console.log("like clicked"); },
+                dislikeHandler: () => { console.log("dislike clicked"); },
+                expandCallback: () => { console.log("expand clicked"); }
+            })
+        }
+
+        case ElementKind.ChartAtom: {
+            return h(ChartAtom, {
+                ...element,
+                url: ""
```
JamieB-gu, line 543: What's the `url` field meant to be used for?

src/server/csp.ts
```diff
@@ -60,7 +67,7 @@ const assetHashes = (assets: string[]): string =>
 
 const buildCsp = ({ styles, scripts }: Assets, twitter: boolean): string => `
     default-src 'self';
-    style-src 'self' ${assetHashes(styles)} https://interactive.guim.co.uk ${twitter ? 'https://platform.twitter.com' : ''};
+    style-src 'self' 'unsafe-hashes' ${assetHashes(styles)} https://interactive.guim.co.uk ${twitter ? 'https://platform.twitter.com' : ''};
```
JamieB-gu, line 70: Have you included this because there are `style` attributes on tags inside Chart atoms?

JamieB-gu (commented):
This is great!

A few questions.

src/renderer.ts
webb04: It's actually not used, I'll try to make a PR to atoms-rendering.

src/server/csp.ts
webb04: yes, I managed to add the inline styles to the csp, but it looks like it still requires this.

https://github.com/guardian/apps-rendering/pull/701/files#diff-153287d02c1f9fa0719c0f23a8a8198eR150

src/renderer.ts
webb04: It's already been updated, let me bump atoms-rendering

src/server/csp.ts
JamieB-gu: What does `unsafe-hashes` do when applied to `style-src`?

webb04: There is not much information I could find, but it seems to enable inline style hashes. Without it, the inline style hashes are not accepted by the csp.

Screen Shot 2020-09-18 at 14 22 58


| With | Without |
| --- | --- |
|  |  |

JamieB-gu: When `unsafe-hashes` *only* exists in `style-src`, not `script-src`, does it just apply to `style` attributes? Both event handlers and `javascript:` navigation targets remain forbidden?

src/atoms.ts
```diff
@@ -78,7 +94,102 @@ function parseAtom(
...
+                toDate: formatOptionalDate(event.toDate),
+            }));
+
+            const description = atom.data.timeline.description;
+
+            if (!title || events.length === 0) {
+                return err(`No title or body for atom: ${id}`);
+            }
+
+            if (!events.every(event => event.date !== '')) {
```
JamieB-gu, line 183: Alternatively?

```suggestion
            if (events.some(event => event.date === '')) {
```

src/server/csp.ts
webb04: `unsafe-hashes` is CSP Level 3 which is only supported in Chrome and not in Safari so this won't work.

The only other thing I can think of is to match dotcom by doing
`script-src https: 'unsafe-inline';`

src/server/csp.ts
```diff
@@ -67,7 +67,7 @@ const assetHashes = (assets: string[]): string =>
 
 const buildCsp = ({ styles, scripts }: Assets, twitter: boolean): string => `
     default-src 'self';
-    style-src 'self' 'unsafe-hashes' ${assetHashes(styles)} https://interactive.guim.co.uk ${twitter ? 'https://platform.twitter.com' : ''};
+    style-src https: 'unsafe-inline';
```
JamieB-gu, line 70: This is something I really want to avoid if possible, it undoes a lot of the work of the CSP. How broken does it look without these `style` attributes?

webb04: Unfortunately not great on Safari

93602976-4c264b00-f9bb-11ea-9c4d-6fb760e06721

webb04: We could write a css-in-js like library to abstract these inline styles into `style` elements and generate unique classnames... 😅

JamieB-gu: I don't know if you're serious but I did actually think about it 🙂. Decided that a) it might be too complicated and brittle, and b) we control atoms, if we're going to be doing a bunch of work I'd prefer to do it at the other end.

webb04:
@JamieB-gu I've left in the work to render chart atoms as I would imagine the solution would require something similar of using the component in `atoms-rendering`.

I'm returning an error in the renderer so they won't be showing yet and the CSP has been reverted.

src/components/atoms/interactiveAtom.tsx
```diff
@@ -38,7 +38,8 @@ const atomScript = `
         window.frameElement.height = document.body.offsetHeight;
     }
     window.addEventListener('resize', resize);
-    resize();`
+    setTimeout(resize, 1000);
```
JamieB-gu, line 41: What does this do?

JamieB-gu (approved):
Nice!

PR #702, opened 2020-09-10
restrict headline sizes to pixels

webb04:
## Why are you doing this?
Feedback from the Beta showed the headlines sizes were very large, particularly when the font size was increased.

Setting the font-size to pixels will stop the headlines from changing size.

| Before | After |
| --- | --- |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |

### DIFF
--- a/src/components/headline.tsx
+++ b/src/components/headline.tsx
@@ -98,6 +98,11 @@ const immersiveLabs = css`
     }
 `;
 
+// stop headlines from growing in size with font resizer
+const fontSizeRestriction = css`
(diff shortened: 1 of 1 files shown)


JamieB-gu (approved):
👍 let's try this out and see how it's received. Do you have any screenshots btw?

The downside to this is that all headlines have the same font size at all breakpoints 🤔, overriding the rules the design system has laid down. This was the approach we said we'd try out at the editorial design group @SiAdcock

PR #703, opened 2020-09-10
Add link to fontSize.css

maxspencer:
## Why are you doing this?

So that the Android native layer can intercept this request for `fontSize.css` and provide the correct "base" font size for the page.

@JamieB-gu I think putting `/fontSize.css` should cause the request to go to the MAPI domain as we discussed?

## Changes

- Add a stylesheet link to the head of all pages.

This PR is a replacement for #346

### DIFF
--- a/src/server/page.tsx
+++ b/src/server/page.tsx
@@ -94,7 +94,7 @@ function renderHead(
 
     return `
         ${renderToString(meta)}
-        <link rel="stylesheet" types="text/css" href="/fontSize.css"/>
+        <link rel="stylesheet" type="text/css" href="/fontSize.css">
         <style>${generalStyles}</style>
         <style data-emotion-css="${emotionIds.join(' ')}">${itemStyles}</style>
         <script id="targeting-params" type="application/json">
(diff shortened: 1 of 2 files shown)


maxspencer:
@JamieB-gu cool, done 👍

src/server/page.tsx
```diff
@@ -94,6 +94,7 @@ function renderHead(
 
     return `
         ${renderToString(meta)}
+        <link rel="stylesheet" types="text/css" href="/fontSize.css"/>
```
JamieB-gu, line 97:
```suggestion
        <link rel="stylesheet" type="text/css" href="/fontSize.css">
```

JamieB-gu (commented):
Looks good to me, couple small tweaks:

- The [MIME type attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-type) is `type` rather than `types`
- You can omit the closing slash on the `link`- it's always empty (has no children)

JamieB-gu (approved):

PR #708, opened 2020-09-11
Live:594 - Upgraded Keyline components to Lines with styling

dskamiotis:
## Why are you doing this?

Currently Keylines are rendered from internally (AR) created components.
We would like to upgrade this to using Lines from source to allow for more consistency across projects.

## Changes

- Installed @guardian - `Lines` dependency in package.json "@guardian/src-ed-lines": "^2.3.0",
- Replaced and upgraded instances `Keyline` tags in articles to `Lines` identifying appropriate line numbers (for example 8 lines in comment articles)
- On comment article ensuring the bottom Line is perfectly aligned with the bottom of the avatar.
- Upgraded the use of Lines in standard, liveblog and advertisement feature articles
- The width of lines reached across the whole display view/browser and not centred and sized to `contentWidth`.
- Added styling to each of Lines parent div<> in order to achieve this.
- The styling for liveblog needed different CSS parameters in order to centre appropriately- this was influence in part by the Keyline styles for consistency.
- Deleted the Keyline component and test file as well as the storybook instance as now redundant.
**Note: the spacing of Lines as opposed to the Keylines is maybe a more.**

## Screenshots

| Before | After |
| --- | --- |
|  |  |


dskamiotis:
@jacobwinch would be great to be able to test this on different article types

src/components/comment/article.tsx
```diff
@@ -54,6 +54,14 @@ const topBorder = css`
     `}
 `;
 
+const lineStyles = css`
+    margin-top: 5.2rem;
```
webb04, line 58: What does this do? is it the same height as the cutout (image in front of the opinion lines)? Maybe they could share a variable?

webb04 (commented):
Good work so far! nice job cleaning up the old Keyline files.
How do they look in dark mode?

src/components/comment/article.tsx
dskamiotis: This is for the positioning of the Lines component in relation to top of document, in order to line up with bottom of image.
Apparently the Cutout component is constructed from an Img component, of which the parent is a <div>.
The style for it is:
`position: absolute;
    top: -48px;`
It is moving 48px above the parent. Since my Lines component is in a separate <div> tag below I dont think i can share any parameters in the style variables. Unless perhaps I place the Lines component inside and part of the Cutout component? This may make the article/comment more difficult to deconstruct in future? Placing in a span didn't seem to work with any styles, Thinking out loud.

dskamiotis:
@webb04 @JamieB-gu @frankie297 @SiAdcock 
Dark mode for `Lines` to be implemented as a new ticket separately once discussion with Simon resolved on his return

src/components/comment/article.tsx
dskamiotis: - Look into supporting dark mode for Lines component with @SiAdcock

dskamiotis:
@JamieB-gu @webb04 @frankie297 - I have applied a simple solution to dark mode for the Lines component after discussion with Simon Adcock. In the long run it would be better to have a meeting and potentially use dark mode within source, however at this point it is a long term plan.

webb04:
The PR deployment isn't showing, do you need to merge master first?

src/components/standard/article.tsx
```diff
@@ -78,6 +78,25 @@ interface Props {
     children: ReactNode[];
 }
 
+const lineStyles = css`
+    ${from.wide} {
+        width: ${wideContentWidth}px;
+        margin-left: auto;
+        margin-right: auto;
+    }
+    div {
```
webb04, line 87: These styles appear in three different places, would it be a good idea defining it once and importing where they're used?

dskamiotis: I have now placed all in one class in the styles.tsx
Note: I have kept the style customised for liveblog articles as they are different

webb04: Nice, that'll be easier to delete when source adds dark mode support

webb04 (approved):

PR #714, opened 2020-09-14
Don't Ask CAPI For 'fontSize.css`

JamieB-gu:
## Why are you doing this?

In local development we were requesting `/fontSize.css` from CAPI. This is only a relevant resource when running in the Android client (see #703). We should serve a `404` for this from our dev server, as we do for the favicon.

## Changes

- Serve 404 for Android `fontSize.css` stylesheet

### DIFF
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -267,6 +267,7 @@ app.get('/:articleId(*)/live-blocks', liveBlocks);
 app.get('/healthcheck', (_req, res) => res.send("Ok"));
 
 app.get('/favicon.ico', (_, res) => res.status(404).end());
+app.get('/fontSize.css', (_, res) => res.status(404).end());
 
 app.get('/*', bodyParser.raw(), serveArticle);
 


webb04 (approved):
Are iOS clients now requesting a file that doesn't exist?

JamieB-gu:
> Are iOS clients now requesting a file that doesn't exist?

Yes, see the conversation on #346.

src/server/server.ts
```diff
@@ -267,6 +267,7 @@ app.get('/:articleId(*)/live-blocks', liveBlocks);
 app.get('/healthcheck', (_req, res) => res.send("Ok"));
 
 app.get('/favicon.ico', (_, res) => res.status(404).end());
+app.get('/fontSize.css', (_, res) => res.status(404).end());
```
maxspencer, line 270: This is fine, but we could use `501 Not Implemented` instead here? Practical upshot is the same, but the semantics are nicer (and it's fun to use less-used response codes :man_shrugging:)

```suggestion
app.get('/fontSize.css', (_, res) => res.status(501).end());
```

maxspencer (approved):
With optional suggestion :+1:

src/server/server.ts
JamieB-gu: Unfortunately servers are **required** to support `GET`, and therefore **must not** return 501 for `GET` requests, as detailed [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501).

maxspencer: Ah nevermind!

PR #715, opened 2020-09-14
resize interactives

webb04:
## Why are you doing this?
The bottom of interactives can be cut off. They are currently set to 500px in height.

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/client/article.ts
+++ b/src/client/article.ts
@@ -214,6 +214,45 @@ function hasSeenCards(): void {
     })
 }
 
+function safeParse(item: {} | string): any {
+    try {
+        const str = typeof item === 'string' ? item : JSON.stringify(item);
+        return JSON.parse(str);
+    } catch(e) {
+        return {};
+    }
+}
+
+function iframeHeightChanges() {
+    const figures = Array.from(document.querySelectorAll('figure.interactive'));
+    figures.forEach(figure => {
+        const serverRenderedIframe = figure.querySelector('iframe');
+        if (!serverRenderedIframe) return;
+        const iframe = document.createElement('iframe');
+        iframe.style.width = '100%';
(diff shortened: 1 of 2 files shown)


src/client/article.ts
```diff
@@ -214,6 +214,45 @@ function hasSeenCards(): void {
     })
 }
 
+function safeParse(item: {} | string): any {
```
JamieB-gu, line 217: It looks like most of the validation errors can be traced back to this `any` - I think we can get around this by changing it to `unknown`?

src/client/article.ts
```diff
@@ -214,6 +214,45 @@ function hasSeenCards(): void {
...
+
+            switch (message.type) {
+                case 'set-height':
+                    iframe.height = message.value;
+                    break;
+                default:
+            }
+        }, false);
+
+        figure.replaceChild(iframe, serverRenderedIframe);
```
JamieB-gu, line 252: What made you opt to replace the server-rendered `iframe` instead of attaching listeners for messages from it?

JamieB-gu (commented):
Looks good, couple comments.

src/client/article.ts
webb04: The parent window receives messages via `postMessage` in the iframe to `set-height` but I couldn't work out how to match that request to a particular iframe.

We have the `origin` available in every request to update the height, but we would have a problem if we have two interactives from the same origin. The only solution I could find was to replace them all client side - this is how the templates currently handle them.

src/client/article.ts
webb04: Updated this PR, I thought we had handled this before...

JamieB-gu (approved):
Nice! 🎉

PR #731, opened 2020-09-18
Live 527 migrate enzyme to test renderer

dskamiotis:
## Why are you doing this?

- Currently testing with Enzyme adds a layer of complexity to testing in terms of its API
- Particularly when testing components for CSS with `CSS Emotion`
- Integrates well with Jest 
- Jest configured incorrectly and not working with `CSS Emotion`
- Tags are tested using Enzyme and as a result overly complex particularly when having to search the HTML for a string

## Changes

- Configure and set up AR to use `test-renderer` 
- `Tags.test.tsx` and `Keyline.tests.tsx` upgraded and refactored using test-renderer library, deprecating the Enzyme library

paired with @JamieB-gu


JamieB-gu:
Just to elaborate on a couple of the above points for anyone interested:

> Jest configured incorrectly and not working with CSS Emotion

As far as I can tell, Emotion wasn't previously working within the Jest tests. I'm not sure why this didn't cause us any problems - possibly because we were able to analyse props instead of the output? Anyway, if you're interested in the fix it's encapsulated within the second commit on this PR: c4f7bd3.

> Currently testing with Enzyme adds a layer of complexity to testing in terms of its API

> Tags are tested using Enzyme and as a result overly complex particularly when having to search the HTML for a string

To clarify, I don't think there are any major problems with Enzyme. But `react-test-renderer` is [part of the React project](https://reactjs.org/docs/test-renderer.html), seems a bit simpler to use, and is actually [used by Enzyme](https://github.com/enzymejs/enzyme/blob/c4b0a996307054023994b95cc7bbf00cc27a2086/packages/enzyme-adapter-react-16/package.json#L48) under the hood.

The [documentation for `jest-emotion`](https://emotion.sh/docs/jest-emotion) also focuses on `react-test-renderer`, although it does also offer support for Enzyme. For context, we've introduced `jest-emotion` in this PR because it provides some convenient features for working with Jest and Emotion, including some nice [custom matchers](https://emotion.sh/docs/jest-emotion#custom-matchers) and [snapshotting capabilities](https://emotion.sh/docs/jest-emotion#snapshot-serializer).

PR #735, opened 2020-09-21
Consolidate Jest Config

JamieB-gu:
## Why are you doing this?

Thought it would be nice to consolidate our Jest config in a single file within the `config` directory; this takes it out of `package.json`. Also deleted `babel.config.js` and migrated it to the same Jest config file; it was only used for Jest so I decided to put it into a single place to avoid confusion - our main Babel config is in our `webpack.config.ts` file.

## Changes

- Extracted Jest config from `package.json` to `jest.config.js` file
- Moved Babel Jest config into `jest.config.js` from `babel.config.js`


dskamiotis (approved):
Have tested these config changes several times and are working just fine 👍

PR #766, opened 2020-10-01
poll insert epics

webb04:
## Why are you doing this?
Resolves a race condition issue where sometimes epics haven't been fetched natively.

### DIFF
--- a/src/components/standard/article.tsx
+++ b/src/components/standard/article.tsx
@@ -21,6 +21,7 @@ import { remSpace } from '@guardian/src-foundations';
 import RelatedContent from 'components/shared/relatedContent';
 import ImmersiveCaption from 'components/immersiveCaption';
 import { Lines } from "@guardian/src-ed-lines";
+import Epic from 'components/shared/epic';
 
 // ----- Styles ----- //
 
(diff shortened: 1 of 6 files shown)


JamieB-gu (approved):
Nice use of `Promise.all`!

src/components/standard/article.tsx
```diff
@@ -82,8 +83,15 @@ interface Props {
 const Standard: FC<Props> = ({ item, children }) => {
     // client side code won't render an Epic if there's an element with this id
     const epicContainer = item.shouldHideReaderRevenue
-        ? <div id="epic-container"></div>
-        : null
+        ? null
+        : <div id="epic-placeholder">
+            <Epic
```
JamieB-gu, line 88: Ah, I think this may be causing us some errors on the server. The epic code imports and uses the Bridget comms layer, which doesn't seem to work server-side due to reliance on `window` amongst other things...

webb04: Did you see this in the logs? it seems to have deployed ok. I was seeing these problems but adding a `try catch` around the window had resolved it locally.

Ideally we wouldn't want to render an Epic component server side, but I added it so that the styles would be added to the CSP.

JamieB-gu: It's spitting out several errors when running locally for me.

Does the native comms code run anything as a side effect on import? If not we could pass a flag to the `Epic` component to tell it to avoid calling that code when running on the server.

PR #781, opened 2020-10-05
Audio atoms

webb04:
Support audio atoms

## Screenshots

| src-foundations 2.4.0 | src-foundations 1.5.0 |
| --- | --- |
|  |  |


src/server/csp.ts
```diff
@@ -67,12 +67,13 @@ const assetHashes = (assets: string[]): string =>
 
 const buildCsp = ({ styles, scripts }: Assets, twitter: boolean): string => `
     default-src 'self';
-    style-src 'self' ${assetHashes(styles)} https://interactive.guim.co.uk ${twitter ? 'https://platform.twitter.com' : ''};
+    style-src 'self' 'unsafe-inline';
```
JamieB-gu, line 70: I see you've flagged this in the description already, but just a reminder that we probably want to change this before merging.

src/atoms.ts
```diff
@@ -246,6 +246,28 @@ function parseAtom(
...
+            const atom = atoms.audios?.find(audio => audio.id === id);
+
+            if (atom?.data?.kind !== "audio") {
+                return err(`No atom matched this id: ${id}`);
+            }
+            const { id: audioId, title } = atom;
+            const { kicker, trackUrl } = atom?.data?.audio;
+
+            if (!title) {
+                return err(`No videoId for atom: ${id}`);
```
JamieB-gu, line 259:
```suggestion
                return err(`No title for audio atom with id: ${audioId}`);
```

src/client/article.ts
```diff
@@ -222,6 +224,21 @@ function hasSeenCards(): void {
...
+    Array.from(document.querySelectorAll('.js-audio-atom'))
+        .forEach(atom => {
+            const id = atom.getAttribute('id');
+            const trackUrl = atom.getAttribute('trackurl');
+            const kicker = atom.getAttribute('kicker');
+            const title = atom.getAttribute('title');
+            const pillar = Pillar.Culture;
+
+            if (id && trackUrl && kicker && title && pillar) {
+                ReactDOM.hydrate(h(AudioAtom, { id, trackUrl, kicker, title, pillar }), atom)
```
JamieB-gu, line 237: Is there any way we could render audio atoms server-side @shtukas ?

src/renderer.ts
```diff
@@ -595,6 +595,11 @@ const render = (format: Format, excludeStyles = false) =>
             const figcaption = h(FigCaption, { format, caption, credit: none });
             return styledH('figure', figureAttributes, [ styledH('div', attributes), figcaption ]);
         }
+
+        case ElementKind.AudioAtom: {
+            const { pillar } = format;
+            return h('div', { ...element, pillar: pillar.valueOf(), className: 'js-audio-atom audio-atom' }, h(AudioAtom, { ...element, pillar }))
```
JamieB-gu, line 601: If `...element` and `pillar` exist to add attributes to the `<div>` could we use [`data-*`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*) attributes instead?

What do `valueOf` and the `audio-atom` class do here?

src/server/page.tsx
```diff
@@ -71,6 +71,10 @@ const styles = (format: Format): string => `
         line-height: 1.5;
     }
 
+    .audio-atom figure {
+        margin: 0;
+    }
```
JamieB-gu, line 76: Can we apply this directly in the `renderer` using Emotion?

JamieB-gu (commented):
Looks good, a few comments/questions.

src/renderer.ts
webb04: Passing `Pillar.Sport` doesn't seem to work as no pillar styles are added (The screenshot above is missing the blue sport colours).

`.valueOf` isn't working either actually, I need to take another look at this.

webb04: Had to downgrade our src-foundations library to match dotcom in order to see the pillar colours`"@guardian/src-foundations": "^1.5.0"`

cc @shtukas

package.json
```diff
@@ -42,12 +42,12 @@
     "@guardian/src-brand": "^2.4.0",
     "@guardian/src-button": "^2.4.0",
     "@guardian/src-ed-lines": "^2.4.0",
-    "@guardian/src-foundations": "^2.4.0",
+    "@guardian/src-foundations": "^1.5.0",
     "@guardian/src-icons": "^2.4.0",
-    "@guardian/src-radio": "^2.4.0",
+    "@guardian/src-radio": "^1.5.0",
```
JamieB-gu, line 47: Would it be better to fix this upstream in `atoms-rendering` rather than downgrade our version of Source?

webb04: That could be possible. I also had to downgrade src-foundations to 2.0.0 on the discussion-rendering branch to be compatible, so there might be more work involved in updating a few shared libraries

src/renderer.ts
webb04: Upgraded to atoms-rendering 2.0.0 which fixes the problem

src/renderer.ts
```diff
@@ -628,6 +628,17 @@ const render = (format: Format, excludeStyles = false) =>
                 });
                 return styledH('figure', figureAttributes, [ styledH('div', attributes), figcaption ]);
             }
+
+            case ElementKind.AudioAtom: {
+                const { theme } = format;
+                const pillar = themeFromString('pillar/' + themeToPillar(theme));
```
JamieB-gu, line 634: I'm a little puzzled by this line... there seem to be a couple of conversions back and forth?

JamieB-gu (approved):
Looks good!

PR #788, opened 2020-10-06
only stop rendering interactives with no content

webb04:
## Why are you doing this?
Some interactives may not have any html so won't be rendered like http://localhost:8080/world/2020/apr/29/revealed-the-inside-story-of-uk-covid-19-coronavirus-crisis.

This change wouldn't fix this example, as it's attempting to control parts of the page which it shouldn't do on an article of `type: article`.

There might be valid cases for generating interactives using just css or javascript though.

### DIFF
--- a/src/atoms.ts
+++ b/src/atoms.ts
@@ -36,8 +36,8 @@ function parseAtom(
 
             const { html, css, mainJS: js } = atom?.data?.interactive;
 
-            if (!html || !css) {
-                return err(`No html or css for atom: ${id}`);
+            if (!html && !css && !js) {
+                return err(`No content for atom: ${id}`);
             }
 
             return ok({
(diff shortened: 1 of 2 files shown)


src/atoms.ts
```diff
@@ -36,8 +36,8 @@ function parseAtom(
 
             const { html, css, mainJS: js } = atom?.data?.interactive;
 
-            if (!html || !css) {
-                return err(`No html or css for atom: ${id}`);
+            if (!html && !css && !js) {
```
JamieB-gu, line 39: So *all three* fields have to be missing to prevent a render? If for example it _just_ had CSS it would still try to render?

webb04: I think it's unlikely people would use just css but maybe not impossible https://a.singlediv.com/

JamieB-gu (approved):

PR #789, opened 2020-10-06
csp third party spike

dskamiotis:
## Why are you doing this?

This is a spike to investigate to allow `Instagram`, `Youtube` and `Spotify` URLs on the page where there are blocks.
We only allow content from those domains when we know it appears in that particular article.

## Changes

- Included asset hashing for Insta and Youtube
- Included functions to render the hashed UrL to the head in `page.tsx`
- Included functions in `capi.tsx` to check the content type for the element types (video and audio)

Tested with https://mobile.guardianapis.com/rendered-items/culture/2020/oct/05/why-is-dreams-by-fleetwood-mac-back-in-the-charts-and-why-is-everyone-skateboarding-along-to-it

Result:
As a Result of these changes no apparent breakages or blocking of third party content.


src/renderer.test.tsx
```diff
@@ -0,0 +1,447 @@
...
+       const transformed = transformHref(href);
+       expect(transformed).toBe("https://www.theguardian.com/world/series/coronavirus-live")
+    });
+
+    test('Does not transform valid link', () => {
+        const href = "https://www.theguardian.com/world/series/coronavirus-live";
+        const transformed = transformHref(href);
+        expect(transformed).toBe("https://www.theguardian.com/world/series/coronavirus-live")
+     });
+})
```
frankie297, line 447: It might be a good idea to write some tests to validate what you have done so far?

src/server/csp.ts
```diff
@@ -65,24 +65,24 @@ const assetHashes = (assets: string[]): string =>
 //     child-src https: blob:
 // `.trim()
 
-const buildCsp = ({ styles, scripts }: Assets, twitter: boolean): string => `
+const buildCsp = ({ styles, scripts }: Assets, twitter: boolean, instagram: boolean, youtube: boolean, spotify: boolean): string => `
```
frankie297, line 68: We should think if there is a better way to do this rather than passing loads of parameters into the function, @JamieB-gu any ideas?

src/capi.ts
```diff
@@ -102,6 +102,42 @@ const includesTweets = (content: Content): boolean => {
         .some(Boolean)
 }
 
+const includesInstagram = (content: Content): boolean => {
```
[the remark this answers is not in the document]
webb04: I was thinking it could be a generic function with an argument of the element type? `ElementType.INSTAGRAM`

src/renderer.test.tsx
```diff
@@ -0,0 +1,447 @@
+import { renderAll, renderAllWithoutStyles, renderStandfirstText, renderText, transformHref } from 'renderer';
```
webb04, line 1: Is this a new test file? I thought we already had a `renderer.test.tsx` file?

frankie297:
I know we discussed this IRL but we should think about how we add in TikTok, we might need generate the type in the element type enum if we wanted to add in TikTok.

src/renderer.test.tsx
JamieB-gu: I think this may be some changes committed accidentally from #795?

src/server/csp.ts
```diff
@@ -65,24 +65,30 @@ const assetHashes = (assets: string[]): string =>
...
-    frame-src https://www.theguardian.com https://www.scribd.com https://www.instagram.com https://www.tiktok.com https://interactive.guim.co.uk https://open.spotify.com https://www.youtube-nocookie.com https://player.vimeo.com/ ${twitter ? 'https://platform.twitter.com https://syndication.twitter.com https://twitter.com' : ''};
+    script-src 'self' ${assetHashes(scripts)} ${instagram ? 'http://www.instagram.com/embed.js' : ''} https://interactive.guim.co.uk https://s16.tiktokcdn.com https://www.tiktok.com/embed.js https://sf16-scmcdn-sg.ibytedtos.com/ ${twitter ? 'https://platform.twitter.com https://cdn.syndication.twimg.com' : ''};
+    frame-src https://www.theguardian.com https://www.scribd.com ${instagram ? 'https://www.instagram.com' : ''} https://www.tiktok.com https://interactive.guim.co.uk ${spotify ? 'https://open.spotify.com' : ''} ${youtube ? 'https://www.youtube-nocookie.com' : ''} https://player.vimeo.com/ ${twitter ? 'https://platform.twitter.com https://syndication.twitter.com https://twitter.com' : ''};
```
JamieB-gu, line 73: This is nice!

src/capi.ts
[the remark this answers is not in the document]
JamieB-gu: To avoid iterating through the list multiple times you could do something like this:

```ts
type ThirdPartyEmbeds = {
  twitter: boolean;
  youtube: boolean;
  instagram: boolean;
  spotify: boolean;
};

const noThirdPartyEmbeds: ThirdPartyEmbeds = {
  twitter: false;
  youtube: false;
  instagram: false;
  spotify: false;
};

const checkForThirdPartyEmbed = (thirdPartyEmbeds: ThirdPartyEmbeds, element: BodyElement): ThirdPartyEmbeds => {

  switch (element.type) {

    case ElementType.INSTAGRAM:
        return { ...thirdPartyEmbeds, instagram: true };

    case ElementType.TWEET:
        return { ...thirdPartyEmbeds, twitter: true };

    case ElementType.VIDEO:
        return { ...thirdPartyEmbeds, youtube: true };

    case ElementType.AUDIO:
        return { ...thirdPartyEmbeds, spotify: true };

    default:
      return thirdPartEmbeds;

}

body.reduce(
  (thirdPartyEmbeds, block) => block.elements.reduce(checkForThirdPartyEmbed, thirdPartyEmbeds),
  noThirdPartyEmbeds,
);
```

with `checkForThirdPartyEmbed` playing the role of the `switch...case`/function @frankie297 and @webb04 mention above. Haven't tested that but I think it should work 🤔.

The `ThirdPartyEmbeds` object can then be used to reduce the number of arguments to the `buildCsp` function, as @frankie297 [mentions below](https://github.com/guardian/apps-rendering/pull/789/files#r500271885).

JamieB-gu (commented):
Some great stuff here!

src/capi.ts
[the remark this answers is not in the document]
dskamiotis: Wow thats a great implementation, never thought of reduce as a solution here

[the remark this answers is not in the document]
dskamiotis: Thanks Jamie going to implement all these, maybe in incremental pushes in some cases, as they are quite involved👍

src/renderer.test.tsx
dskamiotis: yes thats right, bad merge, renderer.test.tsx changed back to .ts

src/renderer.test.tsx
dskamiotis: @frankie297 a little stuck here on how to test using transformHref, any pointers?

.husky/_/husky.sh
```diff
@@ -0,0 +1,30 @@
...
+  sh -e "$0" "$@"
+  exitCode="$?"
+
+  if [ $exitCode != 0 ]; then
+    echo "husky - $hook_name hook exited with code $exitCode (error)"
+    exit $exitCode
+  fi
+
+  exit 0
+fi
```
JamieB-gu, line 30: Did you mean to commit this file?

src/server/csp.test.ts
```diff
@@ -0,0 +1,55 @@
+import { ElementType } from '@guardian/content-api-models/v1/elementType';
+import { getThirdPartyEmbeds } from 'capi'
+import { articleContentWith } from 'item.test';
```
JamieB-gu, line 3: By `import`ing the `item.test.ts` file here, will all the tests in that file also run (because they're side-effects)? If so, won't that mean that we end up running those tests twice? Once when Jest finds and runs `item.test.ts`, and once when it runs `csp.test.ts`.

src/server/csp.test.ts
```diff
@@ -0,0 +1,55 @@
...
+        id: "id",
+        html: "<blockquote>spotify<blockquote>"
+    }})
+
+describe('Third party embeds function returns true when element type is found in body block', () => {
+    test('returns true when twitter embed found', () => {
+        const newContentTwitter = getThirdPartyEmbeds(contentTwitter)
+        expect(newContentTwitter.twitter).toBe(true);
+    });
+    test('returns true when twitter embed found', () => {
```
JamieB-gu, line 42:
```suggestion
    test('returns true when instagram embed found', () => {
```

src/server/csp.test.ts
```diff
@@ -0,0 +1,55 @@
...
+        const newContentInstagram = getThirdPartyEmbeds(contentInstagram)
+        expect(newContentInstagram.instagram).toBe(true);
+    });
+    test('returns true when twitter embed found', () => {
```
JamieB-gu, line 46:
```suggestion
    test('returns true when youtube embed found', () => {
```

src/server/csp.test.ts
```diff
@@ -0,0 +1,55 @@
...
+        const newContentYoutube = getThirdPartyEmbeds(contentYoutube)
+        expect(newContentYoutube.youtube).toBe(true);
+    });
+    test('returns true when twitter embed found', () => {
```
JamieB-gu, line 50:
```suggestion
    test('returns true when spotify embed found', () => {
```

src/server/csp.ts
```diff
@@ -91,4 +95,5 @@ function csp(item: Item, additionalAssets: Assets, twitter: boolean): string {
 export {
     csp,
     assetHashes,
+    buildCsp,
```
JamieB-gu, line 98: Does this need to be exported?

JamieB-gu (commented):
Some great stuff here! A few comments/suggestions.

src/server/csp.test.ts
dskamiotis: @JamieB-gu Im not entirely sure whether or how this will run twice? i can suggest to directly place the function from item.test.ts into the cspt.test file to ensure no cross over/side effects?

dskamiotis: Just looked through again, I would have thought that it would merely call the standalone function once for the item tests and once for the csp.
The function would merely return a `body`. unless I'm over simplifying here?

JamieB-gu: We've discussed this offline, but to briefly summarise here for posterity:

1. Jest discovers and runs `csp.test.ts`. `csp.test.ts` `import`s `item.test.ts`. `item.test.ts` contains a number of tests. When `csp.test.ts` imports `item.test.ts`, those tests run as side effects of the `import`. That's the first time they run.

2. Jest then discovers and runs `item.test.ts`. `item.test.ts` contains a number of tests that are run. That's the second time they run.

src/capi.ts
```diff
@@ -90,16 +90,49 @@ const articleMainMedia = (content: Content, context: Context): Option<MainMedia>
...
 
-    return body
-        .flatMap(block => block.elements.some(element => element.type === ElementType.TWEET))
-        .some(Boolean)
+const getThirdPartyEmbeds = (content: Content): ThirdPartyEmbeds => {
+    const body = content?.blocks?.body;
+    if (!body) {
+        return noThirdPartyEmbeds
+    }
+    const allThirdPartyEmbeds = body.reduce(
```
webb04, line 130: You could also return the output of `body.reduce`?

.husky/_/husky.sh
webb04: Will +1 after this 👍

src/server/csp.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const contentSpotify = articleContentWith({
+    type: ElementType.AUDIO,
+    assets: [],
+    tweetTypeData: {
+        id: "id",
+        html: "<blockquote>spotify<blockquote>"
+    }})
+
+describe('Third party embeds function returns true when element type is found in body block', () => {
```
webb04, line 37: Good tests 🎉

webb04 (approved):

PR #754, opened 2020-09-28
Testing: do not review

jacobwinch:
## Changes

- Change 1
- Change 2

## Screenshots

| Before | After |
| --- | --- |
|  |  |

PR #796, opened 2020-10-07
Add screenshots workflow

jacobwinch:
## Why are you doing this?

This PR adds functionality for automatically capturing screenshots from a running version of each pull-request branch and comparing them with screenshots captured from the main branch. It builds upon the work started in https://github.com/guardian/apps-rendering/pull/684, which allows us to run the application within the GitHub Actions infrastructure.

During my [original experiment](https://github.com/guardian/apps-rendering/pull/754) I was planning to replace the PR deployment workflow with this new one, but I noticed that a few people are [already using PR deployments](https://github.com/guardian/apps-rendering/pulls?q=is%3Apr+label%3A%22PR+Deployment%22+) in their current form (🎉). Perhaps it is useful to keep both as they have slightly different use-cases. 

https://theguardian.atlassian.net/browse/LIVE-605

## Changes

- Add new workflow file

## Screenshots

See below 🤞

### DIFF
--- a/.github/workflows/screenshots.yml
+++ b/.github/workflows/screenshots.yml
@@ -1,0 +1,79 @@
+name: Screenshots
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main    
+
+jobs:
+  screenshots:
+    
+    runs-on: ubuntu-latest
+    
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - uses: actions/setup-node@v1
+      with:
+        node-version: 12
+    - uses: actions/cache@v2
+      with:
+        path: ~/.npm
+        key: node-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: node-    
+    - name: install
+      run: npm ci
+    - name: run
+      run: |
+        npm run watch & 
+        sleep 60s
+      env: 
+        APIS_IMG_SALT: ${{ secrets.APIS_IMG_SALT }}  
+        CAPI_KEY: ${{ secrets.CAPI_KEY }}
+    - name: Take screenshot of apps-rendering
+      uses: "lannonbr/puppeteer-screenshot-action@1.3.0"
+      with:
+        url: http://localhost:8080/
+    - name: Upload screenshot
+      run: aws s3 cp screenshots/screenshot*.png s3://gu-mobile-pr-images/${{ github.sha }}/screenshot.png --acl public-read
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.CI_IMAGE_UPLOADER_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.CI_IMAGE_UPLOADER_SECRET_ACCESS_KEY }}
+    - name: Prepare image urls
+      id: image-urls
+      if: ${{ github.ref != 'main' }}
+      run: |
+        MAIN_COMMIT_SHA=$(git show-ref main --hash)
+        echo "::set-output name=before::https://gu-mobile-pr-images.s3-eu-west-1.amazonaws.com/$MAIN_COMMIT_SHA/screenshot.png"
+        echo "::set-output name=after::https://gu-mobile-pr-images.s3-eu-west-1.amazonaws.com/${{ github.sha }}/screenshot.png"
+    - name: Search for previous screenshot comment
+      uses: peter-evans/find-comment@v1
+      id: fc
+      with:
+        issue-number: ${{ github.event.pull_request.number }}
+        body-includes: <!-- github-actions-screenshots -->
+    - name: Create comment if necessary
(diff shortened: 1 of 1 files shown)


.github/workflows/screenshots.yml
```diff
@@ -0,0 +1,79 @@
+name: Screenshots
+
+on:
+  push:
```
jacobwinch, line 4: This allows us to collect a new baseline (or 'Before' screenshot) each time a new change is merged to `main`.

jacobwinch: Obviously we'll only know that works correctly for sure once this PR is merged!

.github/workflows/screenshots.yml
```diff
@@ -0,0 +1,79 @@
...
+    branches:
+      - main    
+
+jobs:
+  screenshots:
+    
+    runs-on: ubuntu-latest
+    
+    steps:
+    - uses: actions/checkout@v2
```
jacobwinch, line 17: A lot of this code is duplicated between several workflow files, which is a pain. Unfortunately GitHub Actions doesn't provide a good way to share these steps between multiple workflows yet. See https://github.com/actions/runner/issues/646 for more details.

.github/workflows/screenshots.yml
```diff
@@ -0,0 +1,79 @@
...
+      run: |
+        npm run watch & 
+        sleep 60s
+      env: 
+        APIS_IMG_SALT: ${{ secrets.APIS_IMG_SALT }}  
+        CAPI_KEY: ${{ secrets.CAPI_KEY }}
+    - name: Take screenshot of apps-rendering
+      uses: "lannonbr/puppeteer-screenshot-action@1.3.0"
+      with:
+        url: http://localhost:8080/
```
jacobwinch, line 40: Obviously we could capture a more interesting collection of screenshots here!

.github/workflows/screenshots.yml
```diff
@@ -0,0 +1,79 @@
...
+    - name: Upload screenshot
+      run: aws s3 cp screenshots/screenshot*.png s3://gu-mobile-pr-images/${{ github.sha }}/screenshot.png --acl public-read
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.CI_IMAGE_UPLOADER_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.CI_IMAGE_UPLOADER_SECRET_ACCESS_KEY }}
+    - name: Prepare image urls
```
jacobwinch, line 46: This is all getting a bit ugly - we could write our own [Javascript action](https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/about-actions#types-of-actions) (and store it within this repo) if this bothers people too much.

JamieB-gu: This sounds like a nice solution longer term, if we'd like to port this functionality to other projects.

.github/workflows/screenshots.yml
```diff
@@ -0,0 +1,79 @@
...
+      id: image-urls
+      if: ${{ github.ref != 'main' }}
+      run: |
+        MAIN_COMMIT_SHA=$(git show-ref main --hash)
+        echo "::set-output name=before::https://gu-mobile-pr-images.s3-eu-west-1.amazonaws.com/$MAIN_COMMIT_SHA/screenshot.png"
+        echo "::set-output name=after::https://gu-mobile-pr-images.s3-eu-west-1.amazonaws.com/${{ github.sha }}/screenshot.png"
```
JamieB-gu, line 52: Are `::set-output` and `name=before::` special GH Actions syntax?

.github/workflows/screenshots.yml
JamieB-gu: Do you mean a different page, or a list of pages?

.github/workflows/screenshots.yml
JamieB-gu: If we wrote an action to combine these steps would that work?

.github/workflows/screenshots.yml
JamieB-gu: So if the branch is `main` this runs up to and including the "Upload screenshot" step, but doesn't add the comments?

JamieB-gu (approved):
Awesome! 🚀

.github/workflows/screenshots.yml
jacobwinch: I guess taking screenshots of several different articles would help us to catch more problems?

.github/workflows/screenshots.yml
jacobwinch: There is a thing called a [composite action](https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/creating-a-composite-run-steps-action), but it won't allow you to have the `uses` syntax (and reference other actions)

.github/workflows/screenshots.yml
jacobwinch: Yep - here is the [documentation](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-output-parameter).

PR #797, opened 2020-10-07
Don't search for comment on main

jacobwinch:
## Why are you doing this?

I made a silly mistake with https://github.com/guardian/apps-rendering/pull/796.

Don't say I didn't [warn you](https://github.com/guardian/apps-rendering/pull/796#discussion_r500994700) @JamieB-gu!

PR #798, opened 2020-10-07
Attempt to fix screenshots workflow

jacobwinch:
## Why are you doing this?

Because [this change](https://github.com/guardian/apps-rendering/pull/797) didn't work as expected.


webb04 (approved):

JamieB-gu (approved):
Would making a separate GH action for `main` with only the steps needed there be an option?

jacobwinch:
> Would making a separate GH action for `main` with only the steps needed there be an option?

It's definitely an option (and I briefly considered it). The main arguments against are:

1. All of [this code](https://github.com/guardian/apps-rendering/pull/798/files#diff-d6360abd39e109ea92528f570d5e2389R1-L45) would be duplicated.
2. On a related note, if we start capturing multiple urls we'd need to keep them in sync across the two different workflows.

jacobwinch:
[Success](https://github.com/guardian/apps-rendering/actions/runs/293738758) (at last!)

PR #802, opened 2020-10-08
Run validation workflow against main

jacobwinch:
## Why are you doing this?

This should help us to catch conflicts after merging to `main`. See: https://theguardian.atlassian.net/browse/LIVE-352


webb04 (approved):

jacobwinch:
This is [working as desired](https://github.com/guardian/apps-rendering/actions/runs/295700114).

PR #803, opened 2020-10-08
Quiz articles and quiz atoms

webb04:
- Renders quiz articles
- Renders quiz atoms
- Uses different csp for articles with quiz, audio and chart atoms.

This would allow audio atoms to work in this PR https://github.com/guardian/apps-rendering/pull/781
This would enable chart atoms: http://localhost:8080/society/2020/sep/09/almost-half-of-britons-drinking-more-due-to-covid-survey-finds

http://localhost:8080/sport/2020/oct/02/sports-quiz-of-the-week-losers-lavish-spending-lifesaver-football


JamieB-gu:
> I don't think they are any different to standard articles

Interesting, I wasn't aware of that. So the definition of a "quiz" article is any article with a Quiz Atom? If that's true then I agree we probably don't need to differentiate it. @gtrufitt @nitro-marky @shtukas what do you think?

On a side note, my understanding of atoms is that they're meant to represent evergreen content. Is an atom the correct place for a quiz to live, or were they used as a convenient bit of functionality to get quizzes working?

gtrufitt:
> my understanding of atoms is that they're meant to represent evergreen content

https://docs.google.com/document/d/11EJuB_zb8KeUcvGNotFH00zkPPmj7laUWkFgRqvh-qw/edit#

A Content Atom is:

- Embeddable piece of content - does not have a canonical page
- Can be present on multiple pages
- Has a URI as an identifier but not a gu.com url
- Can be placed / embedded on a piece of content / page
- Edited and managed outside of composer
- Owned / edited by a guardian system (i.e. not a youtube video - that’s a conventional embed)

Quiz atoms fall into all of those categories I believe.

JamieB-gu:
Ah thanks @gtrufitt! Good to know (maybe we want to include this write-up in our recommendations repo? cc @SiAdcock @sndrs)

On these two points:

> - Embeddable piece of content - does not have a canonical page
> - Can be present on multiple pages

Is that true of quiz atoms? I thought that quizzes appeared on a single piece of content (so only one page), and the entirety of that page was the quiz (so it's the _canonical_ page for that quiz) - but I may well be wrong on this 😬.

gtrufitt:
> I thought that quizzes appeared on a single piece of content (so only one page), and the entirety of that page was the quiz

Nope, quizzes are embedded atoms so could be on any number of pages. It's just that we happen to create articles that are just quizzes. But, for example, you could have a World Cup quiz that was on several articles

src/atoms.ts
```diff
@@ -246,6 +246,35 @@ function parseAtom(
...
+            }
+
+            const questions = content.questions.map(question => {
+                return {
+                    text: question.questionText,
+                    ...question,
+                    answers: question.answers.map(answer => {
+                        return {
+                            text: answer.answerText,
+                            isCorrect: !!answer.weight,
```
JamieB-gu, line 268: `isCorrect` is derived from weighting?

src/client/article.ts
```diff
@@ -222,6 +223,21 @@ function hasSeenCards(): void {
     })
 }
 
+function hydrateQuizAtoms(): void {
+    Array.from(document.querySelectorAll('.js-quiz'))
+        .forEach(atom => {
+            const props = atom.querySelector('.js-quiz-params')?.innerHTML;
+            try {
+                if (props) {
+                    const quizProps = JSON.parse(props.replace(/&quot;/g, '"'));
```
JamieB-gu, line 232: I think you can deal with this warning by using `unknown` and checking each field.

src/server/csp.ts
```diff
@@ -65,24 +65,30 @@ const assetHashes = (assets: string[]): string =>
 //     child-src https: blob:
 // `.trim()
 
-const buildCsp = ({ styles, scripts }: Assets, twitter: boolean): string => `
+const styleSrc = (styles: string[], twitter: boolean, hasInlineStyles: boolean) => {
+    return hasInlineStyles
+        ? `https: 'unsafe-inline';`
+        : `'self' ${assetHashes(styles)} https://interactive.guim.co.uk ${twitter ? 'https://platform.twitter.com' : ''};`
+}
```
JamieB-gu, line 72: Hmm, this narrows the issue to a subset of articles, but it still drops much of the CSP security for any articles with these embeds right? 🤔

Also, if we were to go with the `unsafe-inline` approach, wouldn't we still want to narrow the external domains? By that I mean we'd want to allow specific domains rather than whitelisting all `https:`?

JamieB-gu (commented):
Looks good in general!

I still have one concern on the CSP front.

src/atoms.ts
webb04: I think so, it seems to check out on the quiz I tried 🤷‍♂️ It sounds like some answers are worth more than others

src/client/article.ts
```diff
@@ -9,10 +9,10 @@ import setup from 'client/setup';
 import Epic from 'components/shared/epic';
 import ReactDOM from 'react-dom';
 import { ads, slideshow, videos, reportNativeElementPositionChanges } from 'client/nativeCommunication';
+import { QuizAtom } from '@guardian/atoms-rendering';
+import { QuizAtomType } from '@guardian/atoms-rendering/dist/QuizAtom';
```
JamieB-gu, line 13: Do you think we should make this type and `QuestionType` top-level exports from `atoms-rendering`? It looks a little bit like we're using a private API here, and it worries me that someone might change it without realising it's being relied on...?

JamieB-gu (approved):
Looks good!

Heads up, whichever of this or #789 gets merged first, the other may have some merge conflicts to untangle.

PR #807, opened 2020-10-09
Enforce 4-Space Indentation

JamieB-gu:
## Why are you doing this?

Added a 4-space indent rule to ESLint. Makes the codebase more consistent and also saves conversations and developer time during PR review.

More details on the rule [here](https://eslint.org/docs/rules/indent).

**Note:** The _only_ thing that should have changed across all files (except `.eslintrc.js` and `package.json`) is indentation. GitHub seems to be having trouble displaying this clearly in the diff in some cases (e.g. `bodyElement.ts`) - it might be worth checking it using another diff tool (for example, VSCode represents this more clearly for me).

## Changes

- Set ESLint 'error' for indent of 4 spaces
- Set indent rule for switch...case
- Added npm script for ESLint auto-fix
- Fixed errors caused by this new rule

### DIFF
--- a/src/components/series.tsx
+++ b/src/components/series.tsx
@@ -42,10 +42,10 @@ const immersiveStyles = ({ kicker }: PillarStyles, isLabs: boolean): SerializedS
 const font = (isLabs: boolean): SerializedStyles => {
     return css`
         ${
-            isLabs
-                ? textSans.medium({ lineHeight: 'loose', fontWeight: 'bold' })
-                : headline.xxxsmall({ lineHeight: 'loose', fontWeight: 'bold' })
-        }`
+    isLabs
+        ? textSans.medium({ lineHeight: 'loose', fontWeight: 'bold' })
+        : headline.xxxsmall({ lineHeight: 'loose', fontWeight: 'bold' })
+}`
 }
 
 const linkStyles = ({ kicker, inverted }: PillarStyles, isLabs: boolean): SerializedStyles => css`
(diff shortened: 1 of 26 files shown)


src/components/series.tsx
```diff
@@ -42,10 +42,10 @@ const immersiveStyles = ({ kicker }: PillarStyles, isLabs: boolean): SerializedS
...
     return css`
         ${
-            isLabs
-                ? textSans.medium({ lineHeight: 'loose', fontWeight: 'bold' })
-                : headline.xxxsmall({ lineHeight: 'loose', fontWeight: 'bold' })
-        }`
+    isLabs
+        ? textSans.medium({ lineHeight: 'loose', fontWeight: 'bold' })
+        : headline.xxxsmall({ lineHeight: 'loose', fontWeight: 'bold' })
+}`
```
webb04, line 48: Is this alignment correct?

src/styles.ts
```diff
@@ -32,7 +32,7 @@ export const darkModeCss = (
         ${styles
         .map((style, i) => `${style}${placeholders[i] ? placeholders[i] : ''}`)
         .join('')
-    }
+}
```
webb04, line 35: This spacing correct?

webb04 (approved):

src/components/series.tsx
JamieB-gu: Yeah... ESLint sometimes seems to get a bit confused by template strings... I've refactored the two areas you flagged, let me know what you think!

webb04 (approved):
LGTM 🚀

PR #821, opened 2020-10-13
Move DOCTYPE And Charset

JamieB-gu:
## Why are you doing this?

Refactor only. Moving the `DOCTYPE` declaration to the `page.tsx` file where we build our HTML document, and moving the `meta[charset]` tag to the `Meta` component.

## Changes

- Moved `DOCTYPE` to `page.tsx`
- Moved `meta[charset]` to `Meta`

### DIFF
--- a/src/components/meta.tsx
+++ b/src/components/meta.tsx
@@ -12,6 +12,7 @@ interface Props {
 
 const Meta: FC<Props> = ({ title, cspString }) =>
     <>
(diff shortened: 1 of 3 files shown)


webb04 (approved):
I remember having some problems about where we set `charset` before but I don't remember the exact issue, maybe worth keeping an eye on. Maybe the problem was just not setting it 😀

PR #836, opened 2020-10-16
include meta description

webb04:
## Why are you doing this?
SEO might not be the most important feature for us, but I still think it's a good idea to have a meta description.

## Screenshots

| Before | After |
| --- | --- |
|  |  |


JamieB-gu (approved):
👍

FYI dotcom appear to use the standfirst for this, but I'm not sure it's worth passing that through for us considering we're not too worried about search and bookmarking.

PR #844, opened 2020-10-19
Image-Rendering

JamieB-gu:
## Why are you doing this?

Swapping out some of our image code to use the `image-rendering` library. The focus of this PR is the `BodyImage` component.

FYI @mxdvl @gtrufitt @SiAdcock @shtukas 

## Changes

- Added `image-rendering` v1.0.0 as dependency
- Handled transition from `Option<Role>` to `Role`
- Updated `image.ts` to use types from `image-rendering`
- Deleted `FigCaption` and moved `renderCaption` into `renderer.ts`
- Used new `BodyImage` in renderer
- Introduced `Credit` component


webb04 (commented):
Code lgtm and the images look high quality 🎉

I noticed at some screen widths we no longer have the right `.5rem` margin:

Screen Shot 2020-10-22 at 12 02 11

JamieB-gu:
> I noticed at some screen widths we no longer have the right `.5rem` margin:

Hmm, good spot. There appears to be something weird going on with scrollbar overlap. If you "hide scrollbars" in macOS settings it looks fine, but if you show them they overlap the image. I'll look into it.

JamieB-gu:
Right @webb04, fixed that with a change in `image-rendering`: guardian/image-rendering#85.

config/jest.config.js
```diff
@@ -39,6 +39,6 @@ module.exports = {
       'src',
     ],
     transformIgnorePatterns: [
-      'node_modules/(?!(@guardian/src-foundations|@guardian/types|@guardian/src-icons)/)',
+      'node_modules/(?!(@guardian/src-foundations|@guardian/types|@guardian/src-icons|@guardian/image-rendering)/)',
```
mxdvl, line 42: This seems like an array already, would it make sense to use it that way instead of an ever-growing regex?

mxdvl (approved):

config/jest.config.js
JamieB-gu: That sound sensible. At the moment this says "ignore `node_modules` except for these four libraries". Is there a way to achieve that with an array?

*Side note* I think ideally we'd say "ignore `node_modules` except for `@guardian/*`".

mxdvl: Ah yes of course! I was reading it the other way around.

I think that your suggestion is [easy to implement](https://regex101.com/r/h1ZXGw/3):

```regex
node_modules/(?!@guardian)
```

PR #850, opened 2020-10-20
Updgrade icon fonts to SVG

dskamiotis:
## Why are you doing this?

References and use of src-fonts are still being used in AR and need to be upgraded.
This is part of a several sub tasks to deprecate its use.

Currently pullquotes are using icons which cause issues when rendering in certain browser and ios/android versions, with strange fallback icons when not loading (square box).

Issues regarding Icons and how SVg are more desirable:

Please see: https://github.com/guardian/apps-rendering/pull/677

## Changes

- Upgraded Blockquote to SVG with customised styles to have consistent positioning and size
- Upgraded read me arrow in renderer to be svg with updated styling
- Eliminated `@font-Face` from code base
- Deleted `icons from 'styles'`
- Eliminated  `font-family: 'Guardian Icons'`

## Screenshots

| Renderer Read More | BlockQuote |
| --- | --- |
|  |  |


src/components/blockquote.tsx
```diff
@@ -16,31 +17,37 @@ interface Props {
...
-    position: relative;
-    margin: ${remSpace[4]} 0 ${remSpace[9]} 0;
-    padding: 0 ${remSpace[6]};
+const styles = (format: Format): SerializedStyles => {
+    const { kicker } = getPillarStyles(format.pillar);
+    return css`
+        font-style: italic;
+        position: relative;
+        margin: ${remSpace[4]} 0 ${remSpace[9]} 0;
+        padding: 0 ${remSpace[5]};
```
webb04, line 26: I think there was another element that used `padding: 0 ${remSpace[6]};` so the blockquote lined up

src/components/blockquote.tsx
```diff
@@ -16,31 +17,37 @@ interface Props {
...
+
+        svg {
```
webb04, line 28: Could we use the same space formatting used in this file?

webb04 (commented):
Good work! just a couple of small things

src/components/blockquote.tsx
dskamiotis: correct I was playing around with that and forgot to change back

src/components/blockquote.tsx
dskamiotis: 👍

src/components/blockquote.tsx
```diff
@@ -16,31 +17,37 @@ interface Props {
...
+                height: 62.4px;
+                margin-left: -9.2625px;
+                margin-bottom: -12px;
+                margin-top: 4.4px;
```
JamieB-gu, line 32: Regarding the use of pixel values here, what happens if you scale the page up and down (`cmd +` and `cmd -` in a browser)?

src/renderer.ts
```diff
@@ -317,19 +318,16 @@ const richLinkStyles = (format: Format): SerializedStyles => {
...
-        button::before {
-            ${icons}
-            content: '\\e005';
+        svg{
+            width: 1.0625rem;
             border-radius: 100%;
             border: solid 1px ${neutral[7]};
-            font-size: 12px;
-            padding: 3px 6px 4px 6px;
+            padding: 4px;
```
JamieB-gu, line 327: Same question about pixels as above.

JamieB-gu (commented):
Nice!

Couple questions about scaling.

webb04 (approved):

PR #852, opened 2020-10-20
Upgrade @guardian/types to use Theme

webb04:
## Why are you doing this?
- Updates `@guardian/types` to `0.5.2`
- Rename pillar helper functions to `theme`
- Removed `featureHeadline` and `soft` colours that aren't being used
- This should let us start work on special reports

### DIFF
--- a/src/components/media/byline.tsx
+++ b/src/components/media/byline.tsx
@@ -3,7 +3,6 @@
 import React, { ReactNode, FC } from 'react';
 import { css, SerializedStyles } from '@emotion/core';
 import { neutral } from '@guardian/src-foundations/palette';
-import { Pillar } from '@guardian/types/Format';
 import { Option, map, withDefault } from '@guardian/types/option';
 import { Item, getFormat } from 'item';
 import { textSans } from "@guardian/src-foundations/typography";
@@ -39,13 +38,12 @@ const styles: SerializedStyles = css`
 // ----- Component ----- //
 
 interface Props {
-    pillar: Pillar;
     publicationDate: Option<Date>;
     className: SerializedStyles;
     item: Item;
 }
 
-const Byline: FC<Props> = ({ pillar, publicationDate, className, item }) => {
+const Byline: FC<Props> = ({ publicationDate, className, item }) => {
     const byline = pipe2(
         item.bylineHtml,
         map(html => <address>{ renderText(html, getFormat(item)) }</address>),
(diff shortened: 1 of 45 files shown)


frankie297 (approved):
Do we have somewhere where we can add a definition of what is the difference between `Theme` and `Pillar`?

webb04:
> Do we have somewhere where we can add a definition of what is the difference between `Theme` and `Pillar`?

Not sure if we have any documentation around this @JamieB-gu ?

The types repo defines it here:
https://github.com/guardian/types/blob/main/Format.ts#L3-L15

We have the problem where the special report colours need to override pillar colours on special report articles.

It might be confusing to call them a `pillar` as we don't treat them as pillars in some places like CAPI and the navigation bar on dotcom for example. So we updated `Format` to use a `Theme` which can be a `Pillar` or something `Special`

src/components/media/byline.tsx
```diff
@@ -39,13 +38,12 @@ const styles: SerializedStyles = css`
 // ----- Component ----- //
 
 interface Props {
-    pillar: Pillar;
     publicationDate: Option<Date>;
     className: SerializedStyles;
     item: Item;
 }
 
-const Byline: FC<Props> = ({ pillar, publicationDate, className, item }) => {
```
JamieB-gu, line 48: Interesting that this didn't previously throw an unused argument warning 🤔

src/pillarStyles.ts
```diff
@@ -60,10 +60,18 @@ export const pillarColours: PillarColours = {
...
+        liveblogBackground: palette.lifestyle[300],
+        liveblogDarkBackground: palette.lifestyle[200]
     }
 }
 
-const getPillarStyles = (pillar: Pillar): PillarStyles => pillarColours[pillar];
+const getPillarStyles = (pillar: Theme): PillarStyles => pillarColours[pillar];
 
 function pillarFromString(pillar: string | undefined): Pillar {
     switch (pillar) {
```
JamieB-gu, line 77: Some of the names in this file might be a bit confusing now? For example, `PillarColours` where the key is `theme`?

JamieB-gu (approved):
Herculean refactor here, 💪

JamieB-gu:
> Do we have somewhere where we can add a definition of what is the difference between `Theme` and `Pillar`?

> Not sure if we have any documentation around this

It's a good point. We should probably put some markdown docs into the `@guardian/types` repo.

JamieB-gu:
@webb04 you may find the conversation [here](https://github.com/guardian/types/issues/4) useful for future work on this.

PR #853, opened 2020-10-20
Full width Epic

webb04:
the width of the epics is a bit too small so there's more space to the right compared to the left. It should take up the whole width

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/client/article.ts
+++ b/src/client/article.ts
@@ -103,7 +103,7 @@ function formatDates(): void {
 function insertEpic(): void {
     const epicPlaceholder = document.getElementById('epic-placeholder');
     if (epicPlaceholder) {
-        // epicPlaceholder.innerHTML = "";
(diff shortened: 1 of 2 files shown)


frankie297:
Sorry, not sure what has changed here?

webb04:
The screenshot on mobile might be a bit better, the width of the epics is a bit too small so there's more space to the right compared to the left. It should take up the whole width

IMG_6449

frankie297:
@webb04 aha thanks for that! yep gotcha! 👍 💯

JamieB-gu:
> the width of the epics is a bit too small so there's more space to the right compared to the left. It should take up the whole width

Do you want to drop this quote into the main PR description for future reference? 😄

JamieB-gu (approved):

PR #874, opened 2020-10-26
added function for android presence check

dskamiotis:
## Why are you doing this?

At the moment two footers are being rendered on the Android platform, the SS as well as the native.
Max has asked to disable the footer component for Android
So we need the footer to still render for iOS by default

## Changes

- created a function to read whether the platform is Android and as a result run the isCcpa function which renders the ccpa link 

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3914,9 +3914,9 @@
       "requires": {
         "@emotion/core": "^10.0.35",
         "@guardian/src-foundations": "^2.5.0",
-        "@guardian/types": "github:guardian/types#da18ca13c822c9eed91ec9b6722c344c6f8088b4",
-        "react": "^17.0.0",
-        "react-dom": "^17.0.0",
+        "@guardian/types": "github:guardian/types#semver:^0.5.2",
+        "react": "^16.14.0",
+        "react-dom": "^16.14.0",
         "typescript": "^4.0.3"
       },
       "dependencies": {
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -3914,9 +3914,9 @@
       "requires": {
         "@emotion/core": "^10.0.35",
         "@guardian/src-foundations": "^2.5.0",
-        "@guardian/types": "github:guardian/types#da18ca13c822c9eed91ec9b6722c344c6f8088b4",
-        "react": "^17.0.0",
-        "react-dom": "^17.0.0",
+        "@guardian/types": "github:guardian/types#semver:^0.5.2",
+        "react": "^16.14.0",
+        "react-dom": "^16.14.0",
```
JamieB-gu, line 3919: Why has this changed? 🤔

src/client/article.ts
```diff
@@ -123,7 +125,15 @@ function insertEpic(): void {
...
 }
 
-
+function footerInit(): void {
+    const isAndroid = /(android)/i.test(navigator.userAgent);
+    const footer = document.getElementById('articleFooter')
+    if (footer && isAndroid){
+        ReactDOM.unmountComponentAtNode(footer)
+    } else {
+        isCCPA();
```
JamieB-gu, line 131: Should this be called something like `renderFooter`? `isCCPA` sounds like it should return a boolean.

dskamiotis: isCCPA was part of the previous footer commit, so we can change this once updating the footer design

webb04 (approved):

PR #875, opened 2020-10-26
Editions

JamieB-gu:
## Why are you doing this?

Initial support for rendering Editions articles. This introduces a query param `?editions` that's only available when running locally. It also includes some of the UI work for supporting a "standard" Editions article.

## Changes

- Added a `maybeRender` helper function for `Option`s
- Added a new Editions article
- Added Editions `Byline`, `HeaderImage`, `Headline`, `Series` and `Standfirst`
- Added an `editionsPage` module for building the HTML document
- Updated our server to accept an `editions` query param in dev

## Screenshots

**Note:** The camera icon on the App screenshots is an existing bug on Firefox - doesn't appear on Chrome and Safari.

| App | Editions |
| - | - |
| app | editions |

### DIFF
--- a/src/components/editions/article.tsx
+++ b/src/components/editions/article.tsx
@@ -1,0 +1,62 @@
+// ----- Imports ----- //
+
+import React, { FC } from 'react';
+import { css } from '@emotion/core';
+import { border } from '@guardian/src-foundations/palette';
+import { remSpace } from '@guardian/src-foundations';
+import { Lines } from '@guardian/src-ed-lines';
+import { Design } from '@guardian/types/Format';
+import { partition } from '@guardian/types/result';
+
+import { Item } from 'item';
+import { renderAll } from 'renderer';
+import Standfirst from 'components/editions/standfirst';
+import HeaderImage from 'components/editions/headerImage';
+import Series from 'components/editions/series';
+import Headline from 'components/editions/headline';
+import Byline from 'components/editions/byline';
+
+
+// ----- Component ----- //
+
+interface Props {
+    item: Item;
+}
+
+const headerStyles = css`
+    margin: 0 ${remSpace[3]} ${remSpace[4]};
+`;
+
+const bodyStyles = css`
+    border-top: 1px solid ${border.secondary};
+    padding: 0 ${remSpace[4]};
+`;
+
+const Article: FC<Props> = ({ item }) => {
+    if (item.design === Design.Live) {
+        return <p>Not implemented</p>
+    }
+    
+    return (
+        <main>
+            <article>
+                <header css={headerStyles}>
+                    <HeaderImage item={item} />
+                    <Series item={item} />
+                    <Headline item={item} />
+                    <Standfirst item={item} />
+                    <Lines />
+                    <Byline item={item} />
+                </header>
+                <section css={bodyStyles}>
+                    {renderAll(item, partition(item.body).oks)}
+                </section>
+            </article>
+        </main>
+    );
+}
+
+
+// ----- Exports ----- //
+
+export default Article;
--- a/src/components/editions/article.tsx
+++ b/src/components/editions/article.tsx
@@ -36,7 +36,7 @@ const Article: FC<Props> = ({ item }) => {
     if (item.design === Design.Live) {
         return <p>Not implemented</p>
     }
-    
+
     return (
         <main>
             <article>
--- a/src/components/editions/series.tsx
+++ b/src/components/editions/series.tsx
@@ -27,7 +27,7 @@ const Series: FC<Props> = ({ item }) =>
     maybeRender(item.series, series =>
         <nav css={styles}>
             {series.webTitle}
-        </nav>                            
+        </nav>
     )
 
 
--- a/src/components/editions/series.tsx
+++ b/src/components/editions/series.tsx
@@ -15,7 +15,7 @@ import { maybeRender } from 'lib';
 const styles = css`
     ${titlepiece.small()}
     color: ${news[400]};
-    font-size: 17px;
+    font-size: 1.0625rem;
     padding: ${remSpace[1]} 0 ${remSpace[2]};
 `;
 
--- a/src/components/editions/series.tsx
+++ b/src/components/editions/series.tsx
@@ -1,0 +1,36 @@
+// ----- Imports ----- //
+
+import React, { FC } from 'react';
+import { css } from '@emotion/core';
+import { titlepiece } from '@guardian/src-foundations/typography';
+import { news } from '@guardian/src-foundations/palette';
+import { remSpace } from '@guardian/src-foundations';
+
+import { Item } from 'item';
+import { maybeRender } from 'lib';
+
+
+// ----- Component ----- //
+
+const styles = css`
+    ${titlepiece.small()}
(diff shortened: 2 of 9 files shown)


src/server/editionsPage.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+
+// ----- Functions ----- //
+
+const renderHead = (request: RenderingRequest): string =>
+    renderToString(
+        <>
+            <meta charSet="utf-8" />
+            <title>{request.content.webTitle}</title>
+            <meta name="viewport" content="initial-scale=1" />
```
webb04, line 34: I think we set a max scale on apps-rendering to avoid users zooming in, could this be an issue on editions?

src/server/server.ts
```diff
@@ -131,6 +137,7 @@ async function serveArticlePost(
 async function serveArticleGet(req: Request, res: ExpressResponse): Promise<void> {
     try {
         const articleId = req.params[0] || defaultId;
+        const isEditions = req.query.editions === '';
```
webb04, line 140: `req.query?.editions === ''` ?

src/components/editions/series.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Item } from 'item';
+import { maybeRender } from 'lib';
+
+
+// ----- Component ----- //
+
+const styles = css`
+    ${titlepiece.small()}
+    color: ${news[400]};
+    font-size: 17px;
```
webb04, line 18: No rems?

src/lib.ts
```diff
@@ -39,6 +45,13 @@ function errorToString(error: unknown, fallback: string): string {
 const isObject = (a: unknown): a is Record<string, unknown> =>
     typeof a === 'object' && a !== null;
 
+const maybeRender = <A>(oa: Option<A>, f: (a: A) => ReactElement | null): ReactElement | null =>
```
webb04, line 48: Nice, we can use this in a few other places

webb04 (commented):
The articles are looking good! There's some components in `renderer.ts` that are turning dark in dark mode. Maybe we could add a small addition to `darkModeCss` while adding a class to the body of the editions article?

`
body:not(.editions) {
    @media (prefers-color-scheme: dark) {}
}
`

src/server/editionsPage.tsx
SiAdcock: Out of interest, why do we prevent users zooming?

webb04: Good question, It's been like that for a while in the templates, I'm not sure when the original decision was made but maybe preventing zoom made the articles feel more "native"?

We have the font resize feature on apps which works nicely defining everything in rem. It doesn't work for all cases though e.g. cartoons with text 

First article is templates and second is apps-rendering
80618203-ef8db800-8a3a-11ea-9cac-be5937ed7ccc

JamieB-gu: I was going to ask this too actually. The `maximum-scale` limit was flagged by Lighthouse as an accessibility problem so I didn't include it here. I have two questions on this:

1. Why do we include it at the moment in Apps-Rendering?
2. Does it make a difference considering we allow users to zoom everything on the page with the slider? Does it respect the system zoom level?

src/server/server.ts
JamieB-gu: `req.query` always exists doesn't it?

src/components/editions/series.tsx
JamieB-gu: Good point. Sorry, this is a temporary override of the fonts allowed in Source (cc @SiAdcock) because Titlepiece 17 is not currently allowed. However, you're right, it should be in `rem`s, I'll fix that.

JamieB-gu:
> The articles are looking good! There's some components in `renderer.ts` that are turning dark in dark mode.

Ah, good point. What's the story with Editions and dark mode @groakland @pradasa @prisalcalde? Is it supported and, if so, do we have designs for it?

groakland:
> > There's some components in `renderer.ts` that are turning dark in dark mode.
> 
> Ah, good point. What's the story with Editions and dark mode @groakland @pradasa @prisalcalde? Is it supported and, if so, do we have designs for it?

Dark mode is not currently supported in Editions

groakland (approved):
Ok this looks like a great starter for ten and provides the Editions team a template of how ER Article types can be added to apps-rendering. 

The question now becomes - how we can make this a real endpoint in CODE/PROD so the Editions Archiver can start consuming these new ER Articles with the **editions** query parameter - perhaps this is for a subsequent PR?

src/components/editions/article.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+const Article: FC<Props> = ({ item }) => {
+    if (item.design === Design.Live) {
+        return <p>Not implemented</p>
+    }
+
+    return (
+        <main>
+            <article>
+                <header css={headerStyles}>
```
jimhunty, line 43: Wouldn't it be better to use `emotion.styled` as you already seperate out the css and turn it into a component? More stylistic than anything

src/components/editions/article.tsx
```diff
@@ -0,0 +1,62 @@
...
+import HeaderImage from 'components/editions/headerImage';
+import Series from 'components/editions/series';
+import Headline from 'components/editions/headline';
+import Byline from 'components/editions/byline';
+
+
+// ----- Component ----- //
+
+interface Props {
+    item: Item;
```
jimhunty, line 23: Item seems a strange name but assume its something legacy?

jimhunty (approved):
All looks good to me. A few stylistic things. 

There is a lack of storybook stories and tests which would help going forward.

src/components/editions/article.tsx
JamieB-gu: We could do. Bear in mind that the `css` prop is used everywhere else in this codebase so it would be a divergence from that, but also happy to consider switching if you think there are advantages to `styled`. What do you reckon it gives us over the `css` prop?

src/components/editions/article.tsx
JamieB-gu: It's not legacy, it's probably the most important type in the codebase 😄. It refers to a CAPI "item" - you can see the same terminology used, for example, [in MAPI](https://github.com/guardian/mobile-apps-api/tree/master/mobile-item).

JamieB-gu:
> There is a lack of storybook stories and tests which would help going forward.

Yup - we'll start to add these over time. This branch is just the initial spike that we'll build upon.

src/components/editions/article.tsx
```diff
@@ -0,0 +1,62 @@
+// ----- Imports ----- //
```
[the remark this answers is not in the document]
JamieB-gu: This component handles an Editions "article".

src/lib.ts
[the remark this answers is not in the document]
JamieB-gu: This line is the function signature, the 5 lines below it are the implementation.

I should probably add a JSDoc comment to this. In summary, a common pattern for us it to conditionally render some UI based upon whether a prop is `Some` or `None. This function takes care of that so we don't end up writing the same `None` checks over and over.

PR #882, opened 2020-10-27
HorizontalRule Component

JamieB-gu:
## Why are you doing this?

Migrated `HorizontalRule` to standalone component. I'm trying to narrow the focus of the `renderer` file, and also introduce stories for components that don't have them.

## Changes

- Moved `HorizontalRule` to standalone file
- Added stories

### DIFF
--- a/src/components/horizontalRule.tsx
+++ b/src/components/horizontalRule.tsx
@@ -1,0 +1,32 @@
+// ----- Imports ----- //
+
+import React, { FC } from 'react';
+import { css } from '@emotion/core';
+import { neutral } from '@guardian/src-foundations/palette';
+import { remSpace } from '@guardian/src-foundations';
+
+import { darkModeCss } from 'styles';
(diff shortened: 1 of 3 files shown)


webb04 (approved):

frankie297 (approved):

src/components/horizontalRule.tsx
```diff
@@ -0,0 +1,32 @@
...
+    border: 0;
+    background-color: ${neutral[93]};
+
+    ${darkModeCss`
+        background-color: ${neutral[20]};
+    `}
+`;
+
+const HorizontalRule: FC = () =>
+    <hr css={styles} />
```
dskamiotis, line 27: So this will be a re usable component throughout AR? Sounds like a good idea. 👍 
I like the way the style is a part of the inner element

dskamiotis (approved):

PR #888, opened 2020-10-28
apply new lint rules

webb04:
## Why are you doing this?
Use the shared guardian eslint rules to keep consistency in the project.
Makes use of prettier, which can do most of the work when formatting files `Cmd + Shift + P; format document` in VSCode

There's also an option to format documents on save https://www.robinwieruch.de/how-to-use-prettier-vscode

Most notable changes:
- `"` instead of `'`
- use `import type`
- use semi colons
- use `Array<string>` over `string[]`
- css selectors start on new lines
- removes optional chaining on types that don't require it
- alphabetically sorted imports
- need to give reasons where eslint-disable comments are used

This is a huge PR that will conflict with most open branches, I'm happy to wait for open PRs to be merged, rebase this PR and then release.


src/ads.test.ts
```diff
@@ -30,101 +29,122 @@ const renderParagraphs = compose(render, generateParas);
...
-            '<div></div>',
-            '<div></div>',
-            '<div></div>',
-            '<div></div>',
+            "<p></p>",
+            "<p></p>",
+            "<div></div>",
+            "<div></div>",
+            "<div></div>",
+            "<div></div>",
```
dskamiotis, line 126: Im assuming you had to change what was in the codebase before applying the new linting rules for ' > "

src/server/ssmConfig.ts
```diff
@@ -1,31 +1,36 @@
-import { ssm } from './aws';
-import { App, Stack, Stage } from './appIdentity';
-import { Option, some, none, map, withDefault } from '@guardian/types/option';
-import { pipe2 } from 'lib';
+import type { Option } from "@guardian/types/option";
```
dskamiotis, line 1: I assume this makes the Option type available throughout the code base without exporting? Is this like a d.ts file?

src/server/page.tsx
```diff
@@ -1,27 +1,30 @@
...
-import { csp } from 'server/csp';
-import { pageFonts } from 'styles';
-import Meta from 'components/meta';
-
+import { CacheProvider } from "@emotion/core";
+import type { RenderingRequest } from "@guardian/apps-rendering-api-models/renderingRequest";
+import { background } from "@guardian/src-foundations/palette";
+import { Design, Display } from "@guardian/types/Format";
+import type { Format } from "@guardian/types/Format";
+import type { Option } from "@guardian/types/option";
```
dskamiotis, line 8: This is importing from the same place ("@guardian/types/Format") however broken down separately into new lines. Is this just for the linter or for efficiency?

dskamiotis (approved):

config/.eslintrc.js
```diff
@@ -84,6 +86,9 @@ module.exports = {
         '@typescript-eslint/explicit-function-return-type': 'error',
         // This check seems to be flaky, and complains about things that TS is happy about
         'react/prop-types': 'off',
+        'import/no-unresolved': 'off',
+        'indent': 'off',
+        'import/no-default-export': 'off'
```
JamieB-gu, line 91: FYI @sndrs

src/ads.tsx
```diff
@@ -1,16 +1,17 @@
-import { createElement as h, ReactNode, ReactElement, isValidElement } from 'react';
-import Paragraph from 'components/paragraph';
-import { Button, buttonBrandAlt } from '@guardian/src-button';
-import React from 'react';
-import { ThemeProvider } from 'emotion-theming'
+import { Button, buttonBrandAlt } from "@guardian/src-button";
+import Paragraph from "components/paragraph";
+import { ThemeProvider } from "emotion-theming";
+import type { ReactElement, ReactNode } from "react";
+import React, { createElement as h, isValidElement } from "react";
```
JamieB-gu, line 5: Does it sort imports alphabetically?

src/atoms.ts
```diff
@@ -1,12 +1,14 @@
...
-import { DocParser } from "types/parserContext";
-import Int64 from 'node-int64';
-import { isValidDate } from "date";
+import type { Atoms } from "@guardian/content-api-models/v1/atoms";
+import type { BlockElement } from "@guardian/content-api-models/v1/blockElement";
+import { fromNullable } from "@guardian/types/option";
+import type { Result } from "@guardian/types/result";
+import { err, ok } from "@guardian/types/result";
+import type { BodyElement } from "bodyElement";
+import { ElementKind } from "bodyElement";
```
JamieB-gu, line 7: This is interesting. So it's smart enough to figure out that `interface`s can be made into `type` imports, but `enum`s need to remain normal imports so that the lookups work? @sndrs

For context, `BodyElement` is an `interface` and `ElementKind` is an `enum`.

src/atoms.ts
```diff
@@ -18,23 +20,25 @@ function formatOptionalDate(date: Int64 | undefined): string | undefined {
 function parseAtom(
     element: BlockElement,
     atoms: Atoms,
-    docParser: DocParser,
+    docParser: DocParser
```
JamieB-gu, line 23: No trailing commas can result in a larger diff when changes are made - have we intentionally decided that we're fine with that? @sndrs 

This is interesting because a few lines below it's _added_ a trailing comma to an Object declaration 🤔

src/atoms.ts
```diff
@@ -255,45 +279,45 @@ function parseAtom(
...
+            const atom = atoms.quizzes?.find((quiz) => quiz.id === id);
+            if (atom?.data.kind !== "quiz" || !id) {
                 return err(`No atom matched this id: ${id}`);
             }
 
-            const { content } = atom?.data?.quiz;
+            const { content } = atom.data.quiz;
 
-            if (!content) {
+            if (content.questions.length === 0) {
```
JamieB-gu, line 294: @webb04 did you make this change or did it do it automatically?

JamieB-gu (commented):
I'm going to have to review this in chunks due to its size. Made it as far as `atoms.ts` this time. Looks very handy. Few comments questions (some for @sndrs).

src/atoms.ts
webb04: I think this was one of the things `eslint --fix` couldn't automatically do so I had to do manually to fix the errors. It would convert to `import type` for all imports of a module, which wasn't working.

src/atoms.ts
webb04: I made this one, it required it to be a more specific check

src/ads.tsx
webb04: Yes, which makes this PR huge 😀

src/ads.test.ts
webb04: `eslint --fix` was able to convert all the single quotes into `"`

src/server/ssmConfig.ts
webb04: This is a recommended new feature in TypeScript https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html

src/server/page.tsx
webb04: I'm not sure if there's a better way to handle this, but the following wouldn't work because not all of the imports are types

``` import type { Design, Display, Format } from "@guardian/types/Format"; ```

Comment (sndrs) [reply src/server/page.tsx:8]: yeah this is the rule that requires type-only imports to be explicit

File: src/atoms.ts (hunk 7 above)
Comment (sndrs) [reply src/atoms.ts:23]: hmmm htat doesn't seem right, agree trailing comma's should be there. wil lhave a look

File: config/.eslintrc.js (hunk 4 above)
Comment (sndrs) [reply config/.eslintrc.js:91]: i think this is ok – requiring named exports makes it easier to trace them through codebases (same name everywhere unless you deliberately rename the import). it's fair enough if a project wants to override anything, but it proved a useful setting in frontend, prob a good default?

File: config/jest.config.js [hunk 9]
```diff
@@ -1,44 +1,32 @@
 module.exports = {
     collectCoverage: true,
-    preset: 'ts-jest/presets/js-with-ts',
-    rootDir: '../',
+    preset: "ts-jest/presets/js-with-ts",
```
sndrs, line 3: if you use https://www.npmjs.com/package/@guardian/prettier too these will all revert to `'`

src/ads.test.ts
sndrs: if you use https://www.npmjs.com/package/@guardian/prettier too these will all revert to `'`

src/ads.tsx
sndrs: but only once :)

src/atoms.ts
sndrs: yeah enums become actual JS don't they?

src/atoms.ts
sndrs: mentioned above, i think you need to use https://www.npmjs.com/package/@guardian/prettier too

src/atoms.ts
JamieB-gu: Yeah, they output code, whereas something like an `interface` or a `type` only exists at compile time. It's interesting that the automated changes would have broken this 🤔

src/atoms.ts
webb04: Thanks, I've added `@guardian/prettier` in this commit https://github.com/guardian/apps-rendering/pull/888/commits/a51475c64db5fa74e6e4930cf1faace99f3540f1

src/atoms.ts
sndrs: would they? i thought it convert pure type imports to `import type` but left enum as a normal import? did i misdunderstand?

JamieB-gu: > It would convert to `import type` for all imports of a module, which wasn't working.

Based on what @webb04 said above, yes? I believe this means that ESLint thought that _all_ type-related imports would become `import type`, including `enum`s, so he had to manually convert the ones that were `enum`s back?

webb04: I can't remember if they were `enum`s or normal imports, but the eslint fix command did try to use `import type` for some modules that were not types. I think if one type was exported from a file, all the other exports were grouped as a type

PR #908, opened 2020-11-03
Live 1018 update design footer

dskamiotis:
## Why are you doing this?

Updated Footer

## Changes

- Change 1
- Change 2

## Screenshots

| Default | wWith Ccpa |
| --- | --- |
|  |  |


src/client/article.ts
```diff
@@ -137,7 +137,7 @@ function footerInit(): void {
 
 function isCCPA(): void {
     userClient.doesCcpaApply().then(isOptedIn => {
-        const comp = h(FooterCcpa, {isCcpa: isOptedIn});
```
webb04, line 140: To be updated

src/components/shared/footer.tsx
```diff
@@ -1,16 +1,17 @@
...
 
 // Footer content styles
 const container = css`
-    border-style: solid;
-    border-color: ${brandBorder.primary};
     border-width: 0 1px;
+    ${textSans.small({ lineHeight: "regular" })};
+    margin-left: 10px;
+    margin-right: 10px;
+    margin-bottom: 10px;
```
webb04, line 14: The left/right margin for the main article body is `.5rem` which is remSpace[2] in the design system I think

src/components/shared/footer.tsx
```diff
@@ -82,6 +83,23 @@ interface FooterCcpaProps {
     isCcpa: boolean;
 }
 
-const FooterCcpa: FC<FooterCcpaProps> = ({ isCcpa }) => <Footer>{footerContents(isCcpa)}</Footer>
+const renderContent = (ccpaStatus: boolean): JSX.Element => {
+    if (!ccpaStatus){
```
webb04, line 87: Is this either/or? or do we need to show both?

src/components/shared/footer.tsx
```diff
@@ -82,6 +83,23 @@ interface FooterCcpaProps {
...
+        return <a css={anchor} href="https://www.theguardian.com/help/privacy-policy">Privacy Settings</a>
+    } else {
+        return <a css={anchor} href="https://www.theguardian.com/help/privacy-policy">California Residents - Do Not Sell</a>
```
webb04, line 90: What does the "Do Not Sell" mean?

src/components/shared/footer.tsx
webb04: It's also below, so could we see this twice?

src/components/shared/footer.tsx
```diff
@@ -82,6 +83,23 @@ interface FooterCcpaProps {
...
+    }
+}
+
+
+const FooterCcpa: FC<FooterCcpaProps> = ({ isCcpa }) => {
+    return <div css={container}>
+        &#169;
+        2020 Guardian News and Media Limited or its affiliated companies. All rights reserved.<br/>
```
webb04, line 98: Can we make this 2020 dynamic to avoid us a job in January? 😀

src/components/shared/footer.tsx
```diff
@@ -1,16 +1,17 @@
 import { css } from '@emotion/core';
-import { Footer } from '@guardian/src-footer';
```
webb04, line 2: Is there more code we can remove? like the package.json and package-lock.json

src/components/shared/footer.tsx
dskamiotis: yes, that was the older version should have been ommited

src/components/shared/footer.tsx
dskamiotis: Not quite sure :) it was on the design system text content

src/components/shared/footer.tsx
dskamiotis: new date () added 👍

src/components/shared/footer.tsx
dskamiotis: do we use `remspace` for the top and bottom margins also?

JamieB-gu: Ideally we always use `rem`s unless there's a specific reason to use pixels, otherwise the UI will break when scaled.

In this case I think we should also try to match this to the closest spacing value available in the design system, which is `remSpace[2]` (equivalent to `8px`) or `remSpace[3]` (equivalent to `12px`).

src/components/shared/footer.tsx
JamieB-gu: I believe it's specific wording related to CCPA - perhaps @petternitter can confirm?

src/components/shared/footer.tsx
```diff
@@ -1,91 +1,68 @@
...
-	border-right-width: 1px;
-`;
+interface FooterCcpaProps {
+	isCcpa: boolean;
+}
 
-const CcpaListItem = (ccpaStatus: bool): JSX.Element | null => {
-	if (ccpaStatus) {
+const renderContent = (ccpaStatus: boolean): JSX.Element => {
+	if (!ccpaStatus) {
```
webb04, line 28: I think it's nicer to have `if (ccpaStatus)` in the boolean checks.

src/components/shared/footer.tsx
```diff
@@ -1,91 +1,68 @@
...
 		return (
-			<li css={[li, link2]}>
-				<a
-					css={anchor}
-					href="https://www.theguardian.com/help/privacy-policy"
-				>
-					California Residents - Do not sell
-				</a>
-			</li>
+			<a
```
webb04, line 30: Do we need both this and the privacy policy link below on line 62?

webb04 (approved):

src/components/shared/footer.tsx
prisalcalde: Yes, it's the California law thing. I raised the fact that users might not know what this is about with @petternitter . It'd be good to review the copy at some point to make it more intuitive.

PR #923, opened 2020-11-05
render match reports as standard articles

webb04:
## Why are you doing this?
Match reports should look like standard articles, render them in the same way as other articles.
Article: `football/2020/nov/03/atalanta-liverpool-champions-league-group-d-match-report`
## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/item.ts
+++ b/src/item.ts
@@ -210,9 +210,6 @@ const isLive =
 const isRecipe =
     hasTag('tone/recipes');
 
-const isMatchReport =
-    hasTag('tone/matchreports');
-
 const isInterview =
     hasTag('tone/interview');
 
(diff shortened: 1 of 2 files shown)


frankie297 (approved):

groakland (approved):
LGTM

src/item.ts
```diff
@@ -283,11 +280,6 @@ const fromCapi = (context: Context) => (request: RenderingRequest): Item => {
             design: Design.Recipe,
             ...itemFieldsWithBody(context, request),
         };
-    } else if (isMatchReport(tags)) {
-        return {
-            design: Design.MatchReport,
```
JamieB-gu, line 288: Should we be removing this from the `Design` type altogether then?

webb04: Dotcom add stats to the match report page so I think it's useful to have https://www.theguardian.com/football/2020/nov/03/atalanta-liverpool-champions-league-group-d-match-report. At the moment they just look like standard articles on apps

webb04: That looks like a possible `stats-rendering` project

PR #934, opened 2020-11-11
downgrade atoms-rendering

webb04:
## Why are you doing this?
The builds have been failing in TeamCity since this update was merged in https://github.com/guardian/apps-rendering/pull/924

Strange how we got the green ticks on that PR in GitHub actions but failures in TeamCity. Could it be the node module caching we do in GitHub actions? @JamieB-gu @jacobwinch 

## Changes

- Downgrade `atoms-rendering from 2.0.9 to 2.0.7 
- Update `image-rendering`


dskamiotis (approved):

JamieB-gu:
> on that PR in GitHub actions but failures in TeamCity

That's odd... Are there any obvious differences between the two environments? What was the error in TC?

PR #929, opened 2020-11-10
Pillar rich links

webb04:
## Why are you doing this?

- Enhance rich links client side
- Match the designs on dotcom
- Give readers more information about rich links (e.g. what pillar/topic)
- Make them more appealing to drive onward content

## Changes

- Accept new `richlink` query parameter to serve article details (pillar and image) and without html document, scripts and styles
- Request article details for rich links client side
- New `darkModeStyles` function that returns a string of dark mode styles instead of a `SerializedStyles` object
- Move `RichLink` component into separate file
- Add `RichLink` story

## Other considerations

Creating a new endpoint saves us having to send large unused html, css and js. It does split the cache though and will put extra load on MAPI and apps-rendering.

## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |

PR #943, opened 2020-11-12
fetch from rendered item endpoint

webb04:
## Why are you doing this?
Additional changes to enable https://github.com/guardian/apps-rendering/pull/929

### DIFF
--- a/src/client/article.ts
+++ b/src/client/article.ts
@@ -329,15 +329,14 @@ function richLinks(): void {
 			if (articleId) {
 				const options = {
 					headers: {
-						Accept: 'text/html',
+						Accept: 'application/json',
 					},
 				};
-				void fetch(`${articleId}?richlink`, options).then(
-					(response) => {
-						const pillar = response.headers
-							.get('pillar')
-							?.toLowerCase();
-						const image = response.headers.get('image');
+				void fetch(`${articleId}?richlink`, options)
+					.then((resp) => resp.json())
+					.then((response) => {
+						const pillar = response?.pillar?.toLowerCase();
+						const image = response?.image;
 
 						if (pillar) {
 							richLink.classList.add(`js-${pillar}`);
@@ -347,8 +346,7 @@ function richLinks(): void {
 						if (placeholder && image) {
 							placeholder.innerHTML = `<img src="${image}" alt="related article"/>`;
 						}
-					},
-				);
+					});
 			}
 		});
 }
--- a/src/client/article.ts
+++ b/src/client/article.ts
@@ -327,21 +327,28 @@ function richLinks(): void {
 		.forEach((richLink) => {
 			const articleId = richLink.getAttribute('data-article-id');
 			if (articleId) {
-				void fetch(`${articleId}?richlink`).then((response) => {
-					const pillar = response.headers
-						.get('pillar')
-						?.toLowerCase();
-					const image = response.headers.get('image');
+				const options = {
+					headers: {
+						Accept: 'text/html',
+					},
+				};
+				void fetch(`${articleId}?richlink`, options).then(
+					(response) => {
+						const pillar = response.headers
+							.get('pillar')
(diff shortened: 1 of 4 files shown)


src/components/richLink.tsx
```diff
@@ -143,7 +143,7 @@ const RichLink = (props: {
 	const webUrl = 'https://www.theguardian.com';
 
 	const articleId = url.includes(webUrl)
-		? { 'data-article-id': url.replace(webUrl, '') }
+		? { 'data-article-id': url.replace(webUrl, '/rendered-items') }
```
JamieB-gu, line 146: Will this work locally?

webb04: Seems to work with https://github.com/guardian/apps-rendering/pull/943/files#diff-0ec854d4226b5921f1520f7bfae4e2eb27cdedadf904662fade4342f6a57472cR258

src/client/article.ts
```diff
@@ -327,21 +327,28 @@ function richLinks(): void {
...
+							?.toLowerCase();
+						const image = response.headers.get('image');
+
+						if (pillar) {
+							richLink.classList.add(`js-${pillar}`);
+						}
+
+						const placeholder = richLink.querySelector('.js-image');
+						if (placeholder && image) {
+							placeholder.innerHTML = `<img src="${image}" alt="related article"/>`;
```
JamieB-gu, line 348: Do you have an example of what the URL of an image will look like?

src/client/article.ts
```diff
@@ -329,15 +329,14 @@ function richLinks(): void {
...
 				};
-				void fetch(`${articleId}?richlink`, options).then(
-					(response) => {
-						const pillar = response.headers
-							.get('pillar')
-							?.toLowerCase();
-						const image = response.headers.get('image');
+				void fetch(`${articleId}?richlink`, options)
+					.then((resp) => resp.json())
+					.then((response) => {
```
JamieB-gu: This should force you to perform checks on the JSON data, which will in turn sidestep the `any` error.

```suggestion
					.then((response: unknown) => {
```

src/client/article.ts
```diff
@@ -347,8 +346,7 @@ function richLinks(): void {
 						if (placeholder && image) {
 							placeholder.innerHTML = `<img src="${image}" alt="related article"/>`;
 						}
-					},
-				);
+					});
```
JamieB-gu, line 349: We may want a `catch` to avoid crashing the client JS? For example, if the user is offline this request will fail and there won't be any JSON in the response. Perhaps we should check the response code?

src/client/article.ts
JamieB-gu: This was a bit vague, sorry. There are a couple of reasons why I asked this.

1. I wanted to check that it was using the image resizer (`i.guim` rather than `media.guim`). I've done some digging and I think it does because we're pulling it from `mainMedia` on `Item`. 👍 

2. I wanted to check that we're making use of responsive images here. I'm assuming not because this is a straightforward `img[src]` - can we make use of `image-rendering` to include a responsive image?

webb04: Yep it's using the same `src` as the main image. I did think about serving the `srcset` string to make it responsive but used `src` due to the small size of the image and the fixed with of the rich links.

dskamiotis (approved):

src/client/article.ts
JamieB-gu: Ok, makes sense.

If we're not going with React here can we do something like this:

```ts
const img = document.createElement('img');
img.setAttribute('alt', 'Related article');
img.setAttribute('src', image);
placeholder.appendChild(img);
```

It's a bit more verbose but may be:

1) More performant, because there's no need to engage the HTML parser.
2) [More secure](https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#Security_considerations)

JamieB-gu (approved):

PR #945, opened 2020-11-12
Editions CSP

JamieB-gu:
## Why are you doing this?

Added a Content Security Policy for the Editions page. This will restrict the resources, such as styles, images and scripts, that we allow on pages used by the Editions project. For more information on CSPs see [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP). Go [here](https://emotion.sh/docs/ssr#advanced-approach) for more information about the Emotion APIs.

## Changes

- Used Emotion critical to extract styles into single tag
- Hashed style tags in CSP
- Allowed some image domains

### DIFF
--- a/src/server/editionsPage.tsx
+++ b/src/server/editionsPage.tsx
@@ -1,13 +1,20 @@
 // ----- Imports ----- //
 
+import { CacheProvider } from '@emotion/core';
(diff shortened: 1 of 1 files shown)


webb04 (approved):

groakland (commented):
LGTM

PR #951, opened 2020-11-13
add screenshots for multiple article types

webb04:

### DIFF
--- a/.github/workflows/screenshots.yml
+++ b/.github/workflows/screenshots.yml
@@ -52,7 +52,7 @@ jobs:
                   CAPI_KEY: ${{ secrets.CAPI_KEY }}
 
             - name: Take screenshot of apps-rendering
-              uses: 'lannonbr/puppeteer-screenshot-action@1.3.0'
+              uses: 'lannonbr/puppeteer-screenshot-action@1.3.1'
               with:
                   url: 'http://localhost:8080/uk-news/2020/nov/12/stonehenge-road-tunnel-given-go-ahead-despite-backlash'
             - name: Upload screenshot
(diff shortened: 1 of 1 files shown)


.github/workflows/screenshots.yml
```diff
@@ -47,33 +67,125 @@ jobs:
...
+                        'http://localhost:8080/full-of-beanz/2020/jan/30/healthy-eating-seven-small-changes-that-can-make-a-big-difference?Labs',
+                        'http://localhost:8080/fashion/2020/nov/12/rainbow-bright-how-the-symbol-of-optimism-and-joy-spread-across-our-clothes-homes-and-lives-in-2020?Feature',
+                        'http://localhost:8080/news/gallery/2020/nov/12/morning-fog-and-a-dog-in-deep-water-thursdays-best-photos?Gallery',
+                        'http://localhost:8080/commentisfree/2020/nov/12/lee-cain-no-10-boris-johnson-chief-of-staff-britain?Opinion',
+                        'http://localhost:8080/film/2020/nov/12/billie-review-singer-documentary-historical-voice-harassment-exploitation?Review',
+                        'http://localhost:8080/empowering-business/2020/oct/31/why-would-i-ever-go-back-to-shopping-new-how-vintage-has-become-the-future-of-buying-ethically?ImmersiveLab',
+                        'http://localhost:8080/us-news/2020/nov/13/joe-biden-cabinet-selection-trump-obstruction?Analysis',
+                    ]
```
JamieB-gu, line 99: My YAML isn't great, but do you think something like this would work?

```yaml
matrix:
  article:
    - format: Immersive
      url: 'http://localhost:8080/news/2020/nov/12/western-travel-influencers-social-media-pakistan-politics'
    - format: Labs
      url: 'http://localhost:8080/full-of-beanz/2020/jan/30/healthy-eating-seven-small-changes-that-can-make-a-big-difference'
...
```

So instead of the regex and faux query param we could do:

```yaml
${{ matrix.article.format }}
${{ matrix.article.url }}
```

JamieB-gu (commented):
I think this is going to be really useful! Couple of questions/comments. I'd be curious to hear @jacobwinch's opinion as well.

jacobwinch:
The results of this look great - I think it’ll be really useful 💯 

However, I wonder if it is worth rethinking the implementation slightly. For example, it seems a little odd that we need to run the application in two different jobs in order to achieve this result. Similarly, the performance hit we’re taking here concerns me a little (the checks used to take ~2 mins, now they take ~7 mins). Perhaps we should:

1. Checkout
2. Setup and run the application
3. Take screenshots of all of the different article urls that interest us and upload them to s3
4. Build a single comment and post it

If it’s fiddly to do all of this using bash, then we could consider using a [custom Javascript action](https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/about-actions#javascript-actions) to handle some of this work. I appreciate that this suggestion is a substantial change that will take more time, so if you'd like to merge this as is for now then feel free to do that!

.github/workflows/screenshots.yml
```diff
@@ -47,33 +67,125 @@ jobs:
...
-              with:
-                  issue-number: ${{ github.event.pull_request.number }}
-                  body-includes: <!-- github-actions-screenshots -->
+                  echo "::set-output name=before::https://gu-mobile-pr-images.s3-eu-west-1.amazonaws.com/$MAIN_COMMIT_SHA/standard/screenshot.png"
+                  echo "::set-output name=after::https://gu-mobile-pr-images.s3-eu-west-1.amazonaws.com/${{ github.sha }}/standard/screenshot.png"
             - name: Create comment if necessary
-              if: ${{ github.event_name == 'pull_request' && steps.fc.outputs.comment-id == '' }}
+              if: ${{ github.event_name == 'pull_request' }}
               uses: peter-evans/create-or-update-comment@v1
               with:
```
jacobwinch, line 75: If we set `replace` as the `edit-mode` here, then I don't think we'll need the `delete_screenshots` job?

jacobwinch: (Obviously we'd need to reinstate the find comment step too)

.github/workflows/screenshots.yml
webb04: This works! I'll remove the regex

.github/workflows/screenshots.yml
```diff
@@ -47,33 +67,125 @@ jobs:
...
+                        'http://localhost:8080/full-of-beanz/2020/jan/30/healthy-eating-seven-small-changes-that-can-make-a-big-difference?Labs',
+                        'http://localhost:8080/fashion/2020/nov/12/rainbow-bright-how-the-symbol-of-optimism-and-joy-spread-across-our-clothes-homes-and-lives-in-2020?Feature',
+                        'http://localhost:8080/news/gallery/2020/nov/12/morning-fog-and-a-dog-in-deep-water-thursdays-best-photos?Gallery',
+                        'http://localhost:8080/commentisfree/2020/nov/12/lee-cain-no-10-boris-johnson-chief-of-staff-britain?Opinion',
+                        'http://localhost:8080/film/2020/nov/12/billie-review-singer-documentary-historical-voice-harassment-exploitation?Review',
+                        'http://localhost:8080/empowering-business/2020/oct/31/why-would-i-ever-go-back-to-shopping-new-how-vintage-has-become-the-future-of-buying-ethically?ImmersiveLab',
+                        'http://localhost:8080/us-news/2020/nov/13/joe-biden-cabinet-selection-trump-obstruction?Analysis',
+                    ]
```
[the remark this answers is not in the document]
webb04: I can't see a way to do this... you can share data between steps but not jobs (maybe using this action https://github.com/marketplace/actions/persist-data-between-jobs). I've made two separate jobs as the screenshot collection can be run in parallel but the comment posting needs to be synchronous to avoid overwrites and knowing when to create/replace or append.

.github/workflows/screenshots.yml
jacobwinch: Thanks @webb04 - this is much clearer 🎉

.github/workflows/screenshots.yml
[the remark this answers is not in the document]
jacobwinch: One option to consider: we could store this in a separate file (e.g. create a `screenshot_urls.json` in the `.github` directory) and loop over all of the items in the array, instead of using the `matrix` concept at all.

I can see why we wanted to use matrix, but in practice it seems to have a few downsides:

1. Forces us to declare the list of article types and urls twice
1. Forces us to run two separate jobs (which seems unpleasant due to the associated checkout/container setup costs)
1. I think the matrix usage (and separate job) is the reason that this shows up as `22 successful checks`.

jacobwinch (approved):
I've [suggested](https://github.com/guardian/apps-rendering/pull/951#discussion_r526898003) a way that we might be able to simplify this further, but overall this is a big improvement so perhaps we can merge this initial version 👍

PR #959, opened 2020-11-17
update screenshot action to avoid using set env

webb04:
## Why are you doing this?
All the open pull requests are failing builds because of the screenshot action.
I saw a warning message on Friday saying `set-env` would be removed and actions wouldn't run if they used it after the 16th November.

Luckily it appears the action we're using to take screenshots has been updated to use the suggested environment files.
https://github.com/lannonbr/puppeteer-screenshot-action/commit/ecc08ce649f3aa4ff689388f903d38d83eccfc22


JamieB-gu (approved):

jacobwinch:
Thanks for fixing this @webb04 👍

PR #964, opened 2020-11-18
stop calling unimplemented bridget function

webb04:
## Why are you doing this?
This bridget function hasn't been implemented yet in iOS, causing the iOS code to hang inside the bridget functon when debugging.

Commenting this code out for the time being to avoid any possible memory leaks for beta users.


frankie297 (approved):
👍

dskamiotis (approved):

PR #968, opened 2020-11-19
post message to react native when images have been clicked

webb04:
## Why are you doing this?

Post messages to react native when the header image or body images are clicked.

I haven't got the editions project running locally to test but I would be happy to pair with someone on that.

## Changes
- New `editions.ts` client side script
- Adds click event listeners to inform React native layer to open lightbox

## Questions/assumptions
- Does the editions client side JavaScript need to be inlined?
- Are the index values lining up to the images used in the lightbox?
- There's also an `isLightboxEnabled` flag set in remote config, what are the conditions for this?

I'm not sure why I can't add you as a reviewer, maybe we need to add you as a collaborator to apps-rendering? cc @abeddow91

### DIFF
--- a/src/client/editions.ts
+++ b/src/client/editions.ts
@@ -27,7 +27,7 @@ Array.from(body?.querySelectorAll('.js-launch-slideshow') ?? []).forEach(
 			window.ReactNativeWebView.postMessage(
 				JSON.stringify({
 					type: 'openLightbox',
-					index: index + 1,
+					index,
 					isMainImage: 'false',
 				}),
 			);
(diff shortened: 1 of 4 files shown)


groakland:
FYI- The relevant editions code related to lightbox interaction can be found here: https://github.com/guardian/editions/blob/master/projects/Mallard/src/components/article/types/article.tsx#L266

webb04:
This PR has been replaced with the newer https://github.com/guardian/apps-rendering/pull/977. This PR inlines the JavaScript which should make it work straight away when offline.

PR #971, opened 2020-11-20
upgraded to react v17

dskamiotis:
## Why are you doing this?

Upgraded to react v17, this is required for the upgrading to JSX Transform.
Doing this a separate pull to ensure the safe construction of the process.


webb04 (approved):

JamieB-gu:
It looks like this may have caused several warnings from other libraries that require React 16, including a number of the Source libraries and `atoms-rendering`. Have you noticed any compatibility issues from this mismatch?

PR #992, opened 2020-11-30
JSX transform upgrade

dskamiotis:
## Why are you doing this?

Upgrading React to 17.0 in order to stay up to date as well as reducing the need to use the `import react`

## Changes

- Upgraded to React 17
- added `"runtime": "automatic"` to the Presets in webpack config

NOTE:
Still testing but opening this up for any feedback.
I manually deleted imports as the macro code seemed to not allow it to run in watch.
I have noticed that the following files are causing problems:
`logo.tsx`  `headerMedia.tsx`   `article.tsx`  `credit.tsx` `meta.tsx` `scripts.tsx`
@frankie297 @abeddow91 @JamieB-gu  , is this something we should continue with? Doesnt seem to be any other conflicts but unsure of the long term affects.

Still having errors on this PR on Screenshots in UI Tests with 2 files which are new:
`CommentCount:Default`
`Footer:WithCCPA`


JamieB-gu:
I suspect Emotion is the reason for the errors. We're using that to handle JSX. The React [docs suggest](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#manual-babel-setup):

> If you use JSX with a library other than React, you can use the importSource option to import from that library instead — as long as it provides the necessary entry points.

Emotion [provide support](https://emotion.sh/docs/css-prop#babel-preset) for this in their Babel **plugin** (not their preset, which is what we're currently using).

Unfortunately, the timing isn't great because Emotion recently restructured most of their libraries in version 11, and we're still on version 10. The documentation refers to `@emotion/babel-plugin` and `@emotion/react`, both of which only exist in version 11.

I've made some assumptions and managed to get this working by altering their recommended `importSource` from `@emotion/react` to `@emotion/core`. Fortunately the `jsx-runtime` that we need seems to have remained in the same place between versions, so this works. However, relying on libraries interacting across a major version boundary like this isn't ideal - I think we'll probably want to jump to Emotion 11 ASAP. Alternatively we could revert to v10 of `babel-plugin-emotion` (which is what `@emotion/babel-plugin` was called prior to the Emotion 11 renaming), but I'd prefer to go forwards if we can.

You can see the changes I made to get it working in 2bf19ed44ec09939f1904ac5618afe804fcc1b31.

FYI @SiAdcock @oliverlloyd @gtrufitt

JamieB-gu:
Fixed in #1005.

oliverlloyd:
@liywjl

PR #1000, opened 2020-12-02
Update hasSeenCards

faresite:
## Why are you doing this?

I'm implementing the iOS method in [this PR](https://github.com/guardian/ios-live/pull/5431). Apart from uncommenting I needed to strip the leading `/` from the ids as that's now how ids work in the iOS app. I'm checking with @maxspencer that this will be ok for Android as well.


src/client/article.ts
```diff
@@ -334,21 +334,20 @@ function callouts(): void {
...
-// 			document
-// 				.querySelector(`.js-card[data-article-id='${id}']`)
-// 				?.classList.add('fade');
-// 		});
-// 	});
-// }
+function hasSeenCards(): void {
+	const articleIds = Array.from(document.querySelectorAll('.js-card')).map(
+		(card) =>
+			card.getAttribute('data-article-id')?.replace(/^\//, '') ?? '',
```
JamieB-gu, line 340: Since we control `data-article-id` I think it might be easier to do this when we generate it on the server rather than relying on a client regex? This can be achieved by adding a `.slice(1)` [here](https://github.com/guardian/apps-rendering/blob/fa3f0de18a6015d8cbdf01f3ae3d1a6d4e72833c/src/components/shared/card.tsx#L418) and [here](https://github.com/guardian/apps-rendering/blob/fa3f0de18a6015d8cbdf01f3ae3d1a6d4e72833c/src/components/shared/bylineCard.tsx#L193).

On a related topic, as far as I can tell we don't need the `.split().pop()` on either of those lines? I think the `link` property is already in path format (i.e. it doesn't _have_ the `www.theguardian.com` domain prefix).

JamieB-gu (commented):
Looks good, one suggestion.

src/client/article.ts
faresite: 👍 Took a while but it's done now :)

joecowton1 (approved):
Nice work!

PR #1005, opened 2020-12-03
Emotion jsx react 17

dskamiotis:
## Why are you doing this?

This is a fix by Jamie B to implement the emotion changes in Babel required to allow JSX to work

Fom PR https://github.com/guardian/apps-rendering/pull/992


.metals/metals.lock.db
```diff
@@ -0,0 +1,6 @@
+#FileLock
+#Thu Dec 03 12:08:40 GMT 2020
+hostName=localhost
+id=1762881221a7123c544904eed0ec4292c00c5eee843
+method=file
+server=localhost\:55841
```
JamieB-gu, line 6: I think you may want to add the `.metals` directory to either the project `.gitignore` or your global `.gitignore`.

buck06191:
You might be able to fix the ESLint errors by following [this advice](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#eslint) from the React docs. Now that you no longer need to explicitly import `React`, I think you can turn off the eslint config that controls that

.metals/metals.lock.db
frankie297: Do we need metals in apps rendering?! do we have any scala code?

buck06191: Is it the Thrift definitions?

JamieB-gu: There's a _tiny_ bit of Scala/SBT code related to [generating the Thrift models](https://github.com/guardian/apps-rendering/tree/main/api-models) for communication between MAPI and AR, as @buck06191 says.

src/components/credit.tsx
```diff
@@ -8,8 +8,8 @@ import { Design } from '@guardian/types/Format';
 import type { Option } from '@guardian/types/option';
 import { map, withDefault } from '@guardian/types/option';
 import { pipe2 } from 'lib';
-import React from 'react';
 import type { FC, ReactElement } from 'react';
+import React from 'react';
```
buck06191, line 12: I've just realised it might be possible to make this an explicit import of `Fragment` and then change the shirt hand `<>` to `<Fragment>`. This would meet the original goal of removing the full `import React from 'react';` statement. 

Something like 
```typescript
import { Fragment } from 'react';
```
might do it, but you'd have to test after

JamieB-gu:
What's the difference between this and #992? Do we need both or can we close that one?

dskamiotis:
> What's the difference between this and #992? Do we need both or can we close that one?

Good point Ill close that one, used it to branch to create this one with a cleaner history of commits

src/components/credit.tsx
JamieB-gu: Are you sure this is needed? I'm not seeing any issues running without this import 🤔

src/components/shared/epic.tsx
```diff
@@ -14,7 +14,8 @@ import { body, headline } from '@guardian/src-foundations/typography';
 import { SvgArrowRightStraight } from '@guardian/src-icons';
 import { ThemeProvider } from 'emotion-theming';
 import { acquisitionsClient } from 'native/nativeApi';
-import React, { useEffect, useRef, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
+import * as React from 'react';
```
JamieB-gu, line 18: As above, are you sure this is needed?

src/components/shared/leftColumn.tsx
```diff
@@ -5,7 +5,7 @@ import { css } from '@emotion/core';
 import { remSpace } from '@guardian/src-foundations';
 import { from } from '@guardian/src-foundations/mq';
 import type { FC } from 'react';
-import React from 'react';
+import * as React from 'react';
```
JamieB-gu, line 8: As above, are you sure this is needed?

webpack.config.ts
```diff
@@ -69,9 +69,11 @@ const serverConfig = (
 			? []
 			: [
 					nodeExternals({
-						allowlist: [/@guardian/],
+						allowlist: [/@guardian/, 'React'],
```
JamieB-gu, line 72: What does this do?

JamieB-gu (commented):
Looks good, a few small questions.

src/components/shared/epic.tsx
buck06191: @JamieB-gu I'd be interested to know why you weren't getting problems with the `Fragment`s, but they were showing up on @dskamiotis's machine. Commenting here so that I can keep up to date on any discussion around that.

JamieB-gu: What kind of errors were you seeing? Lint, compile or runtime?

src/components/credit.tsx
dskamiotis: Comes up with an error on mine when not imported

JamieB-gu: What kind of errors were you seeing? Lint, compile or runtime?

src/components/shared/epic.tsx
dskamiotis: I believe runtime (during npm run test)

src/components/shared/leftColumn.tsx
dskamiotis: Just tested and makes no difference when omitted, not sure how that got in there

src/components/shared/epic.tsx
buck06191: I believe it was `ReferenceError: React is not defined`. I think that there's potentially some sort of problem with how Webpack and Babel are transpiling the `Fragment`s, though I'm not sure how or why.

JamieB-gu: Oh it's a _test_ failure? Right that makes sense, the new config needs to be applied to Jest as well. I've pushed a commit to this PR that should fix things.

src/components/credit.tsx
```diff
@@ -6,8 +6,8 @@ import { textSans } from '@guardian/src-foundations/typography';
 import type { Format, Option } from '@guardian/types';
 import { Design, map, withDefault } from '@guardian/types';
 import { pipe2 } from 'lib';
-import React from 'react';
 import type { FC, ReactElement } from 'react';
+import { Fragment } from 'react';
```
JamieB-gu, line 10: I don't think you need to do this now.

```suggestion
```

src/components/credit.tsx
```diff
@@ -29,7 +29,7 @@ const Credit: FC<Props> = ({ format, credit }) =>
 				case Design.Media:
 					return <p css={styles}>{cred}</p>;
 				default:
-					return <> {cred}</>;
+					return <Fragment> {cred}</Fragment>;
```
JamieB-gu, line 32: I don't think this is needed now that we have the Jest config updated.
```suggestion
					return <> {cred}</>;
```

webpack.config.ts
dskamiotis: Not sure if this was automatically done as part of the upgrade

JamieB-gu: What happens if you remove it?

dskamiotis: It may have been something i tried last week? cant remember, omitting it still passes the tests locally. Should we take it out?

dskamiotis: > What happens if you remove it?

passes local tests and run watch no problems

JamieB-gu: If it doesn't cause any errors I don't think you need it. Adding here is saying "when in development, please include React in my Webpack bundle".

dskamiotis: 👍

JamieB-gu (approved):

PR #1008, opened 2020-12-07
New endpoint for edition backend to serve rendered article

mohammad-haque:
## Why are you doing this?
This PR introduces a new endpoint for the edition backend to serve rendered articles based on capi search results. It receives a capi blob as a payload which then convert into `RenderingRequest` and then returns rendered html.

Edition backend can deal with all capi environments (live, preview, and printsent) to it make sense for edition backend to fetch the article from capi and then make a request with the payload.

The reason payload is capi blob but not `RenderingRequest` is that the edition backend is using a different version of the thrift model library than `app-rendering` and they are not compatible.


JamieB-gu:
> the edition backend is using a different version of the thrift model library than `app-rendering` and they are not compatible

Potential solution to that in https://github.com/guardian/editions/pull/1608.

src/server/server.ts
```diff
@@ -185,6 +190,27 @@ async function serveArticlePost(
...
+		const searchResponse = await capiSearchDecoder(req.body);
+		// this end point serve single article so take the first item only
+		const content = searchResponse.results[0];
+		const mockedRenderingRequest: RenderingRequest = {
+			content,
+		};
+		void serveArticle(mockedRenderingRequest, res, false);
+	} catch (e) {
+		console.error(e);
+		res.status(400).send('bad request');
```
JamieB-gu, line 210: We use a logging package to avoid some performance problems with using `console` under high request load, and some custom middleware to handle errors:

```suggestion
		logger.error('This error occurred', e);
		next(e);
```

src/server/server.ts
```diff
@@ -185,6 +190,27 @@ async function serveArticlePost(
...
+	req: Request,
+	res: ExpressResponse,
+	next: NextFunction,
+): Promise<void> {
+	try {
+		// req.body: should contain a direct response from capi
+		// fetched by the edition backend
+		const searchResponse = await capiSearchDecoder(req.body);
+		// this end point serve single article so take the first item only
+		const content = searchResponse.results[0];
```
JamieB-gu, line 203: Longer term how do we want to handle this? Are we always going to send a full search response and pull out the first article in the list? Is there an overhead cost associated with that?

src/server/server.ts
```diff
@@ -262,6 +288,8 @@ app.get('/*', bodyParser.raw(), serveArticleGet);
 
 app.post('/article', bodyParser.raw(), serveArticlePost);
 
+app.post('/render-edition-article', bodyParser.raw(), serveEditionArticlePost);
```
JamieB-gu, line 291: Perhaps name the endpoint this for consistency with the MAPI endpoint?

```suggestion
app.post('/editions-article', bodyParser.raw(), serveEditionArticlePost);
```

JamieB-gu (commented):
Nice! A few small comments.

src/server/server.ts
mohammad-haque: I have refactored this and now it receives a single 'Content' object from the payload

mohammad-haque:
@JamieB-gu I have updated this PR (refactoring + PR feedback)

Please have another look

mohammad-haque:
@JamieB-gu Could you please have another look at this PR?

src/server/server.ts
```diff
@@ -184,6 +189,25 @@ async function serveArticlePost(
...
+async function serveEditionsArticlePost(
+	req: Request,
+	res: ExpressResponse,
+	next: NextFunction,
+): Promise<void> {
+	try {
+		// The "req.body" should contain a 'Content' object which fetched by the
+		// Edition backend from the capi
+		const content = await capiContentDecoder(req.body);
+		const mockedRenderingRequest: RenderingRequest = {
```
JamieB-gu, line 201: I think you could probably just call this `renderingRequest`? "mocked" may cause confusion because it's typically only something that's done in tests?

JamieB-gu (approved):
🚀

PR #1019, opened 2020-12-14
LIVE-1250: Split out AR and ER renderers

faresite:
## Why are you doing this?

We use a `render` function to render blocks, now that some designs will diverge for ER (pullquote for example). We need two separate renderers, however a lot of the element renderers have been refactored out so we don't have to duplicate their code if the element is the same.

## Changes

- Refactor out element renderers that are > 1 line
- Created `renderEditions` and `renderEditionsAll`
- typed three elements: `Text`, `Embed` and `Instagram`

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -792,8 +792,105 @@ const render = (format: Format, excludeStyles = false) => (
 	}
 };
 
+const renderEditions = (format: Format, excludeStyles = false) => (
+	element: BodyElement,
+	key: number,
+): ReactNode => {
+	switch (element.kind) {
+		case ElementKind.Text:
+			return textRenderer(format, excludeStyles, element);
+
+		case ElementKind.Image:
+			return imageRenderer(format, element, key);
+
+		case ElementKind.Pullquote: {
+			const { quote, attribution } = element;
+			return h(Pullquote, { quote, attribution, format, key });
+		}
+
+		case ElementKind.RichLink: {
+			const { url, linkText } = element;
+			return h(RichLink, { url, linkText, key, format });
+		}
+
+		case ElementKind.LiveEvent: {
+			return h(LiveEventLink, { ...element, key });
+		}
+
+		case ElementKind.Interactive:
+			return h(Interactive, {
+				url: element.url,
+				key,
+				title: element.alt,
+			});
+
+		case ElementKind.Tweet:
+			return h(Tweet, { content: element.content, format, key });
+
+		case ElementKind.Audio:
+			return h(Audio, {
+				src: element.src,
+				width: element.width,
+				height: element.height,
+			});
+
+		case ElementKind.Video:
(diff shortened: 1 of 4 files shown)


src/renderer.ts
```diff
@@ -492,226 +752,148 @@ const render = (format: Format, excludeStyles = false) => (
...
-				element.alt,
-				map((alt) => figure(alt)),
-				withDefault(figure(null)),
-			);
+		case ElementKind.Instagram:
+			return instagramRenderer(element);
+
+		case ElementKind.ExplainerAtom: {
+			return h(ExplainerAtom, { ...element });
 		}
```
JamieB-gu, line 763:
```suggestion
		case ElementKind.ExplainerAtom:
			return h(ExplainerAtom, { ...element });
```

src/renderer.ts
```diff
@@ -492,226 +752,148 @@ const render = (format: Format, excludeStyles = false) => (
...
-			return h('div', props);
+		case ElementKind.ProfileAtom:
+			return profileAtomRenderer(format, element);
+
+		case ElementKind.TimelineAtom:
+			return timelineAtomRenderer(format, element);
+
+		case ElementKind.ChartAtom: {
+			return h(ChartAtom, { ...element });
 		}
```
JamieB-gu, line 779:
```suggestion
		case ElementKind.ChartAtom:
			return h(ChartAtom, { ...element });
```

src/renderer.ts
```diff
@@ -492,226 +752,148 @@ const render = (format: Format, excludeStyles = false) => (
...
+	switch (element.kind) {
+		case ElementKind.Text:
+			return textRenderer(format, excludeStyles, element);
+
+		case ElementKind.Image:
+			return imageRenderer(format, element, key);
+
+		case ElementKind.Pullquote: {
+			const { quote, attribution } = element;
+			return h(Pullquote, { quote, attribution, format, key });
```
JamieB-gu, line 808: Is the next step to build an `EditionsPullquote` component?

src/renderer.ts
```diff
@@ -492,226 +752,148 @@ const render = (format: Format, excludeStyles = false) => (
...
-				},
-				dislikeHandler: () => {
-					console.log('dislike clicked');
-				},
-				expandCallback: () => {
-					console.log('expand clicked');
-				},
-			});
+		case ElementKind.LiveEvent: {
+			return h(LiveEventLink, { ...element, key });
```
JamieB-gu, line 817: I think you can drop the enclosing `{}` here (it won't let me add the suggestion).

JamieB-gu (approved):
Excellent! This is the direction I've been hoping we'd go in ⬆️

src/renderer.ts
faresite: Indeed 👍

PR #1023, opened 2020-12-15
LIVE-1282: Editions-Rendering CSP

faresite:
## Why are you doing this?

The CSP is somewhat broken on editions-rendering. The inline styles weren't working and embeds (like YouTube for example) were also broken. I have decided to get inspired by the way it works on AR and use a similar but custom function to render ER's head. 

## Changes

- Change CSP generation
- upgraded `renderHead`

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/server/csp.ts
+++ b/src/server/csp.ts
@@ -138,4 +138,4 @@ function csp(
 
 // ----- Exports ----- //
 
-export { csp, assetHashes };
+export { csp, assetHashes, buildCsp };
--- a/src/server/csp.ts
+++ b/src/server/csp.ts
@@ -134,19 +134,13 @@ function cspEditions(
 			: ''
 	};
     script-src 'self' ${
-		thirdPartyEmbed.instagram ? 'http://www.instagram.com/embed.js' : ''
-	} https://interactive.guim.co.uk https://s16.tiktokcdn.com https://www.tiktok.com/embed.js https://sf16-scmcdn-sg.ibytedtos.com/ ${
 		thirdPartyEmbed.twitter
 			? 'https://platform.twitter.com https://cdn.syndication.twimg.com'
 			: ''
 	};
-    frame-src https://www.theguardian.com https://www.scribd.com ${
-		thirdPartyEmbed.instagram ? 'https://www.instagram.com' : ''
-	} https://www.facebook.com https://www.tiktok.com https://interactive.guim.co.uk ${
-		thirdPartyEmbed.spotify ? 'https://open.spotify.com' : ''
-	} ${
+    frame-src https://www.theguardian.com ${
 		thirdPartyEmbed.youtube ? 'https://www.youtube-nocookie.com' : ''
-	} https://player.vimeo.com/ ${
+	} ${
 		thirdPartyEmbed.twitter
 			? 'https://platform.twitter.com https://syndication.twitter.com https://twitter.com'
 			: ''
--- a/src/server/csp.ts
+++ b/src/server/csp.ts
@@ -121,6 +121,39 @@ const buildCsp = (
     media-src 'self' https://audio.guim.co.uk/
 `.trim();
 
+function cspEditions(
+	{ styles }: Assets,
+	thirdPartyEmbed: ThirdPartyEmbeds,
+): string {
+	return `
+	default-src 'self';
+	style-src 'self' ${assetHashes(styles)};
+	img-src 'self' https://static.theguardian.com https://*.guim.co.uk ${
+		thirdPartyEmbed.twitter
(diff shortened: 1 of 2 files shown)


src/server/editionsPage.tsx
```diff
@@ -89,9 +92,16 @@ const buildHtml = (head: string, body: string): string => `
 function render(imageSalt: string, request: RenderingRequest): Page {
 	const item = fromCapi({ docParser, salt: imageSalt })(request);
 	const body = renderBody(item);
+	const head = renderHead(
+		request,
+		getThirdPartyEmbeds(request.content),
+		body.css,
+		body.ids,
+		requiresInlineStyles(request.content),
```
JamieB-gu, line 100: Oh! I didn't realise this had been activated... That's not good, it [applies `unsafe-inline`](https://github.com/guardian/apps-rendering/blob/bbf196d8665902d760f7b3f405f2455203276817/src/server/csp.ts#L84) to our `style-src`.

Please don't rely on `unsafe-inline` to be available, I think I'm going to turn it off again for now.

src/server/editionsPage.tsx
```diff
@@ -46,23 +41,31 @@ const styles = `
...
-				{emotionCritical.css}
-			</style>
-		</>,
+	thirdPartyEmbeds: ThirdPartyEmbeds,
+	itemStyles: string,
+	emotionIds: string[],
+	inlineStyles: boolean,
+): string => {
+	const generalStyles = styles;
+	const cspString = buildCsp(
```
JamieB-gu, line 50: Are you sure you want the same CSP as live apps content? This ties them together, and means that the Editions CSP may be more permissive than it has to be, and likewise for the live apps CSP if Editions needs something enabled.

Would it make sense to abstract out some of the logic in the `csp` and `buildCsp` functions and generate two different CSPs using that logic?

JamieB-gu (commented):
The approach looks good! 👍 for using @dskamiotis's third-party embeds work.

Couple of specific questions/comments.

src/server/editionsPage.tsx
JamieB-gu: #1025

src/server/csp.ts
```diff
@@ -121,6 +121,39 @@ const buildCsp = (
...
+	default-src 'self';
+	style-src 'self' ${assetHashes(styles)};
+	img-src 'self' https://static.theguardian.com https://*.guim.co.uk ${
+		thirdPartyEmbed.twitter
+			? 'https://platform.twitter.com https://syndication.twitter.com https://pbs.twimg.com data:'
+			: ''
+	};
+    script-src 'self' ${
+		thirdPartyEmbed.instagram ? 'http://www.instagram.com/embed.js' : ''
+	} https://interactive.guim.co.uk https://s16.tiktokcdn.com https://www.tiktok.com/embed.js https://sf16-scmcdn-sg.ibytedtos.com/ ${
```
JamieB-gu, line 138: If Editions doesn't currently support Interactives, Instagram or TikTok (`ibytedtos` is a TikTok domain FYI) then I'd be inclined to drop these URLs until you _know_ you need them.

src/server/csp.ts
```diff
@@ -121,6 +121,39 @@ const buildCsp = (
...
+    frame-src https://www.theguardian.com https://www.scribd.com ${
+		thirdPartyEmbed.instagram ? 'https://www.instagram.com' : ''
+	} https://www.facebook.com https://www.tiktok.com https://interactive.guim.co.uk ${
+		thirdPartyEmbed.spotify ? 'https://open.spotify.com' : ''
+	} ${
+		thirdPartyEmbed.youtube ? 'https://www.youtube-nocookie.com' : ''
+	} https://player.vimeo.com/ ${
+		thirdPartyEmbed.twitter
+			? 'https://platform.twitter.com https://syndication.twitter.com https://twitter.com'
+			: ''
```
JamieB-gu, line 152: Same as above WRT Scribd, Instagram, TikTok, Facebook, Spotify, Vimeo and possibly YouTube (is YouTube required for Editions' video hosting?).

JamieB-gu (commented):
I like this approach 👍.

Couple of comments about making the CSP stricter by default.

src/server/csp.ts
dskamiotis: We are dealing with standard articles in the first instance, are there no standard articles with Video embeds? If so looking at: https://miro.com/app/board/o9J_lcA5uzY=/ this fall under MainMedia?

JamieB-gu (approved):

PR #1025, opened 2020-12-15
Disable "unsafe-inline"

JamieB-gu:
## Why are you doing this?

We would like to avoid `unsafe-inline`, it [sidesteps most of](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src#Unsafe_inline_styles) the security benefits provided by the CSP.

## Changes

- Disable `unsafe-inline` by hardcoding `requiresInlineStyles` to `false`


faresite (approved):
👍

JamieB-gu:
Note that this has temporarily broken some of the discussion styles, as seen in the "Review" screenshot above

PR #1027, opened 2020-12-18
Editions screenshots in CI

faresite:
Make use of the lovely screenshot workflow to add some editions articles. For now I only added:

- Standard
- Opinion
- Review

### DIFF
--- a/.github/workflows/screenshots.yml
+++ b/.github/workflows/screenshots.yml
@@ -32,6 +32,12 @@ jobs:
                       url: 'http://localhost:8080/empowering-business/2020/oct/31/why-would-i-ever-go-back-to-shopping-new-how-vintage-has-become-the-future-of-buying-ethically'
                     - format: Analysis
                       url: 'http://localhost:8080/us-news/2020/nov/13/joe-biden-cabinet-selection-trump-obstruction'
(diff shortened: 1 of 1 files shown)


faresite:
@jacobwinch @JamieB-gu I'm a bit struggling with this one, I have the feeling the screenshots of `main` are broken because we might need to merge the PR first so they're populated for `main`?!

JamieB-gu:
> I have the feeling the screenshots of `main` are broken because we might need to merge the PR first so they're populated for `main`?!

Yeah that's correct.

I'm a bit puzzled about why the new screenshots just seem to be header images though 🤔

faresite:
> > I have the feeling the screenshots of `main` are broken because we might need to merge the PR first so they're populated for `main`?!
> 
> Yeah that's correct.
> 
> I'm a bit puzzled about why the new screenshots just seem to be header images though 🤔

I believe they're header images at the moment because of the current state of edition-rendering styles.

JamieB-gu (approved):
> I believe they're header images at the moment because of the current state of edition-rendering styles.

Ah, yeah you're right, they broken 😔 - odd, they were working before... Anyway, unrelated to these changes.

PR #1029, opened 2020-12-21
LIVE-1247: Byline and share component (and editions communication layer)

faresite:
## Editions Native 🔁 Editions Rendering communication
This PR includes work on the communication layer. The initial file can be found in `editionsCommunication.ts` which could be a candidate for a package that the editions team can use. 
To communicate from editions rendering to react native we can use the `pingEditionsNative` function that receives a `Message`. To send a message from react native to editions rendering we can use the `pingEditionsRenderingJsString` function that uses the `pingEditionsRendering` under the hood to fire a custom event that editions-rendering can listen to be in vanilla dom or on the react side (see `usePlatform` hook for example).
The serializer function `pingEditionsRenderingJsString` is just a helper function as you need to pass a string when injecting javascript through `react-native-webview`.

## Byline and share component

@abeddow91 and I worked on the UI via this [Figma file](https://www.figma.com/file/JcDsPjVHSL7f5MEtzIrlHj/Webview-Components?node-id=14%3A0). We took inspiration from the apps-rendering `byline` component.


package-lock.json
```diff
@@ -3992,9 +3992,9 @@
...
     "@types/serve-static": {
-      "version": "1.13.6",
-      "resolved": "https://registry.npmjs.org/@types/serve-static/-/serve-static-1.13.6.tgz",
-      "integrity": "sha512-nuRJmv7jW7VmCVTn+IgYDkkbbDGyIINOeu/G0d74X3lm6E5KfMeQPJhxIt1ayQeQB3cSxvYs1RA/wipYoFB4EA==",
+      "version": "1.13.8",
+      "resolved": "https://registry.npmjs.org/@types/serve-static/-/serve-static-1.13.8.tgz",
+      "integrity": "sha512-MoJhSQreaVoL+/hurAZzIm8wafFR6ajiTM1m4A0kv6AGeVBl4r4pOV8bGFrjjq1sGxDTnCoF8i22o0/aE5XCyA==",
       "dev": true,
       "requires": {
         "@types/mime": "*",
```
JamieB-gu, line 4000: Some Express type definitions being bumped? 🤔

src/client/editionsCommunication.ts
```diff
@@ -0,0 +1,93 @@
...
+};
+
+export type PlatformMessage = {
+	kind: 'platform';
+	value: Platform;
+};
+
+type PlatformQueryMessage = {
+	kind: 'platform-query';
+};
```
JamieB-gu, line 24: How do you feel about using a `const enum` for the `kind` fields?

```ts
const enum MessageKind {
  Share,
  Platform,
  PlatformQuery,
}
```

src/client/editionsCommunication.ts
```diff
@@ -0,0 +1,93 @@
...
+type Message = ShareMessage | PlatformMessage | PlatformQueryMessage;
+
+export enum Platform {
+	IOS = 'IOS',
+	Android = 'Android',
+}
+
+export const isPlatformMessageEvent = (
+	customEvent: CustomEventInit<Message>,
+): customEvent is CustomEvent<PlatformMessage> => {
```
JamieB-gu, line 35: Nice use of type guards!

src/client/editionsCommunication.ts
```diff
@@ -0,0 +1,93 @@
...
+		`%c${logMessage}
+${JSON.stringify(parsedData, null, 2)}`,
+		'color: #45969b;',
+	);
+};
+
+export const pingEditionsNative = (message: Message): void => {
+	const serializedMessage = JSON.stringify(message);
+
+	if (process.env.NODE_ENV === 'development') {
```
JamieB-gu, line 58: Is `process` defined on the client?

src/client/editionsCommunication.ts
```diff
@@ -0,0 +1,93 @@
...
+            window.pingEditionsRendering(${JSON.stringify(message)})
+        } catch {
+            console.error("Editions -> Editions Rendering not initiated")
+        }
+    `;
+};
+
+export const initPingEditionsRendering = (): void => {
+	window.pingEditionsRendering = pingEditionsRendering;
+	window.pingEditionsRenderingJsString = pingEditionsRenderingJsString;
```
JamieB-gu, line 92: Does this function need to live on the `window`? Whereas `pingEditionsRendering` will be running on the webview thread, `pingEditionsRenderingJsString` will be running on the native thread? And the string *returned* by `pingEditionsRenderingJsString` will be running on the webview thread.

JamieB-gu (commented):
Nice 🚀! Haven't had a chance to look through everything yet, couple comments so far.

package-lock.json
faresite: Ha I removed that bump from package.json but not from the lock file 😅

src/client/editionsCommunication.ts
faresite: definitely 👍

src/client/editionsCommunication.ts
faresite: I always believed this is all stripped out in the production bundle thanks to the likes of webpack and friends. Now you said it, I will double check that this doesn't ship in production bundle 👍

src/client/editionsCommunication.ts
faresite: Your three points are correct.
In my head the `pingEditionsRendering` function needs to live on the window as **native** thread needs to be able to call it somehow, do you have an idea around that?
You're right the `JsString` function does not need to live on the window, I think I added so I can test it :) I will remove it 👍

src/client/editionsCommunication.ts
faresite: production bundle strips the whole things out 👍

abeddow91:
@faresite - From an Editions app perspective, this looks good to me! I've tested in the Editions app and messages are being communicated as expected. Using editionsCommunication.ts as a seperate package feels like a logical next step

joecowton1 (approved):

abeddow91 (approved):

PR #1056, opened 2021-01-07
Make links colour palette 300

frankie297:
## Why are you doing this?
Make links colour palette 300

## Screenshots

| Before | After |
| --- | --- |
|  |  |

Paired with @joecowton1

### DIFF
--- a/src/components/anchor.tsx
+++ b/src/components/anchor.tsx
@@ -29,7 +29,7 @@ const styles = css`
 `;
 
 const colour = (format: Format): SerializedStyles => {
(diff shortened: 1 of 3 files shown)


JamieB-gu:
Why does this PR include changes to `package-lock.json`? Is it possible a rebase on `main` is needed?

frankie297:
> Why does this PR include changes to `package-lock.json`? Is it possible a rebase on `main` is needed?

What would be the best way to do this?

frankie297:
@JamieB-gu we have removed the package.lock changes from the PR

JamieB-gu (approved):

PR #1073, opened 2021-01-11
Add Cache Action To Chromatic CI

JamieB-gu:
## Why are you doing this?

Our other workflows (e.g. [nodejs](https://github.com/guardian/apps-rendering/blob/3ee405708fedbed4ce8bc799d86e6b419edc701c/.github/workflows/nodejs.yml) and [screenshots](https://github.com/guardian/apps-rendering/blob/3ee405708fedbed4ce8bc799d86e6b419edc701c/.github/workflows/screenshots.yml)) are using the [cache action](https://github.com/actions/cache) to cache npm dependencies.

**Note:** The chromatic workflow is currently failing on other PRs on the `npm ci` step. This is the only difference I can see between them so I thought I'd add it (because it's useful in general), but I don't see why it would fix the problem 🤔.

## Changes

- Add cache action to chromatic CI

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -18,7 +18,13 @@ jobs:
             - uses: actions/setup-node@v1
               with:
                   node-version: 12
-            - run: npm ci
+            - uses: actions/cache@v2
+              with:
+                  path: ~/.npm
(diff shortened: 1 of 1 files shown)


faresite (approved):
Looks good to me 👍 and the PR passes so that's something!

joecowton1:
👍

PR #1078, opened 2021-01-12
Update .nvmrc To 12.20.1

JamieB-gu:
## Why are you doing this?

Ensures that developers who are using `nvm` with this repo are on the latest version of Node 12 (and its corresponding latest version of npm).

## Changes

- Updated `.nvmrc` to 12.20.1


joecowton1 (approved):

frankie297 (approved):

faresite:
@JamieB-gu can you remind me why can't we merge this or bump it even higher to LTS `14.16`?

JamieB-gu:
> @JamieB-gu can you remind me why can't we merge this or bump it even higher to LTS `14.16`?

We could bump it higher - we'll just have to make sure it works.

On the merge issue, the Node version is configured in different places depending on your environment. For some CI it's configured [here](https://github.com/guardian/apps-rendering/blob/cdccc37a28b98105ff1abc87bc48be434445b359/.github/workflows/nodejs.yml#L19) and for others it's configured in TC (although the TC build may use this `.nvmrc`). For PROD it's configured in the AMI.

Bumping the `.nvmrc` file on its own will only bump Node in local dev and possibly in TC. This is not an ideal situation. I think DevX are thinking about this, and @sndrs has introduced a fork of [`actions-setup-node`](https://github.com/guardian/actions-setup-node), which we could migrate to to solve the problem for GH Actions.

JamieB-gu:
Fixed in #1443.

PR #1081, opened 2021-01-13
Add index Function

JamieB-gu:
## Why are you doing this?

Useful for converting array lookups to `Option`s. If there's nothing at a given array index then JS will return `undefined`, but TS doesn't type it as such. See [this discussion](https://github.com/guardian/apps-rendering/pull/1031#discussion_r556731695) for a use in practice.

Opening as a Draft for now to gather feedback on whether people would find this useful.

## Changes

- Added an `index` utility function

### DIFF
--- a/src/lib.ts
+++ b/src/lib.ts
@@ -1,7 +1,7 @@
 // ----- Imports ----- //
 
 import type { Option } from '@guardian/types';
-import { map, withDefault } from '@guardian/types';
+import { fromNullable, map, withDefault } from '@guardian/types';
 import type { ReactElement } from 'react';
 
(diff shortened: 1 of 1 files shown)


joecowton1 (approved):

src/lib.ts
```diff
@@ -62,6 +62,9 @@ function handleErrors(response: Response): Response | never {
 	}
 	return response;
 }
+
+const index = (i: number) => <A>(arr: A[]): Option<A> => fromNullable(arr[i]);
```
faresite, line 66: I would actually love that! 👍

faresite (approved):

PR #1087, opened 2021-01-15
feat(editions-article): add border to header components

joecowton1:
## Why are you doing this?

Adds borders to Editions Article template headers.

## Changes

- Creates new width variable and applies to all required header components
- Adds borders where needed according to figma designs: https://www.figma.com/file/AgfOrrLI78asahmA3Z9Env/AUS%2FUS-xtra-designs?node-id=1713%3A0

A separate ticket add borders to the Editions Article body components is here:
https://theguardian.atlassian.net/browse/LIVE-1388

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/editions/article.tsx
+++ b/src/components/editions/article.tsx
@@ -8,14 +8,24 @@ import { Design, partition } from '@guardian/types';
 import type { Item } from 'item';
 import type { FC } from 'react';
 import { renderEditionsAll } from 'renderer';
-import { wideContentWidth } from 'styles';
 import Header from './header';
+import { editionsArticleWidth } from './styles';
 
 // ----- Component ----- //
 
 interface Props {
 	item: Item;
 }
+
+const articleStyles = css`
+	padding-left: ${remSpace[2]};
+	padding-right: ${remSpace[2]};
+
+	${from.phablet} {
+		padding-left: 0;
+		padding-right: 0;
+	}
+`;
 
 const bodyStyles = css`
 	border-top: 1px solid ${border.secondary};
@@ -27,7 +37,7 @@ const bodyStyles = css`
 
 	${from.phablet} {
 		margin-left: ${remSpace[24]};
-		width: ${wideContentWidth}px;
+		width: ${editionsArticleWidth}rem;
 	}
 `;
 
@@ -38,7 +48,7 @@ const Article: FC<Props> = ({ item }) => {
 
 	return (
 		<main>
-			<article>
+			<article css={articleStyles}>
 				<Header item={item} />
 				<section css={[bodyStyles]}>
 					{renderEditionsAll(item, partition(item.body).oks)}
--- a/src/components/editions/byline.tsx
+++ b/src/components/editions/byline.tsx
@@ -2,18 +2,30 @@
 
 import { css } from '@emotion/core';
 import { remSpace } from '@guardian/src-foundations';
-import { news } from '@guardian/src-foundations/palette';
+import { from } from '@guardian/src-foundations/mq';
+import { border, news } from '@guardian/src-foundations/palette';
 import { body } from '@guardian/src-foundations/typography';
 import type { Item } from 'item';
 import { maybeRender } from 'lib';
 import type { FC } from 'react';
+import { editionsArticleWidth } from './styles';
 
 // ----- Component ----- //
 
 const styles = css`
 	${body.medium({ fontStyle: 'normal', fontWeight: 'bold' })}
 	color: ${news[400]};
-	margin-bottom: ${remSpace[4]};
+	padding-bottom: ${remSpace[4]};
+	margin: 0 ${remSpace[1]} 0 0;
+
+	${from.wide} {
+		margin: 0 auto;
+	}
+
+	${from.phablet} {
+		border-right: 1px solid ${border.secondary};
+		width: ${editionsArticleWidth}rem;
+	}
 `;
 
 interface Props {
--- a/src/components/editions/headline.tsx
+++ b/src/components/editions/headline.tsx
@@ -1,10 +1,13 @@
 // ----- Imports ----- //
 
 import { css } from '@emotion/core';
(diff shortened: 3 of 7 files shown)


src/components/editions/article.tsx
```diff
@@ -8,15 +8,25 @@ import { Design, partition } from '@guardian/types';
 import type { Item } from 'item';
 import type { FC } from 'react';
 import { renderEditionsAll } from 'renderer';
-import { wideContentWidth } from 'styles';
 import Header from './header';
+import { editionsArticleWidth } from './styles';
```
JamieB-gu, line 12: 👍

src/components/editions/article.tsx
```diff
@@ -27,7 +37,7 @@ const bodyStyles = css`
 
 	${from.phablet} {
 		margin-left: ${remSpace[24]};
-		width: ${wideContentWidth}px;
+		width: ${editionsArticleWidth}px;
```
JamieB-gu, line 40: Would we want to do this with `rem`s? What happens if you zoom (`Cmd +`)?

src/components/editions/byline.tsx
```diff
@@ -2,18 +2,30 @@
...
+	padding-bottom: ${remSpace[4]};
+	margin: 0 ${remSpace[1]} 0 0;
+
+	${from.wide} {
+		margin: 0 auto;
+	}
+
+	${from.phablet} {
+		border-right: 1px solid ${border.secondary};
+		width: ${editionsArticleWidth}px;
```
JamieB-gu, line 27: Same question about `px` vs `rem`s.

src/components/editions/headline.tsx
```diff
@@ -1,17 +1,30 @@
...
 	margin: 0;
+	padding-bottom: ${remSpace[4]};
+
+	${from.wide} {
+		margin: 0 auto;
+	}
+
+	${from.phablet} {
+		border-right: 1px solid ${border.secondary};
+		width: ${editionsArticleWidth}px;
```
JamieB-gu, line 26: Same question about `px` vs `rem`s.

src/components/editions/lines.tsx
```diff
@@ -2,15 +2,21 @@
...
 import type { FC } from 'react';
-import { wideContentWidth } from 'styles';
+import { editionsArticleWidth } from './styles';
 
 // ----- Component ----- //
 
 const styles = css`
 	${from.phablet} {
-		width: ${wideContentWidth}px;
+		width: ${editionsArticleWidth}px;
```
JamieB-gu, line 14: Same question about `px` vs `rem`s.

src/components/editions/series.tsx
```diff
@@ -22,6 +24,16 @@ const styles = (item: Item): SerializedStyles => {
...
 		padding: ${remSpace[1]} 0 ${remSpace[2]};
+		border-top: 1px solid ${border.secondary};
+
+		${from.wide} {
+			margin: 0 auto;
+		}
+
+		${from.phablet} {
+			border-right: 1px solid ${border.secondary};
+			width: ${editionsArticleWidth}px;
```
JamieB-gu, line 35: Same question about `px` vs `rem`s.

src/components/editions/standfirst.tsx
```diff
@@ -1,33 +1,35 @@
...
 	color: ${text.primary};
 
 	${from.wide} {
 		margin: 0 auto;
 	}
 
 	${from.phablet} {
-		width: ${wideContentWidth}px;
+		border-right: 1px solid ${border.secondary};
+		width: ${editionsArticleWidth}px;
```
JamieB-gu, line 26: Same question about `px` vs `rem`s.

src/components/editions/styles.ts
```diff
@@ -0,0 +1 @@
+export const editionsArticleWidth = 457;
```
JamieB-gu, line 1: This is a very specific number 🤔

JamieB-gu (commented):
Looks good 🎉, couple of small questions.

src/components/editions/styles.ts
joecowton1: Indeed - was going of these designs: https://www.figma.com/file/AgfOrrLI78asahmA3Z9Env/AUS%2FUS-xtra-designs?node-id=1713%3A0

Is there a grid value we can use instead?

src/components/editions/article.tsx
joecowton1: So I think this behaves as I'd imagined, ie it zooms ok. Have changed to REM though if that's the preferred approach.

src/components/editions/byline.tsx
joecowton1: Updated.

src/components/editions/headline.tsx
joecowton1: Updated.

src/components/editions/lines.tsx
joecowton1: Updated.

src/components/editions/series.tsx
joecowton1: Updated.

src/components/editions/standfirst.tsx
joecowton1: Updated.

src/components/editions/styles.ts
JamieB-gu: Question for @prisalcalde @pradasa @benwuersching, is there a width from Source's grid that we could use for this?

src/components/editions/styles.ts
```diff
@@ -0,0 +1 @@
+export const editionsArticleWidth = '28rem';
```
JamieB-gu, line 1: Don't all the places this is used also add `rem` to the end? So the result would be `28remrem`?

joecowton1: 🤦 file hadn't saved / committed. Should be there now,

JamieB-gu (approved):

PR #1091, opened 2021-01-18
Feat: Live 986 - Review Article Template

abeddow91:
## Why are you doing this?

Updates styling to reflect the Review format designs

## Changes

- Adds Review-specific styling to Header and Headline (LIVE-1414)
- Adds editions star rating component (LIVE-1394)
- Separates Showcase specific styling from Body styling

## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |
|  |  |

### DIFF
--- a/src/components/editions/header.tsx
+++ b/src/components/editions/header.tsx
@@ -38,6 +38,7 @@ const reviewHeaderStyles = css`
 `;
 
 const showcaseHeaderStyles = css`
+	padding: 0 ${remSpace[2]};
 	margin: 0;
 
 	${from.wide} {
@@ -45,6 +46,8 @@ const showcaseHeaderStyles = css`
 	}
 
 	${from.phablet} {
+		padding-left: 0;
+		padding-right: 0;
 		margin-left: ${remSpace[24]};
 	}
 `;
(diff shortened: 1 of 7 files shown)


src/components/editions/header.tsx
```diff
@@ -12,60 +13,84 @@ import Series from 'components/editions/series';
...
+	</header>
 );
 
 const renderArticleHeader = (item: Item): ReactElement<Props> => {
-	switch (item.display) {
-		case Display.Showcase:
-			return <ShowcaseHeader item={item} />;
+	if (Display.Showcase === item.display)
+		return <ShowcaseHeader item={item} />;
+	switch (item.display | item.design) {
```
joecowton1, line 80: Does it make sense to be more implicit here i.e. 
```
if(item.design === Design.Review){}
```
Do we gain anything from the `switch` as it is?

joecowton1 (commented):
One minor comment but looking really nice 👌

src/components/editions/header.tsx
abeddow91: Good point - I'll edit in favour of if/else here

joecowton1 (approved):

PR #1099, opened 2021-01-20
LIVE-1442: add icon colours to theme styles and pass as optional props to headerImageCaption

joecowton1:
## Why are you doing this?

Adds colour picker switches for both outer circle and camera svg.

## Changes

- creates Editions specific `headerImageCaption` component
- adds Format based colour switches

For screenshots see generated images below.


faresite:
@joecowton1 PR looks good 👍  was just wondering since we kind want to strike a balance between what to reuse and what not to .. The only difference between apps-rendering and editions-rendering for this particular component, would be the colour styling.. 
How do you feel about updating the original component to have a flag prop, something like `pillarStyles` that defaults to false for apps-rendering instead of duplicating it?

joecowton1:
> @joecowton1 PR looks good 👍 was just wondering since we kind want to strike a balance between what to reuse and what not to .. The only difference between apps-rendering and editions-rendering for this particular component, would be the colour styling..
> How do you feel about updating the original component to have a flag prop, something like ~`pillarStyles`~ `withPillarColor` that defaults to false for apps-rendering instead of duplicating it?

Sounds like a good idea! Where do you think we should put the pillar colour switches then? In the AR theme or in the component itself?

faresite:
> @faresite Sounds like a good idea! Where do you think we should put the pillar colour switches then? In the AR theme or in the component itself?

I still like your idea of having the switches outside as a helper function. It might make sense to make the function itself more generic, so it could be used in the component like `getThemeColour(theme, 400)` and `getThemeColour(theme, 800)`, but I can also see the benefit of a more explicit function too.

JamieB-gu:
> was just wondering since we kind want to strike a balance between what to reuse and what not to

Was wondering this too on this one, good call @faresite 👍

joecowton1:
> > @faresite Sounds like a good idea! Where do you think we should put the pillar colour switches then? In the AR theme or in the component itself?
> 
> I still like your idea of having the switches outside as a helper function. It might make sense to make the function itself more generic, so it could be used in the component like `getThemeColour(theme, 400)` and `getThemeColour(theme, 800)`, but I can also see the benefit of a more explicit function too.

@faresite ended up pulling the `getThemeStyles()` function up into the editions `headerImage` component and passing in the colour variables as optional props. If these aren't present the icon renders with defaults so it works for both ER & AR.

faresite (approved):
Nice one @joecowton1 👍

PR #1104, opened 2021-01-21
Live-1407: Interview Header Component + Headline

abeddow91:
## Why are you doing this?
Creates new Header component for the Interview template and also adds Interview-specific headline styles

## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |


src/components/editions/headline.tsx
```diff
@@ -52,8 +52,38 @@ const standardFontStyles = css`
...
+	}
+
+	${from.tablet} {
+		${headline.medium({ lineHeight: 'regular' })}
+		font-weight: 400;
+	}
+	background-color: ${neutral[0]};
+	color: ${neutral[100]};
+	white-space: pre-wrap;
+	padding-bottom: 0.3125rem;
```
joecowton1, line 78: Can we use a `remSpace` value here?

src/components/editions/headline.tsx
```diff
@@ -52,8 +52,38 @@ const standardFontStyles = css`
...
+
+const interviewFontStyles = css`
+	${articleWidthStyles}
+
+	${headline.xsmall({ lineHeight: 'regular' })}
+		font-weight: 400;
+
+	${from.mobileMedium} {
+		${headline.small({ lineHeight: 'regular' })}
+		font-weight: 400;
```
joecowton1, line 68: We probably don't need to re-specify this value here as its set above.

src/components/editions/headline.tsx
```diff
@@ -52,8 +52,38 @@ const standardFontStyles = css`
...
+	}
+
+	${from.tablet} {
+		${headline.medium({ lineHeight: 'regular' })}
+		font-weight: 400;
```
joecowton1, line 73: Same here.

src/components/editions/headline.tsx
```diff
@@ -52,8 +52,38 @@ const standardFontStyles = css`
 		${headline.medium({ lineHeight: 'tight' })}
 	}
 `;
+const interviewStyles = css`
+	margin-left: 0.75rem;
+	border: 0;
+`;
+
+const interviewFontStyles = css`
+	${articleWidthStyles}
```
joecowton1, line 61: We shouldn't need to set the width here as it's already on the `Header` container.

src/components/editions/headline.tsx
```diff
@@ -52,8 +52,38 @@ const standardFontStyles = css`
 		${headline.medium({ lineHeight: 'tight' })}
 	}
 `;
+const interviewStyles = css`
+	margin-left: 0.75rem;
```
joecowton1, line 56: We should probably use a `remSpace` value here

joecowton1 (commented):
Looks good! Was occasionally overlapping the line border for me, but removing the width fixes that. Few other minor CSS comments but otherwise 👍

src/components/editions/headline.tsx
abeddow91: I thought that but, because the headline has a default weight of 500, it gets overwritten. Ideally we might be able to get designs to use Source standard weights but thats tbc.

src/components/editions/headline.tsx
abeddow91: Same as above

src/components/editions/headline.tsx
joecowton1: Oh sorry - that's annoying! Yeah agreed.

src/components/editions/headline.tsx
abeddow91: Nice spot  - converted.

src/components/editions/headline.tsx
abeddow91: Nice spot  - converted.

src/components/editions/headline.tsx
abeddow91: Nice spot  - removed.

joecowton1 (approved):

package.json
```diff
@@ -55,6 +55,7 @@
     "@guardian/src-text-area": "^2.8.1",
     "@guardian/src-text-input": "^2.7.1",
     "@guardian/types": "github:guardian/types#semver:1.1.0",
+    "@guardian/renditions": "git+https://github.com/guardian/renditions.git#0.1.0",
```
JamieB-gu, line 58: Why `git+` rather than `github:` out of curiosity?

abeddow91: Afraid I'm not sure on that. Think this would be a question for @faresite ?

PR #1105, opened 2021-01-22
Live 1407: Interview Standfirst Padding

abeddow91:
## Why are you doing this?
The Interview template has full width Header Image and Headline on mobile. This PR removes the padding from the header and adds it into Standfirst

## Changes

- Remove padding from header
- Add padding to standfirst
- Add style switch for template type

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/editions/header.tsx
+++ b/src/components/editions/header.tsx
@@ -55,7 +55,7 @@ const AnalysisHeader: FC<HeaderProps> = ({ item }) => (
 );
 
 const InterviewHeader: FC<HeaderProps> = ({ item }) => (
-	<header css={headerStyles}>
+	<header>
 		<HeaderImage item={item} />
 		<Headline item={item} />
 		<Standfirst item={item} />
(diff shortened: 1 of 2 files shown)


src/components/editions/header.tsx
```diff
@@ -55,7 +55,7 @@ const AnalysisHeader: FC<HeaderProps> = ({ item }) => (
 );
 
 const InterviewHeader: FC<HeaderProps> = ({ item }) => (
-	<header css={headerStyles}>
```
joecowton1, line 58: Just wonder if should rename the old `headerStyles` css to `headerPadding` or something to make clear what its doing?

joecowton1: or `sidePadding` even?

abeddow91: Yeah I agree assuming we are done adding default header styles? Otherwise headerPadding might be a bit narrow.

joecowton1: Good point - I'd imagine from here things are only going to get less generic? Thats said for the sake of not undoing things, If we leave it as for now maybe then and we can change down the road if it makes sense?

joecowton1 (approved):

src/components/editions/header.tsx
abeddow91: Gotcha

PR #1109, opened 2021-01-25
LIVE-1412: update analysis standfirst and byline spacing

joecowton1:
## Why are you doing this?

Updating Analysis template styles.

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/editions/byline.tsx
+++ b/src/components/editions/byline.tsx
@@ -7,9 +7,10 @@ import { body, headline } from '@guardian/src-foundations/typography';
 import type {
 	FontStyle,
 	FontWeight,
+	LineHeight,
 } from '@guardian/src-foundations/typography/types';
-import { Design } from '@guardian/types';
 import type { Format } from '@guardian/types';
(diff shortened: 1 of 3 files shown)


src/components/editions/byline.tsx
```diff
@@ -91,6 +105,18 @@ const bylineSecondaryStyles = (large?: boolean): SerializedStyles => css`
 		: standardTextStyles('italic', 'light')}
 `;
 
+const getStyles = (format: Format, kickerColor: string): SerializedStyles => {
+	if (format.display === Display.Showcase) {
```
abeddow91, line 109: Should we set out the Format order? Most Interviews have a Showcase display so Interview will need to be above Showcase.

abeddow91 (commented):
One small concern about Format ordering but otherwise looks good!

src/components/editions/byline.tsx
joecowton1: Good shout! Should be fixed now.

abeddow91 (approved):

PR #1112, opened 2021-01-25
Live 1406: interview series

abeddow91:
## Why are you doing this?
The Editions Interview template has the Series component on the Header Image. This PR makes that change

## Changes

- Add Series/Kicker to header image if it's type Interview
- Add interview specific styling to series

## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |


joecowton1:
Looks good - only concern is whether we've got a tiny gap between the kickr / image and headline, and if so can we add a negative value somewhere to offset that?

Screenshot 2021-01-25 at 12 24 13

abeddow91:
@joecowton1 - Nice spot, have updated to close gap

joecowton1 (approved):

PR #1113, opened 2021-01-25
LIVE-1000: Editions-Rendering Lightbox Message

faresite:
# Editions-rendering Lightbox Message

Follow up to this [Renditions PR](https://github.com/guardian/renditions/pull/3)

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "@guardian/src-text-area": "^2.8.1",
     "@guardian/src-text-input": "^2.7.1",
     "@guardian/types": "github:guardian/types#semver:1.1.0",
-    "@guardian/renditions": "git+https://github.com/guardian/renditions.git#0.1.0",
+    "@guardian/renditions": "git+https://github.com/guardian/renditions.git#0.1.1",
     "@types/jsdom": "^16.2.5",
     "@types/uuid": "^8.3.0",
     "@types/webpack-node-externals": "^2.5.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -55,7 +55,7 @@
     "@guardian/src-text-area": "^2.8.1",
     "@guardian/src-text-input": "^2.7.1",
     "@guardian/types": "github:guardian/types#semver:1.1.0",
-    "@guardian/renditions": "git+https://github.com/guardian/renditions.git#0.1.0",
+    "@guardian/renditions": "git+https://github.com/guardian/renditions.git#0.1.1",
```
JamieB-gu, line 58: Why the git URL (`git+`) rather than the GitHub one (`github:`) out of curiosity?

JamieB-gu (approved):

package.json
faresite: I think it's usually added before the protocol, unless the protocol is `git://`. 
`npm init` formats it this way when you init with a git repo.

faresite: https://docs.npmjs.com/cli/v6/configuring-npm/package-json#git-urls-as-dependencies

JamieB-gu: Is there a reason to prefer that over the `github:` URL? The next section in that docs page ^: https://docs.npmjs.com/cli/v6/configuring-npm/package-json#github-urls

faresite: Lol didn't know about this and that was just the next paragraph down 🙈
Looks like the github URL usage is as simple as `user/repo#...` to which I might only ever so slightly prefer having an explicit github url so it doesn't confuse people where this is being pulled from (i.e. looking for it in the npm registry). 
Do you have a preference @JamieB-gu ?

PR #1116, opened 2021-01-27
Add header video component to editions rendering

frankie297:
## Why are you doing this?
Adds a header video component to editions rendering, to match current styling on the editions app

Paired with @faresite 

## Screenshots

| Before | After |
| --- | --- |
|  |  |


src/components/editions/headerMedia.tsx
```diff
@@ -115,6 +129,19 @@ const HeaderImage: FC<Props> = ({ item }) => {
 					<StarRating item={item} />
 				</figure>
 			);
+		} else if (media.video.atomId !== undefined) {
```
joecowton1, line 132: Could we use `MainMediaKind.video` here?

src/video.ts
```diff
@@ -22,6 +24,7 @@ const parseVideo = (element: BlockElement, atoms?: Atoms): Option<Video> => {
 	}
 
 	const { posterUrl, duration, assets, activeVersion } = atom.data.media;
+	const title = atom.data.media.title
```
joecowton1, line 27: Can we just destructure title here from `atom.data.media` as we've done with the others on the line above?

src/video.ts
```diff
@@ -34,6 +37,8 @@ const parseVideo = (element: BlockElement, atoms?: Atoms): Option<Video> => {
 		posterUrl,
 		videoId,
 		duration: duration?.toNumber(),
+		atomId: id,
+        title
```
joecowton1, line 41: Indent here should be inline with the above

frankie297: Strange that the linter didn't pick this up 🤔

src/components/editions/headerMedia.tsx
frankie297: We tried evaulating `MainMediaKind.video && media.video.atomId` , but we was getting an error to say we were doing an unneccessary condition check, it's because `MainMediaKind` only has `Image` or `Video` and we are already checking for Image above so it can only be a Video.

src/video.ts
frankie297: 👍

src/components/editions/headerMedia.tsx
joecowton1: ah ok, in that case can it just be an `else`?

faresite: I believe `atomId` could be undefined even if we have MainMediaKind.video (following a capi type)

src/video.ts
faresite: I thought we tried that 🤔

src/components/editions/headerMedia.tsx
frankie297: Hmm not sure if it can be, because `atomId` is optional, `else` wouldn't handle if `atomId` is missing?

faresite:
@frankie297 let's get rid of the `package-lock.json` update in this PR? It's unrelated.

src/video.ts
frankie297: For some reason it works now

src/components/editions/headerMedia.tsx
joecowton1: ah ok, so you can have a video but no `atomId`?

In that case stick with what you have 👍

faresite: Although looking at it again after [this line](https://github.com/guardian/apps-rendering/blob/25a0994502d587310b8e17c8b3c88fbc6dfbb03f/src/video.ts#L19) we can pretty much depend on atomId always being there, atomId doesn't really need to be optional in the Video interface. What do you think @frankie297? shall we update the type?

frankie297: Yeh I think you are right, lets update the type

src/video.ts
```diff
@@ -21,19 +23,27 @@ const parseVideo = (element: BlockElement, atoms?: Atoms): Option<Video> => {
...
+		assets,
+		activeVersion,
+		title,
+	} = atom.data.media;
 	const videoId = assets.find(
 		(asset) => asset.version.toNumber() === activeVersion?.toNumber(),
 	)?.id;
 
-	if (!posterUrl || !videoId) {
+	if (!posterUrl || !videoId || !id) {
```
joecowton1, line 37: Slightly pedantic (sorry!) but if we call this `atomId` rather than just `id` throughout it makes it a bit easier to follow. Had to go back through the file to see where it was coming from.

joecowton1 (approved):

PR #1117, opened 2021-01-28
Live 1408: Interview Header Background Colour

abeddow91:
## Why are you doing this?
Adds coloured background for Interview template to headline and standfirst only as per designs. 
NB To reduce the PR size, this PR breaks styling on the image, byline and lines component. These will be resolved in separate pulls.

## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |


joecowton1 (approved):
Looks good!

PR #1120, opened 2021-01-28
LIVE 1408: Full Width Header Image

abeddow91:
## Why are you doing this?

Some templates require a full width image in the header. This PR add styles to the image, container, and caption for rendering a full width image.

## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |

### DIFF
--- a/src/components/editions/article.tsx
+++ b/src/components/editions/article.tsx
@@ -65,6 +65,13 @@ const headerBackgroundStyles = (item: Format): SerializedStyles => css`
 	background-color: ${headerBackgroundColour(item)};
 `;
 
+const getSectionStyles = (item: Format): SerializedStyles[] => {
+	if (item.design === Design.Interview) {
+		return [];
+	}
+	return [headerStyles, articleStyles];
+};
+
 const Article: FC<Props> = ({ item }) => {
 	if (item.design === Design.Live) {
 		return <p>Not implemented</p>;
@@ -74,7 +81,7 @@ const Article: FC<Props> = ({ item }) => {
 		<main>
 			<article>
 				<div css={headerBackgroundStyles(item)}>
-					<section css={[headerStyles, articleStyles]}>
+					<section css={getSectionStyles(item)}>
 						<Header item={item} />
 					</section>
 				</div>
--- a/src/components/editions/header.tsx
+++ b/src/components/editions/header.tsx
@@ -1,6 +1,9 @@
 // ----- Imports ----- //
 
 import { css } from '@emotion/core';
+import type { SerializedStyles } from '@emotion/core';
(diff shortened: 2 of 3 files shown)


src/components/editions/headerImage.tsx
```diff
@@ -77,6 +112,15 @@ const sizes: Sizes = {
 	default: '100vw',
 };
 
+const fullWidthSizes: Sizes = {
+	mediaQueries: [],
```
joecowton1, line 116: Should we be using this `mediaQueries` input to add dynamic image sizes for the different breakpoints? Even if it's just tablet and mobile.

abeddow91: It's full width across all devices for Interview so I don't think there would be break point sizes?

joecowton1: My understanding is that the `Img` component pulls in a series of sources for the `<picture />` element based on the values we provide to `mediaQueries`.

So for example when we give it an input of:

```
	mediaQueries: [
		{ breakpoint: 'tablet', size: '740px' },
		{ breakpoint: 'wide', size: '980px' },
	],
```

The browser will use the picture element to automatically load in the appropriate sized image for that breakpoint. If we only provide a `default` value we don't get that.

Appreciate that given this is a template it's only loaded once and therefore we might not get the benefits of that but just seems in keeping with the way the `Img` component is intended to be used.

joecowton1: TLDR: the breakpoint sizes relate to the size of image we're loading within the <picture /> element rather than anything to do with layout.

abeddow91: Gotcha - will update!

abeddow91: Gotcha - will update!

src/components/editions/headerImage.tsx
```diff
@@ -113,7 +113,10 @@ const sizes: Sizes = {
 };
 
 const fullWidthSizes: Sizes = {
-	mediaQueries: [],
+	mediaQueries: [
+		{ breakpoint: 'tablet', size: '100vw' },
+		{ breakpoint: 'wide', size: '100vw' },
+	],
 	default: '100vw',
```
JamieB-gu: If these are all `100vw` do you need anything beyond the `default`?

joecowton1: It was my suggestion to add breakpoint values but had envisaged as `px`s - if 100vw suffices then, no, there's no need for anything more than the default.

abeddow91: Yep, the image dimensions are derived from `size` so px fixes the width which we don't want here. Will revert the change!

PR #1127, opened 2021-01-28
LIVE-1472: only render finished templates

joecowton1:
## Why are you doing this?

Only render finished templates.


abeddow91 (approved):
Looks good!

PR #1128, opened 2021-01-28
Live 1408: Interview Byline

abeddow91:
## Why are you doing this?
This pr adds a bottom border to the Editions byline component.

## Changes

- Adds borderWidthStyles 
- Adds bottom border to byline for Interview
- Replaces px with remSpace in byline

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/editions/byline.tsx
+++ b/src/components/editions/byline.tsx
@@ -19,7 +19,7 @@ import type { FC, ReactNode } from 'react';
 import { getThemeStyles } from 'themeStyles';
 import EditionsAvatar from './avatar';
 import { ShareIcon } from './shareIcon';
-import { borderWidthStyles, wideArticleMargin } from './styles';
+import { borderWidthStyles, wideArticleMargin, articleWidthStyles } from './styles';
 
 // ----- Styles ----- //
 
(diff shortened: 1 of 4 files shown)


joecowton1:
Are we still missing a border-right on the byline? 😬

abeddow91:
Thanks for raising @joecowton1. Interview designs are inconsistent for right byline borders but I've added here as this seems to be the case across the rest of the templates and I suspect omitting the right border on the  wide device template is an mistake.

src/components/editions/byline.tsx
```diff
@@ -18,26 +19,28 @@ import type { FC, ReactNode } from 'react';
...
+		box-sizing: border-box;
+		margin-left: ${remSpace[6]};
+		padding-left: ${remSpace[3]};
+		padding-right: ${remSpace[3]};
 	}
 
 	${from.wide} {
-		margin-left: ${wideArticleMargin}px;
+		margin-left: ${remSpace[24]};
+		left: ${remSpace[12]};
```
joecowton1, line 39: This looks a bit suspect -  can we not just use the margin-left at `9rem` or even the value from the `styles` files that we've used elsewhere?

abeddow91: Think this was just to favour remSpace but agreed, looks gnarly. Will revert this!

joecowton1 (approved):

PR #1133, opened 2021-02-01
LIVE-1473: Editions Gallery

faresite:
## Editions Gallery

This PR implements the Editions gallery template ([Figma file](https://www.figma.com/file/7BWzTDoChAGQKCM4iC27aC/Article-Type-Picker?node-id=1%3A9)) and includes a few extras:
- header caption fixes for full width header images
- fix styled jsx in renderer
- full height article body

### Notes:
- When the caption description gets very long on tablet/wide, I've implemented the first figma design where the caption pushes the next image down. Until design can decide how exactly it needs to be implemented.
- The designs don't include credit, will follow up with design about those but shouldn't be a blocker for this PR.

### Screenshots:

| Mobile |
| --- |
|  |

| Tablet | Wide |
| --- | --- |
|  |  |

| Unreasonably long captions |
| --- |
|  |

### DIFF
--- a/src/components/editions/headerImage.tsx
+++ b/src/components/editions/headerImage.tsx
@@ -56,6 +56,10 @@ const captionStyles = css`
 
 const fullWidthCaptionStyles = css`
 	width: 100%;
+
+	${from.tablet} {
+		width: 100%;
+	}
 `;
 
 const getStyles = (format: Format): SerializedStyles => {
--- a/src/components/editions/headerImage.tsx
+++ b/src/components/editions/headerImage.tsx
@@ -41,7 +41,7 @@ const styles = css`
 const fullWidthStyles = css`
 	margin: 0;
 	position: relative;
-	width: 100vw;
+	width: 100%;
 `;
 
 const captionStyles = css`
@@ -55,11 +55,7 @@ const captionStyles = css`
 `;
 
 const fullWidthCaptionStyles = css`
-	width: 100vw;
-
-	${from.wide} {
-		width: 100vw;
-	}
+	width: 100%;
 `;
 
 const getStyles = (format: Format): SerializedStyles => {
--- a/src/components/editions/headerImage.tsx
+++ b/src/components/editions/headerImage.tsx
@@ -63,11 +63,17 @@ const fullWidthCaptionStyles = css`
 `;
 
 const getStyles = (format: Format): SerializedStyles => {
-	return format.design === Design.Interview ? fullWidthStyles : styles;
+	switch (format.design) {
+		case Design.Interview:
+		case Design.Media:
+			return fullWidthStyles;
+		default:
+			return styles;
+	}
 };
 
 const getCaptionStyles = (format: Format): SerializedStyles => {
-	return format.design === Design.Interview
+	return format.design === Design.Interview || format.design === Design.Media
 		? fullWidthCaptionStyles
 		: captionStyles;
 };
@@ -76,7 +82,7 @@ const getImageStyle = (
 	{ width, height }: Image,
 	format: Format,
 ): SerializedStyles => {
-	if (format.design === Design.Interview) {
+	if (format.design === Design.Interview || format.design === Design.Media) {
 		return css`
 			display: block;
 			width: 100%;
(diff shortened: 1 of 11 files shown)


joecowton1 (commented):
This looks great - good work!!

Couple of little things:

I wonder if some of the slight differences in the designs are just errors? Eg the lack of a border-right on the headline etc. Might be worth adding to the list for Ben if Anna has shared that with you?

Also - can we maybe break up PR's by individual tasks where possible? I know I was guilty of the odd oversized PR early on but we've been trying really hard to keep them small as possible lately. It makes it so much easier to review when they're smaller!

We've been splitting up tasks into subtasks and doing PR's per each of those which seems to work well. It also reduces the potential of conflicts like we discussed recently. 🙂

faresite:
> I wonder if some of the slight differences in the designs are just errors? Eg the lack of a border-right on the headline etc. Might be worth adding to the list for Ben if Anna has shared that with you?

@joecowton1 Good shout! I've added it to the spreadsheet. In case Ben's focusing on other design tasks, we can review this as is for now so the PR doesn't go stale.

> can we maybe break up PR's by individual tasks where possible?

Definitely, I can sometimes go overboard with my PRs 🙈 will keep that in mind. 
The extras are really tiny so I'll comment these so it's easier to review for now 👍

> We've been splitting up tasks into subtasks

How would you split up this PR? Header / body separate, per breakpoint or something else?

src/renderer.ts
```diff
@@ -415,7 +416,7 @@ const captionElement = (format: Format) => (
 				children,
 			);
 		case 'A':
-			return h(
+			return styledH(
```
faresite, line 419: `React.createElement` doesn't support emotion's `css` the way JSX does. If you look at an apps-rendering gallery at the moment for example you'd see that the dom element doesn't have the correct styles and instead has a `css=[Object object]` attribute. This should fix that.

src/server/editionsPage.tsx
```diff
@@ -35,9 +35,12 @@ const docParser = JSDOM.fragment.bind(null);
...
 const styles = `
     ${pageFonts}
-
+	html {
+		height: 100%;
+	}
     body {
-        margin: 0;
+		margin: 0;
+		min-height: 100%;
```
faresite, line 43: There was an extra bit of white at the bottom, the gallery's black background revealed that.

src/components/editions/headerImage.tsx
```diff
@@ -55,19 +55,25 @@ const captionStyles = css`
 `;
 
 const fullWidthCaptionStyles = css`
-	width: 100vw;
+	width: 100%;
 
-	${from.wide} {
-		width: 100vw;
+	${from.tablet} {
+		width: 100%;
```
faresite, line 61: The header image was overflowing with 100vw as it included spacings/ this also shifted the camera icon.

joecowton1:
> > I wonder if some of the slight differences in the designs are just errors? Eg the lack of a border-right on the headline etc. Might be worth adding to the list for Ben if Anna has shared that with you?
> 
> @joecowton1 Good shout! I've added it to the spreadsheet. In case Ben's focusing on other design tasks, we can review this as is for now so the PR doesn't go stale.
> 
> > can we maybe break up PR's by individual tasks where possible?
> 
> Definitely, I can sometimes go overboard with my PRs 🙈 will keep that in mind.
> The extras are really tiny so I'll comment these so it's easier to review for now 👍
> 
> > We've been splitting up tasks into subtasks
> 
> How would you split up this PR? Header / body separate, per breakpoint or something else?

Thanks Fares - the comments are really helpful! I can see what you're doing but it's helpful to understand the motivation for these things too 👍 

Re splitting things, Anna has done a good example here:
https://theguardian.atlassian.net/browse/LIVE-989

Also agree we go ahead and merge this for now rather than wait for design clarifications :)

src/server/editionsPage.tsx
joecowton1: Good spot!

src/components/editions/headerImage.tsx
joecowton1: Nice, I'd seen that popping up last week.

src/renderer.ts
joecowton1: Great, thanks for the explainer!

joecowton1 (approved):

src/renderer.ts
JamieB-gu: Emotion takes over in the JSX-to-createElement pipeline so that it can handle the `css` prop before handing over to `React.createElement` (seen [here](https://github.com/guardian/apps-rendering/blob/5935a859732ecd4824827d1e4c9596d57bd8b8ce/webpack.config.ts#L101) in the config, which uses the [React 17 JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)).

PR #1134, opened 2021-02-01
LIVE-1481: add background colour selector

joecowton1:
## Why are you doing this?

Adds background colour for immersive template.


joecowton1:
Closing to as no longer required.

PR #1138, opened 2021-02-02
Limit Chromatic Runs To main And PRs

JamieB-gu:
## Why are you doing this?

Chromatic is currently running on every branch that's pushed. We probably only want to run it in one of two cases:

1. A PR is opened against `main`
2. A PR is merged to `main` (we want to run against `main` once the PR is merged)

This should limit our usage of the Chromatic API.

## Changes

- Only run chromatic on merges to `main` or PRs to `main`

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -2,6 +2,18 @@ name: 'Chromatic'
 
 on:
     push:
+        branches:
+            - main
+        paths-ignore:
+            - 'api-models/'
+            - 'project/'
+            - '*.sbt'
+            - '*.md'
+            - '*.yml'
+            - '.gitignore'
+    pull_request:
+        branches:
+            - main
         paths-ignore:
             - 'api-models/'
             - 'project/'


gtrufitt (commented):
I would have expected chromatic to ignore this PR?

faresite (dismissed):
good one 👍

faresite:
> I would have expected chromatic to ignore this PR?
@gtrufitt  I think until you merge this PR it won't take effect?

gtrufitt:
> > I would have expected chromatic to ignore this PR?
> 
> 
> 
> @gtrufitt  I think until you merge this PR it won't take effect?

I thought the GitHub actions worked against the PR's branch? So I would have expected the ignore rule for .yml to take effect here - maybe it does, it says 'no visual changes' but not ignored 🤔

JamieB-gu:
> So I would have expected the ignore rule for .yml to take effect here

The `paths-ignore` rules are not new for this PR, they've always applied. However, I don't think that actually matters in this case because the rule is specifically `*.yml` which [only targets](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet) `.yml` files in the TLD - so it won't ignore `.yml` files in any subdirectories (i.e. `.github`). If we wanted it to include those we'd do `**.yml` I believe?

oliverlloyd:
I think this might affect how Chromatic wants to work. To be honest, I don't fully understand this but my impression is that you want chroamtic to snapshot your branches

https://www.chromatic.com/docs/branching-and-baselines

faresite:
@gtrufitt Sorry just re-read your comment .. my after lunch brain 🥴. I would have expected this PR _not_ to be ignored by Chromatic because it's a PR to `main`.

oliverlloyd:
More here: https://www.chromatic.com/docs/github-actions#recommended-configuration-for-build-events

JamieB-gu:
> I think this might affect how Chromatic wants to work

Surely you only need to run Chromatic on branches that you intend to merge, with that intention defined as opening a PR?

oliverlloyd:
> > I think this might affect how Chromatic wants to work
> 
> Surely you only need to run Chromatic on branches that you intend to merge, with that intention defined as opening a PR?

I fairly routinely create PRs, check my diffs, and sometimes havve to push changes to fix the bugs Chromatic found. I think this workflow would then not capture my changes in the baseline. I suspect that might be the problem?

joecowton1 (approved):

JamieB-gu:
> I think this workflow would then not capture my changes in the baseline

I'm not sure I understand this 🤔. If you've made a PR then this workflow will run on the branch for that PR - it only won't run on branches that haven't had a PR opened yet.

> I suspect that might be the problem?

What problem? 🐵

oliverlloyd:
> What problem? 🐵

This one: https://www.chromatic.com/docs/github-actions#recommended-configuration-for-build-events

If I read that right, and I'm only 70% sure here, it is saying that the Chromatic build will happen when the PR is created and _not_ get rerun for subsequent push events to the same PR. So when subsequent diffs are run, they might be comparing against an out of date set of snapshots?

JamieB-gu:
> it is saying that the Chromatic build will happen when the PR is created and _not_ get rerun for subsequent push events to the same PR

I think the `pull_request` event gets fired every time there are pushes to the PR branch - GH refer to these as `synchronize` events? This is backed up by the fact that our validation action re-runs any time you push new commits to a branch _after_ opening the PR. The Chromatic build should fire on new commits to a branch in the same way.

My reading of the docs you linked suggest that `pull_request` events _on their own_ are not enough, because they won't include the merge commit, and therefore Chromatic won't run against `main` and generate the required baselines. That's why I've _also_ included the `push` event trigger for `main`, as I believe GH counts merges as "pushes" to `main`, and therefore the Chromatic workflow should run against `main` whenever this happens.

oliverlloyd:
> My reading of the docs you linked suggest that `pull_request` events _on their own_ are not enough, because they won't include the merge commit, and therefore Chromatic won't run against `main` and generate the required baselines. That's why I've _also_ included the `push` event trigger for `main`, as I believe GH counts merges as "pushes" to `main`, and therefore the Chromatic workflow should run against `main` whenever this happens.

Riiight. I think I see what you mean now. This is actually interesting. I'll be bugging you in a few weeks to see how this worked out and if you're satisfied then we will steal it for some other repos!

PR #1142, opened 2021-02-03
LIVE-997: update widths

joecowton1:
## Why are you doing this?

Fixes Immersive layout and moves Immersive above Interview type in pecking order.

## Changes

- Make all background-colour blacks `neutral[7]`
- Move `Immersive` about `Interview` in terms of priority
- Update margins, widths etc to fit layout.
- Remove all `Immersive` white backgrounds until we receive clarity from design

## Screenshots

| Before | After |
| --- | --- |
|  |  |

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/editions/byline.tsx
+++ b/src/components/editions/byline.tsx
@@ -22,7 +22,9 @@ import { ShareIcon } from './shareIcon';
 import {
 	articleWidthStyles,
 	borderWidthStyles,
+	tabletImmersiveWidth,
 	wideArticleMargin,
+	wideImmersiveWidth,
 } from './styles';
 
 // ----- Styles ----- //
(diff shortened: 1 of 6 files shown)


src/components/editions/headline.tsx
```diff
@@ -53,9 +53,8 @@ const styles = (format: Format): SerializedStyles => css`
 	box-sizing: border-box;
 	border-top: 1px solid ${border.secondary};
 	padding-bottom: ${remSpace[4]};
+	padding-right: ${remSpace[2]};
 	margin: 0;
-
-	${articleWidthStyles}
```
faresite, line 58: This breaks the gallery headline width

src/components/editions/byline.tsx
```diff
@@ -160,14 +165,14 @@ const bylineSecondaryStyles = (
...
-	}
-
 	if (format.display === Display.Immersive) {
 		return css(styles(kickerColor), immersiveStyles);
 	}
 
+	if (format.design === Design.Interview) {
+		return css(styles(kickerColor), interviewStyles);
+	}
+
```
faresite, line 175: Since the ordering matters, maybe add a comment here and in the other place you're reordering in case someone in the future wants to rearrange?

src/components/editions/headline.tsx
faresite: Adding it to `galleryStyles` the way you've added it to interviewStyles should fix this.

joecowton1: Done.

src/components/editions/byline.tsx
joecowton1: Good idea 👍

faresite (approved):
Nice 👍

PR #1143, opened 2021-02-04
LIVE-1547: add editions specific caption component

joecowton1:
## Why are you doing this?

Creates an Editions specific HeaderImageCaption component with editions styling.

## Changes
- fix svg positioning
- fix caption height
- fix caption text layout

## Screenshots

| Before | After |
| --- | --- |
|  |  |

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/editions/headerImageCaption.tsx
+++ b/src/components/editions/headerImageCaption.tsx
@@ -1,0 +1,116 @@
+import type { SerializedStyles } from '@emotion/core';
+import { css } from '@emotion/core';
+import { remSpace } from '@guardian/src-foundations';
+import { from } from '@guardian/src-foundations/mq';
+import { brandAlt, neutral } from '@guardian/src-foundations/palette';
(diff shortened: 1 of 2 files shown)


src/components/editions/headerImageCaption.tsx
```diff
@@ -0,0 +1,116 @@
...
+): SerializedStyles => css`
+	summary {
+		text-align: center;
+		background-color: ${iconBackgroundColor
+			? iconBackgroundColor
+			: brandAlt[400]};
+		width: 34px;
+		height: 34px;
+		position: absolute;
+		bottom: 12px;
```
abeddow91, line 26: Any reason for px here over remSpace or rem?  Should we use remspace where available and if not, fall back to rem for consistency?

joecowton1: Good spot! Updated.

abeddow91 (approved):
Question on px v rem/remspace but otherwise all looks good!

PR #1144, opened 2021-02-04
Live 1287 use secure file parse images

dskamiotis:
## Why are you doing this?

Images should be parsed ideally using `secureFile` from assets, upon call to CAPI.
Currently image parsing only constructs the image URL using the `file` property.

## Changes

- Prioritised the parsing to be based on `secureFile`
- Included `file` url as a back up in case https for secureFile is unavailable (null/undefined) using Nullish coalescing

NOTE: in checking on local host articles return `undefined` for secureFile


package-lock.json
```diff
@@ -20984,6 +20992,7 @@
           "resolved": "https://registry.npmjs.org/to-regex-range/-/to-regex-range-2.1.1.tgz",
           "integrity": "sha1-fIDBe53+vlmeJzZ+DU3VWQFB2zg=",
           "dev": true,
+          "optional": true,
           "requires": {
             "is-number": "^3.0.0",
             "repeat-string": "^1.6.1"
```
JamieB-gu, line 20998: Don't think this file needs to be included in these changes?

JamieB-gu (commented):
Looks good!

frankie297 (approved):

package-lock.json
dskamiotis: @JamieB-gu this lock file manifested itself in the git push
All I did was npm i and perhaps the merge from main?

JamieB-gu:
> Delete package-lock.json
I think you may just want to remove the changes rather than deleting the file entirely 🙃

dskamiotis:
> > Delete package-lock.json
> 
> I think you may just want to remove the changes rather than deleting the file entirely 🙃

🤦‍♂️ of course, Ill do as git in VS Code

JamieB-gu:
Fixed in #1147?

dskamiotis:
> Fixed in #1147?

Yes perhaps a bit overkill but was the surest way since i deleted the file lol 👍

PR #1145, opened 2021-02-04
LIVE-1575: add Headline story

joecowton1:
## Why are you doing this?

Adds Storybook stories for all formats in `Headline` component

## Changes

- add stories
- refactor `Headline` component to use more descriptive function names and put styles into sections

##

To test run storybook and check `editions/Headline` stories


src/components/editions/header.tsx
```diff
@@ -7,7 +7,7 @@ import { from } from '@guardian/src-foundations/mq';
 import { Design, Display } from '@guardian/types';
 import Byline from 'components/editions/byline';
 import HeaderMedia from 'components/editions/headerMedia';
-import Headline from 'components/editions/headline';
```
abeddow91, line 10: What are your thoughts on having an index.js file in each sub folder that exports the component file? This way we won't need to update the import paths and it keeps the paths cleaner to read.

joecowton1: Yeah I'm into this I think. @faresite if you're ok with the idea we can go with that?

faresite: makes sense 👍

joecowton1: Updated 👍

src/components/editions/headline/headline.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+			theme: selectPillar(Pillar.News),
+		}}
+	/>
+);
+
+const Review = (): ReactElement => (
+	<Headline
+		item={{
+			...review,
+			...isImmersive(),
```
faresite, line 64: Don't think it's related to this PR but can a Review be immersive? Seems like the star ratings are broken in immersive.

joecowton1: Thats a good question - I think not. I'll remove it to avoid any confusion.

abeddow91 (approved):

PR #1146, opened 2021-02-05
LIVE-994: Editions Cartoons

faresite:
# Cartoons

Cartoons come into articles in two ways:
- The most common and "proper" way is through a tag `type/picture` and `type/cartoon`. This passes the image as a header main media.
- Less common is via an Interactive iframe. 

This PR focuses on dealing with the proper way, however in case of the interactive cartoon I've decided to open up CSPs and deal with the iframe resizing as well. Just in case it appeared in the Editions app. 

| mobile |
| --- |
|  |

| tablet | wide |
| --- | --- |
|  |  |


joecowton1 (approved):

src/components/editions/header.tsx
```diff
@@ -232,7 +251,11 @@ const renderArticleHeader = (item: Item): ReactElement<HeaderProps> => {
 	} else if (item.design === Design.Comment) {
 		return <CommentHeader item={item} />;
 	} else if (item.design === Design.Media) {
-		return <GalleryHeader item={item} />;
+		return isPicture(item.tags) ? (
+			<PictureHeader item={item} />
+		) : (
+			<GalleryHeader item={item} />
+		);
```
JamieB-gu, line 258: If we need to differentiate "picture" (are these _just_ cartoons?) articles I think it would be preferable to capture that in the `Format` type up front and derive styles from that. The philosophy of AR is that we split the request into two steps:

1) *Parser* - parse the complexity of the CAPI response into a platform specific data type (`Item`), and crucially parse anything that affects the overall style of a page into the `Format` type (which lives within `Item`).

2) *Renderer* - pass the data type built in step 1 to a collection of React components that generate the UI, and specifically use `Format` to determine style and layout variations.

As discussed offline, I'd like to understand the design differences between cartoons and other media articles to figure out whether we definitely need to update `Format` to accommodate them, or whether we can tweak the designs.

PR #1150, opened 2021-02-08
Live 993 letter header

dskamiotis:
## Why are you doing this?

Letter articles are not currently supported in Editions.
Also AR currently renders letters as comments and this separates the two types of articles

Because the Live App designs don't distinguish between the two, we expect Design.Comment and Design.Letter to apply the same styles. However, the existing code only knows to check for Design.Comment to apply said styles - any pieces that are now parsed as Design.Letter will fall through these checks. This means that Design.Letter pieces will likely have incorrect styles applied to them on the Live Apps.

## Changes

- Created letter head component
- added Letter styling to nested Header components for Editions
- Derived Letter type from tags
- Added handling for letter to apps rendering

paired with @abeddow91 

**NOTE**: We are awaiting the update to guardian/types in order to merge this, so the Letter type is included

Tested on : http://localhost:8080/lifeandstyle/2021/feb/04/not-fussy-about-the-perfect-pub?editions, http://localhost:8080/politics/2021/feb/04/labour-doesnt-need-to-wave-flags-to-be-patriotic?editions

### DIFF
--- a/src/__snapshots__/storyshots.test.ts.snap
+++ b/src/__snapshots__/storyshots.test.ts.snap
@@ -8357,6 +8357,1115 @@ exports[`Storyshots Editions/Article Interview 1`] = `
 </main>
 `;
 
+exports[`Storyshots Editions/Article Letter 1`] = `
+.emotion-0 {
+  height: 100%;
+}
+
+.emotion-1 {
+  min-height: 100%;
+  background-color: inherit;
+}
+
+.emotion-2 {
+  background-color: #FFFFFF;
+}
+
+.emotion-3 {
+  border-bottom: 1px solid #DCDCDC;
+}
+
+@media (min-width: 740px) {
+  .emotion-3 {
+    width: 526px;
+  }
+}
+
+@media (min-width: 980px) {
+  .emotion-3 {
+    width: 545px;
+  }
+}
+
+@media (min-width: 740px) {
+  .emotion-3 {
+    padding-right: 0.75rem;
+    border-right: 1px solid #DCDCDC;
+  }
+}
+
+@media (min-width: 740px) {
+  .emotion-3 {
+    margin-left: 24px;
+  }
+}
+
+@media (min-width: 980px) {
+  .emotion-3 {
+    margin-left: 144px;
+  }
+}
+
+.emotion-4 {
+  margin: 0;
+  position: relative;
+}
+
+@media (min-width: 980px) {
+  .emotion-4 {
+    width: 750px;
+  }
+}
+
+.emotion-5 {
+  width: 100vw;
+  height: calc(100vw * 0.6);
+  background-color: #F6F6F6;
+  color: #999999;
+  display: block;
+  display: block;
+  width: 100%;
+}
+
+@media (min-width: 740px) {
+  .emotion-5 {
+    width: 740px;
+    height: calc(740px * 0.6);
+  }
+}
+
+@media (min-width: 1300px) {
+  .emotion-5 {
+    width: 980px;
+    height: calc(980px * 0.6);
+  }
+}
+
+@media (min-width: 740px) {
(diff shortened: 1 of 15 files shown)


JamieB-gu (changes requested):
**TL;DR** We need to update all the code that doesn't specifically apply to Editions to check for `Design.Letter` alongside `Design.Comment`.

**Why?**

Both strands of Apps-Rendering (Editions and the Live Apps) use the same code in `item.ts` to create the `Item` object. By splitting `Letter` off from `Comment`, any `tone/letters` pieces that were previously parsed as `Design.Comment` will now be parsed as `Design.Letter`.

Because the Live App designs don't distinguish between the two, we expect `Design.Comment` and `Design.Letter` to apply the same styles. However, the existing code only knows to check for `Design.Comment` to apply said styles - any pieces that are now parsed as `Design.Letter` _will fall through_ these checks. This means that `Design.Letter` pieces will likely have incorrect styles applied to them on the Live Apps.

For an example, see the [`Byline` component](https://github.com/guardian/apps-rendering/blob/986336e5970154754f4f9a294961a4e17569667b/src/components/byline.tsx#L86). Previously a "letter" piece would have triggered the `Design.Comment` branch of this `switch` statement, and returned `commentStyles`. Now, because said piece has `Design.Letter`, that `switch` statement will fault through to the `default` branch and return `styles`.

cc @oliverlloyd @gtrufitt because DCR will likely hit the same issue (unless they decide to continue to parse `tone/letters` as `Design.Comment` and ignore `Design.Letter` of course - although that might cause us issues down the road...)

oliverlloyd:
We should add the logic to support `Design.Letter` in DCR at the same time as when we bump /types.

We will - I guess - also need to parse out the decision of when an article gets `Design.Letter` based on tags.

@gtrufitt sound about right?

gtrufitt:
> We should add the logic to support `Design.Letter` in DCR at the same time as when we bump /types.

Yup.

I am concerned that we're now significantly splitting off DesignType from https://github.com/guardian/content-api-scala-client/blob/63ca817bb8ad49eb8869603b74ec91247c0c6bfe/client/src/main/scala/com.gu.contentapi.client/utils/CapiModelEnrichment.scala#L19 - we should find out where this is used and applied and update it ASAP. I think it should stay in sync with the TS types or, better, we work out how to share these types.

joecowton1:
@dskamiotis whats the status of this PR? 

I know you said we're waiting on the latest types release - do we know when that's happening and can we leave a note here to say what's going on so this doesn't get confusing please.

dskamiotis:
> @dskamiotis whats the status of this PR?
> 
> I know you said we're waiting on the latest types release - do we know when that's happening and can we leave a note here to say what's going on so this doesn't get confusing please.

Thats correct Joe, thanks I will add to the description to make it clearer. We have a ticket for this in MSS.

joecowton1:
Thanks @dskamiotis 👍

dskamiotis:
> Thanks @dskamiotis 👍
@abeddow91 @groakland @joecowton1 
I had a bit of brain freeze, @JamieB-gu did indeed add `Letter` to types so i believe this is ok to merge.
@JamieB-gu Is it safe to merge now?

JamieB-gu:
> @JamieB-gu Is it safe to merge now?

They're in `main` on `@guardian/types` but not published yet. This is most likely why the build on this PR is failing.

JamieB-gu:
Waiting on #1176.

joecowton1 (commented):
I've rebased with the current `main` and everything is working nicely with the `@guardian/types` `v6.0.0`. 

I've also added a story for the Editions `Letter` header as that was missing. If we can give it a quick once over please I think that's ready to go.

JamieB-gu (approved):

PR #1155, opened 2021-02-08
LIVE-1576: add pullquote story

joecowton1:
## Why are you doing this?

Add a story for the `pullquote` component

## Changes

- Move to new folder
- Add story


src/renderer.ts
```diff
@@ -816,6 +816,7 @@ const renderEditions = (format: Format, excludeStyles = false) => (
 
 		case ElementKind.Pullquote: {
 			const { quote, attribution } = element;
+			console.log(element);
```
abeddow91, line 819: To be removed?

abeddow91 (changes requested):

src/renderer.ts
joecowton1: good spot - gone!

abeddow91 (approved):

PR #1159, opened 2021-02-11
LIVE-1533: add body story and remove interactives and richlink from r…

joecowton1:
## Why are you doing this?

Rich links and Interactive's aren't required in Editions.

## Changes

- remove rich links and interactives
- add article body to template stories (except media template)

### DIFF
--- a/src/fixtures/item.ts
+++ b/src/fixtures/item.ts
@@ -29,6 +29,19 @@ const bylineHtml: Option<DocumentFragment> = pipe2(
 	toOption,
 );
 
+const docFixture = (): DocumentFragment => {
+	const doc = new DocumentFragment();
+
+	const el = document.createElement('p');
+
+	el.innerText =
+		'Readers of Prospect magazine recently voted him the world’s fourth-best thinker. And right now he is thinking about 3 November, and whether the United States will reject or endorse Donald Trump. No one knows what will happen; not even West, not least because in the US he sees contradictions that even he can’t fully explain.';
+
+	doc.appendChild(el);
+
+	return doc;
+};
+
 const srcset =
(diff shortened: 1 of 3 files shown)


src/fixtures/item.ts
```diff
@@ -80,10 +81,119 @@ const mainMedia = {
...
 };
 
+const doc = new DocumentFragment();
+
+const el = document.createElement('p');
+
+el.innerText =
+	'Readers of Prospect magazine recently voted him the world’s fourth-best thinker. And right now he is thinking about 3 November, and whether the United States will reject or endorse Donald Trump. No one knows what will happen; not even West, not least because in the US he sees contradictions that even he can’t fully explain.';
+
+doc.appendChild(el);
```
faresite, line 91: just for Marie Kondo reasons, can we put this in a `docFixture()` function?

src/renderer.ts
```diff
@@ -819,21 +819,9 @@ const renderEditions = (format: Format, excludeStyles = false) => (
...
 		case ElementKind.LiveEvent:
 			return h(LiveEventLink, { ...element, key });
 
-		case ElementKind.Interactive:
-			return h(Interactive, {
-				url: element.url,
-				key,
-				title: element.alt,
-			});
-
```
faresite, line 836: I'm surprised TypeScript doesn't complaing that we didn't exhaust all options .. might be worth adding a default case returning `null` ?

joecowton1: Good shout, updated!

src/fixtures/item.ts
joecowton1: done :)

faresite (approved):
nice 👍

PR #1162, opened 2021-02-11
LIVE-1659: Update to Bridget 1.8.0

faresite:
# Bridget 1.8.0

In the latest release ([PR](https://github.com/guardian/bridget/pull/71)) We have renamed:
- `FrictionScreenReason` -> `PurchaseScreenReason`
- `launchFrictionScreen` -> `launchPurchaseScreen` 

Since we have removed the old types this basically could count as a breaking change, but we're not in production yet, and this shouldn't break it for beta users either (basically they won't get the purchase screen pop up). 
I've considered this better than using the `nativeThriftPackageVersion` and peppering our codebase with if and else before going to prod.


frankie297 (approved):
sounds sensible to me 👍

PR #1164, opened 2021-02-12
LIVE-1665: add viewports (with addon-essentials) to storybook 

joecowton1:
## Why are you doing this?

Adds viewports to Storybook for Small Mobile, Mobile, Tablet and Tablet Landscape.

This allows us to view any breakpoint at the click of a button - its already highlighted a few small layout bugs that will need addressing.

## Changes

- Update storybook config and add addon

## Screenshots

| Before | After |
| --- | --- |
|  |  |

| Before | After |
| --- | --- |
|  |  |
Screenshot 2021-02-12 at 12 00 55

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,3 +1,0 @@
-import { configure } from '@storybook/react';
-
-configure(require.context('../src', true, /\.stories\.tsx$/), module);
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,0 +1,57 @@
+// .storybook/preview.js
+
+import { breakpoints } from '@guardian/src-foundations';
+
+
+const customViewports = {
+    desktop: {
+      name: 'Desktop',
+      styles: {
+        width: breakpoints.desktop,
+        height: '963px',
+        },
+    },    
+    tablet: {
+      name: 'Tablet',
+      styles: {
+        width: breakpoints.tablet,
+        height: '963px',
+        },
+    },      
+    phablet: {
+      name: 'Phablet',
+      styles: {
+        width: breakpoints.phablet,
+        height: '963px',
+        },
+    },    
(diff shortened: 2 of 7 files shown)


faresite:
How about we add `@storybook/addon-essentials` instead? This is what I'm trying to use for the gallery image story as I also need the `backgrounds`

.storybook/config.js
```diff
@@ -1,3 +0,0 @@
-import { configure } from '@storybook/react';
-
-configure(require.context('../src', true, /\.stories\.tsx$/), module);
```
faresite, line 3: Is the configure from `@storybook/react` not needed then?

joecowton1: No not anymore 👍

.storybook/preview.js
```diff
@@ -0,0 +1,23 @@
...
+        width: '1024px',
+        height: '963px',
+        },
+    },
+};
+
+export const parameters = {
+    layout: 'fullscreen',
+    viewport: {
+        viewports: {
```
JamieB-gu, line 18: You may want to use our list of standard breakpoints [from Source](https://github.com/guardian/source/blob/main/src/core/foundations/src/breakpoints.ts).

JamieB-gu (commented):
Fancy!

FYI @SiAdcock @oliverlloyd @sndrs

.storybook/preview.js
joecowton1: Good call  - added :)

joecowton1:
> How about we add `@storybook/addon-essentials` instead? This is what I'm trying to use for the gallery image story as I also need the `backgrounds`

Added now

oliverlloyd:
Oh wow, there's so much great working happening in apps-rendering at the moment!

faresite (approved):

PR #1168, opened 2021-02-15
LIVE-1572: Gallery Image story

faresite:
# Gallery Image Story

In addition to the story, this PR includes fix for the storybook viewports.


.storybook/preview.js
```diff
@@ -1,57 +1,63 @@
...
 export const parameters = {
-    layout: 'fullscreen',
-    viewport: {
-        viewports: {
-            ...customViewports
-        }
-    },
-};
\ No newline at end of file
+	layout: 'fullscreen',
```
joecowton1, line 58: Don't think we need this - I've added it just to the article stories, everything else is fine without it.

joecowton1 (commented):
Needs that one parameter removing, otherwise nice work!

.storybook/preview.js
faresite: done 👍

joecowton1 (approved):

PR #1169, opened 2021-02-16
Re-Model Third-Party Embeds

JamieB-gu:
## Why are you doing this?

I'm re-thinking the way we model third-party embeds, which I hope will help with the upcoming embeds consent work. I think the best place to start in understanding these changes is the new `embed.ts` file.

The headline change here is that I've stopped treating "video" and "audio" embeds as separate elements - they now live under the `Embed` type along with "generic" embeds. I've also split them up into more specific kinds: `YouTube` and `Spotify`. You can see these types in the `embed.ts` file, along with their parsers.

I plan to expand this to include Tweets and Instagram embeds next, and then any other sources of "video" and "audio" embeds (please let me know if you're aware of any!). I also need to write some tests...

**Note:** I've also added a couple of combinator functions for `Result` in `lib.ts` (`resultFromNullable` and `resultMap3`), @frankie297 @johnduffell @dskamiotis @marjisound you may be interested in these 🙂. I plan to migrate them upstream to `types` to live [with the `Result` type definition](https://github.com/guardian/types/blob/6692d8762e28e1a82615da3bc5485db57e76f016/src/result.ts).

## Changes

- Remove `Audio` and `Video` standalone elements, put them under `Embed`
- Expand `Embed` to _all_ third-party embeds, not just generic ones
- Move responsibility for parsing embeds to `embed` from `bodyElement`
- Added new `embed` module to capture embed types and parsing
- Add `Embed` type capturing `YouTube`, `Spotify` and `Generic`
- Add parsing for new `Embed` type
- Temporarily incubate `resultFromNullable` and `resultMap3` in `lib`
- Add `parseIntOpt` for parsing ints from strings
- Move responsibility for rendering embeds out of `renderer.ts`
- Tidy up `Audio` and `Video` components
- Add new `EmbedComponent` for rendering embeds
- Add new `GenericEmbed` component for rendering generic embeds
- Bumped `@guardian/content-api-models` to `15.10.0`

## Screenshots

Should be no different 🤔.


frankie297:
> please let me know if you're aware of any!

TikTok? 😄

frj:
Currently DCR only renders video elements with  youtube, vimeo and facebook.

But there is definitely a long tail of 'sources' for video embed elements. Eg heres one from abc news:

http://content.guardianapis.com/commentisfree/2015/apr/24/ben-affleck-doesnt-want-to-feel-guilty-about-his-slave-owning-ancestors-who-would

Audioboo:
http://content.guardianapis.com/world/middle-east-live/live/2014/aug/14/iraq-crisis-us-says-it-has-broken-isis-seige-of-mount-sinjar-live-updates

Weirdly heres one with a source of youtube but with a iframe source from bbc.co.uk:
http://content.guardianapis.com/music/2014/feb/03/lso-pappano-peter-maxwell-davies-tenth-symphony-review?api-key=crier-v2-prod&show-blocks=all

Along with many others.

Obviously the generic embeds often contain video embeds as well.

Tony did some analysis of all of the contents of CAPI re embedded elements, this spreadsheet details articles with embedded elements their element type and the domain of the src field of the iframe in the embed:

https://docs.google.com/spreadsheets/d/10CsksVumot2hDkd28_vhR66c1DtIP09wDeyhyPT-GcA/edit#gid=1464055278

joecowton1 (approved):
Looks good - think this makes a lot of sense as an approach 👍

src/components/genericEmbed.tsx
```diff
@@ -0,0 +1,45 @@
...
+	color: ${text.supporting};
+`;
+
+interface Props {
+	embed: Generic;
+}
+
+const GenericEmbed: FC<Props> = ({ embed }) => (
+	<figure css={styles}>
+		<iframe srcDoc={embed.html} title={withDefault('Embed')(embed.alt)} />
```
frj, line 36: I realise this is WIP but in this case the iframe just takes on a default size that does not match the size of the embedded content.

For example:

http://localhost:8080/culture/2019/apr/27/idris-elba-marries-sabrina-dhowre-in-morocco

That is an instagram embed in an 'embed' type element.

Again as Im sure you know, the instagram js also gets blocked by the CSP.

JamieB-gu: I've done some work to handle the height issue in 42a6e61. This only works for some generic embeds (the ones that are already wrapped in an `iframe` with `height`s).

Generic instagram embeds having their JS blocked is already an issue on `main` - I'll look at it in a separate PR 👍.

JamieB-gu:
> TikTok? 😄

@frankie297 I think TikTok is a "generic" embed (`type: "embed"`), it doesn't have a custom "video" (`type: "video"`) or "audio" (`type: "audio"`) element.

PR #1172, opened 2021-02-16
LIVE-1685: Collect more coverage

faresite:
We currently only collect coverage from two files: `renderer.ts` and `item.ts`. 
To have realistic ambitions, instead of asking to cover all `.ts` files, I've explicitly selected (subjectively) the bulk of what would be nice to have covered by tests (excluding components since we're trialling storyshots for that, thanks @joecowton1 ).
All these files have a corresponding [Jira subtasks](https://theguardian.atlassian.net/browse/LIVE-1684).

Obviously this means the threshold had to be lowered, but hopefully we can work our way up little by little (maybe a certain percentage every health sprint).


joecowton1:
> We currently only collect coverage from two files: `renderer.ts` and `item.ts`.
> To have realistic ambitions, instead of asking to cover all `.ts` files, I've explicitly selected (subjectively) the bulk of what would be nice to have covered by tests (excluding components since we're trialling storyshots for that, thanks @joecowton1 ).

If it's possible to collect coverage from all`*.test.ts` files, what do we lose by doing that? Worry slightly that this way we have to remember to manually add each test file.

faresite:
> If it's possible to collect coverage from all*.test.ts files, what do we lose by doing that? Worry slightly that this way we have to remember to manually add each test file.

@joecowton1 I'm not an expert on the way [Istanbul](https://istanbul.js.org/) coverage works, but generally it parses the code into an AST and somehow instruments or injects code so it can know exactly which statements, branches or lines of code have been tested or not.. that's why it needs to know the code you are trying to test as opposed to the knowing the test files themselves (the `.test.ts` files). 
I could open up the coverage to all `.ts` files but then the coverage rate will probably get diluted and be meaningless if we don't exclude some files explicitly. 

I'm personally ok with adding each file manually (if at all) for now as I'm hoping we'll be focused on getting our coverage rate up to begin with. Once we're in a good place we can rethink where to collect coverage from.

joecowton1:
> @joecowton1 I'm not an expert on the way [Istanbul](https://istanbul.js.org/) coverage works, but generally it parses the code into an AST and somehow instruments or injects code so it can know exactly which statements, branches or lines of code have been tested or not.. that's why it needs to know the code you are trying to test as opposed to the knowing the test files themselves (the `.test.ts` files).
> I could open up the coverage to all `.ts` files but then the coverage rate will probably get diluted and be meaningless if we don't exclude some files explicitly.
> 
> I'm personally ok with adding each file manually (if at all) for now as I'm hoping we'll be focused on getting our coverage rate up to begin with. Once we're in a good place we can rethink where to collect coverage from.

Ah of course - thanks for clarifying. 

My feeling against this selective approach is that is slightly disingenuous - presumably we're just as interested in the number of files covered as the detail of the specific files, and the coverage log gives us a read out of both.

What I've always done previously is just to add say the src directory: could we do that, or even just sub dirs? Just something where its a bit looser than adding specific files.

@JamieB-gu what's your take on this?

faresite:
I really see your point and agree with you. Just to be clear this is an intermediary step to get us on the right track. If we just opened up the coverage to collect from `src/**/*.ts` we will also get a disingenuous (I like the word btw) stats unless we exclude code that is either unlikely for us to test in the near future at all, test helpers, or styles that are now covered by storyshots and unlikely for us to test them manually ... Files that could be excluded (not comprehensive list): `storybookHelpers.ts`, `helperTest.ts`, `helperTest.ts`, `editorialStyles.ts`, `editorialPalette.ts`, `themeStyles.ts`, `types/parserContext.ts` (just types), `server/appIdentity.ts`, `server/assets.ts`, `server/aws.ts`, `server/decoders.ts`, `server/server.ts`, `server/ssmConfig.ts`, `native/**/*`, `logger/**/*`, `fixtures/**/*` .

NB. Not saying these excluded files shouldn't be tested or are not testable (although some are). But realistically we're not gonna get there anytime soon and for us to have a measurable success at upping our coverage game it's nice to have a specific goal even if temporary (like 5% every health sprint). If and after achieving our goal we can get more holistic stats on our coverage.

Another option is to Marie Kondo our codebase and reorganise it 🍱 so all files under certain folders in `src` should be tested.

joecowton1 (approved):

PR #1175, opened 2021-02-18
[spike]: demo grid layout for header media

joecowton1:
## Why are you doing this?

Spike to look at using a CSS-Grid based approach for Editions Rendering.

Currently in AR we are using fixed widths based on the Figma designs control our layout.

This comes with a number of issues:
- Our designs not based on any kind of grid, merely where the designers see fit to place things.
- Our layouts are not dynamic, we can update the values by breakpoint but not in between breakpoints
- Rather than having a small number of components that handle layout values, we have a large number of width values buried within each component
- Often we end up setting a width on an individual sub-component 

While by no means a definitive solution, what this spike hopes to demonstrate is how with a few simple grid layout components we can have dynamic layers that improve consistency across our components while reducing the amount of code we use to control widths.

By controlling the number columns our layout components sit across - and allowing our subcomponents with to be entirely controlled by the layout components - we reach an elegant solution to many of the above problems.

Screenshot 2021-02-18 at 17 05 32

Its worth mentioning that in order for this to work properly we would need:
- Designers to use grids (although this can be approximated)
- A reasonable size refactor of existing Editions components to remove the hardcode width values 

This PR exists only as a demo of an idea and a starting point for discussion. I think we could and should be using grid based layouts where ever possible as the advantages are huge - let's find a way to make that happen.

** Its worth mentioning too that while I was finishing this off eariler @SiAdcock shared his latest Source Layout components which look fantastic and would probably do a very similar job - my next task is to look at those. **

## Changes

- Create grid components
- Refactor `HeaderMedia` component to fill container width as an example of how Grid can assist our layouts
- Remove `magic number` widths where possible

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/editions/grid/Column.tsx
+++ b/src/components/editions/grid/Column.tsx
@@ -1,0 +1,42 @@
+import type { SerializedStyles } from '@emotion/core';
+import { css } from '@emotion/core';
+import { from } from '@guardian/src-foundations/mq';
+import type { FC } from 'react';
+
+interface Props {
+	smStart?: number;
+	mdStart?: number;
+	lgStart?: number;
+	smSpan?: number;
+	mdSpan?: number;
+	lgSpan?: number;
+	className?: SerializedStyles;
+}
+
+export const Column: FC<Props> = ({
+	children,
+	smStart = 0,
+	mdStart = 0,
+	lgStart = 0,
+	smSpan = 12,
+	mdSpan = 12,
+	lgSpan = 12,
+	className,
+}) => {
+	const styles = css`
+		grid-column-start: ${smStart};
+		grid-column-end: span ${smSpan};
+
+		${from.tablet} {
+			grid-column-start: ${mdStart};
+			grid-column-end: span ${mdSpan};
+		}
+
+		${from.wide} {
+			grid-column-start: ${lgStart};
+			grid-column-end: span ${lgSpan};
+		}
+	`;
+
+	return <div css={[styles, className]}>{children}</div>;
+};
--- a/src/components/editions/grid/Grid.stories.tsx
+++ b/src/components/editions/grid/Grid.stories.tsx
@@ -1,0 +1,93 @@
+import { css } from '@emotion/core';
+import { border, remSpace } from '@guardian/src-foundations';
+import { from } from '@guardian/src-foundations/mq';
+import Byline from 'components/editions/byline';
+import HeaderMedia from 'components/editions/headerMedia';
+import Headline from 'components/editions/headline';
+import Standfirst from 'components/editions/standfirst';
+import type { FC } from 'react';
+import { interview, article as item } from '../../../fixtures/item';
+import Lines from '../lines';
+import { Column } from './Column';
+import Container from './Container';
+import Grid from './Grid';
+
+const video = {
+	kind: 0,
+	value: {
+		kind: 1,
+		video: {
+			posterUrl:
+				'https://media.guim.co.uk/032aa99755664104fbfc4cbe45cfae0243dce462/0_0_3972_2234/master/3972.jpg',
+			videoId: 'wD_bWOEuuoc',
+			duration: 59,
+			atomId: '26401ff7-24d0-4ba5-8882-2c32c2b379f0',
+			title:
+				'Super Bowl LV: Tom Brady MVP as Buccaneers beat Chiefs 31-9 – video report',
+		},
+	},
+};
+
+const borderStyles = css`
+	${from.tablet} {
+		border-right: 1px solid ${border.secondary};
+		padding-right: ${remSpace[3]};
+	}
+`;
+
+const lineStyles = css`
+	${from.tablet} {
+		margin-right: -${remSpace[3]};
+	}
+`;
+
+const Article: FC = () => (
+	<Container>
+		<Grid smCols={12} mdCols={12} lgCols={12}>
+			<HeaderMedia item={{ ...item }} />
+			<Column smSpan={12} mdSpan={8} lgSpan={8} className={borderStyles}>
+				<Headline item={item} />
+				<Standfirst item={item} />
+				<Lines className={lineStyles} />
+				<Byline item={item} />
+			</Column>
+		</Grid>
+	</Container>
+);
+
+const FullWidth: FC = () => (
+	<Container>
+		<Grid smCols={12} mdCols={12} lgCols={12}>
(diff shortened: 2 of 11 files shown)


oliverlloyd (commented):
So yes, big +1 for centralising layout concerns in general and trying to keep sub components ambilivant about things like width (unless relevant). I think the points you've listed are all well thought through and well done on the write up 👍 

One point. Where you've said

> Our layouts are not dynamic, we can update the values by breakpoint but not in between breakpoints

I can't speak for the app layouts, but for dotcom we only make layout changes at breakpoint change, not in-between; we have almost no responsive layout (I think images at certain mobile breakpoints are responsive but that's about it)

### Grid in general
I've actually only ever used css grids using named areas so I'm going to put the case for that approach here but be aware I'm utterly one-sided on this and it could well be that a more column based appraoch is right for your use case.

[Grid template areas](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Grid_Template_Areas) let you define _named_ areas of a page, with position and gutters manages by the grid, and then you drop content into them. My brain gets upset when I ask it to create a mental model of where things will appear when reading other grid code but using areas it is happy and the debugging experience is 😎 
Screenshot 2021-02-19 at 09 27 52

Screenshot 2021-02-19 at 09 26 32

#### What about IE11?
Grid tempalate areas are not supported in IE11 which is a hard blocker for a lot of projects and possible here too. For dotcom we accept that the layout won't be as dynamic for readers using IE11 and serve them a basic layout using [progressive enhancement](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/CSS_Grid_and_Progressive_Enhancement)

joecowton1:
Thanks for the response @oliverlloyd !

I'm intrigued to see how the `grid-template-areas` would work in practice with the Editions layout given that we're working almost entirely in one vertical column - there's not the same requirement of balancing the heights of containers in multiple columns. I've very tempted just to mock out an instance to test that though.

On the dynamic nature of the layouts: I guess this is really a question for design. If we can get over the first hurdle of actually putting the grids in place that will be a win at this stage!

I don't think IE11 should be an issue here this isn't my area of expertise so pls correct me if I'm wrong.

@SiAdcock would be interested to hear your views on any of the above too when you have a minute 🙂

faresite:
Great job @joecowton1! really like where this is going and a big +1 for getting rid of lot of the magic numbers 👍
Few comments:
- No need to worry about IE11, neither AR nor ER need to support it, we have that luxury.
- I don't see much point to `<Container />` ? I think `<Grid />` can take of what Container is trying to do and keep the entry level to using the system to a minimum.
- Do you ever see a `<Row />` component coming in handy too? It's true as you mentioned our layout is mostly in one column, but if we're to think into the future if we needed them, would that be much of a task to add it?
- Just as an API suggestion, if the column span doesn’t change across breakpoints and I want to use one span definition, would be nice to use it across as opposed to defaulting it with 12. For example if I asked for `<Column smSpan={6} />` I probably want it to span 6 columns for `md` and `lg` and not 12. If that makes sense.
- A side effect of adding a default of `span 12` is that if I don't explicitly set it, it will overflow the grid gap. To explain this better if I had these two components for example:
```jsx
<Grid smCols={12} mdCols={12} lgCols={12}>
	  <Column smStart={2}>
		  <h1 style={{ background: 'red' }}>One</h1>
	  </Column>
	  <Column>
		  <h1 style={{ background: 'green' }}>Two</h1>
	  </Column>
</Grid>
```
I would end up with this:
Screenshot 2021-02-23 at 15 27 27

src/components/editions/grid/Column.tsx
```diff
@@ -0,0 +1,42 @@
...
+	mdSpan?: number;
+	lgSpan?: number;
+	className?: SerializedStyles;
+}
+
+export const Column: FC<Props> = ({
+	children,
+	smStart = 0,
+	mdStart = 0,
+	lgStart = 0,
```
faresite, line 20: You probably mean `1` here as 0 is not a valid integer for `grid-column-start`

src/components/editions/grid/Grid.stories.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+const lineStyles = css`
+	${from.tablet} {
+		margin-right: -${remSpace[3]};
+	}
+`;
+
+const Article: FC = () => (
+	<Container>
+		<Grid smCols={12} mdCols={12} lgCols={12}>
```
faresite, line 46: I'm aware this is just an example but should it be `<Grid smCols={12} mdCols={8} lgCols={8}>` otherwise it will look like the fullwidth article?

joecowton1: So these are just the numbers of columns in each row, they should only be different if the designs reflect that.

Ie if on mobile the layout is 4 cols wide then the value should be 4. Its on the `Column` component that we set the widths.

Does that make sense?

joecowton1:
> * No need to worry about IE11, neither AR nor ER need to support it, we have that luxury.
Perfect.

> * I don't see much point to `<Container />` ? I think `<Grid />` can take of what Container is trying to do and keep the entry level to using the system to a minimum.

We could, yes. I think every example of Grids i've seen use some kind of `Container` but we can certainly investigate if thats necessary.

> * Do you ever see a `<Row />` component coming in handy too? It's true as you mentioned our layout is mostly in one column, but if we're to think into the future if we needed them, would that be much of a task to add it?

Yeah definitely, tbh I'm surprised I haven't added one 🤔 

> * Just as an API suggestion, if the column span doesn’t change across breakpoints and I want to use one span definition, would be nice to use it across as opposed to defaulting it with 12. For example if I asked for `<Column smSpan={6} />` I probably want it to span 6 columns for `md` and `lg` and not 12. If that makes sense.

This is a good point.

> * A side effect of adding a default of `span 12` is that if I don't explicitly set it, it will overflow the grid gap. To explain this better if I had these two components for example:

> I would end up with this:
> Screenshot 2021-02-23 at 15 27 27

Yeah good sport 👍  That needs work!

JamieB-gu:
@joecowton1 @faresite any updates on this PR?

joecowton1:
> @joecowton1 @faresite any updates on this PR?

We can probably just close it? Don't see anyone having the time to implement this now.

JamieB-gu:
👍

PR #1179, opened 2021-02-19
LIVE-1709: removed fixed widths on tablet

joecowton1:
## Why are you doing this?

Quick fix for a styling bug where in some situations the tablet header image would be wider than the viewport.

Also introduces `from.desktop` rather than `from.wide` for Editions components to correspond with Storybook viewport sizes.

@faresite I've butchered the `HeaderMedia` component, apologies. Please say if I've removed anything that needs to be there, particularly height values. When testing I couldn't find any need for them but I might be wrong. 

## Changes

- Make tablet image width dynamic

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -56,7 +56,7 @@ const customViewports = {
 
 export const parameters = {
 	viewport: {
-		defaultViewport: 'mobile',
+		defaultViewport: 'tablet',
 		viewports: customViewports,
 	},
 };
--- a/src/components/editions/article/index.tsx
+++ b/src/components/editions/article/index.tsx
@@ -95,7 +95,7 @@ export const galleryWrapperStyles = css`
 		border-right: 1px solid ${neutral[100]};
 	}
 
-	${from.wide} {
+	${from.desktop} {
 		width: ${wide}px;
 	}
 `;
--- a/src/components/editions/headerMedia/index.tsx
+++ b/src/components/editions/headerMedia/index.tsx
@@ -4,7 +4,6 @@ import type { SerializedStyles } from '@emotion/core';
 import { css } from '@emotion/core';
 import type { Sizes } from '@guardian/image-rendering';
 import { Img } from '@guardian/image-rendering';
-import { remSpace } from '@guardian/src-foundations';
 import { from } from '@guardian/src-foundations/mq';
 import type { Format } from '@guardian/types';
 import { Design, Display, none, some } from '@guardian/types';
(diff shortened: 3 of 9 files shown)


.storybook/preview.js
```diff
@@ -56,7 +56,7 @@ const customViewports = {
 
 export const parameters = {
 	viewport: {
-		defaultViewport: 'mobile',
+		defaultViewport: 'tablet',
```
faresite, line 59: choices

src/components/editions/article/index.tsx
```diff
@@ -95,7 +95,7 @@ export const galleryWrapperStyles = css`
 		border-right: 1px solid ${neutral[100]};
 	}
 
-	${from.wide} {
+	${from.desktop} {
```
faresite, line 98: Not too sure we should rename the code to match the storybook viewports. We generally use `from.wide` much more in the codebase, and maybe we should get rid of `from.desktop` altogether to be more semantically correct as both apps-rendering and editions-rendering are not meant for desktop anyway? What do you think?

src/components/editions/headerMedia/index.tsx
```diff
@@ -115,14 +80,13 @@ const getImageStyle = (
 	return css`
 		display: block;
 		width: 100%;
-		height: calc(100vw * ${height / width});
 
 		${from.tablet} {
-			width: ${tabletImageWidth}px;
-			height: ${(tabletImageWidth * height) / width}px;
+			width: calc(100vw - 3.75rem);
```
faresite, line 85: Nice one! 👍
Don't know how that happened again since we fixed it once already [here](https://github.com/guardian/apps-rendering/pull/1133/files#r567728860) .. we didn't have storyshots back then so hopefully next time we'll catch it in time 🤞

src/components/editions/article/index.tsx
joecowton1: @faresite  sorry could have explained my thinking here a bit better!

Basically when we use `${from.wide}` it's _so_ wide that we don't see the changes from it within any of our storybook views (obviously).

Because the Figma designs are split along the lines `tablet-potrait` and `tablet-landscape` with different layouts, if we don't use the `${from.desktop}` breakpoint we have no way of distinguishing the layouts in the code in such a way that reflects this.  Or to put it another way, `${from.desktop}` === `tablet-landscape`. 

Also, I think for consistency across the Guardian, its good to display all the Gu breakpoints in Storybook and in order for that to work we need to swap `${from.wide}` for `${from.desktop}`. 

Does that make sense to you?

.storybook/preview.js
joecowton1: Not much in it but I kept having to change it - happy to revert if mobile more useful but don't think there's much in it!

src/components/editions/article/index.tsx
faresite: Thanks @joecowton1 , that does make sense 👍

faresite (approved):
👍

PR #1180, opened 2021-02-22
updated thrift to include labs for related item type

marjisound:
## Why are you doing this?
This PR is to add the LABS type to related item type. 

## TODO
We need to do an npm release to get the new version in npm

## Changes

- Change 1
- Change 2

## Screenshots

| Before | After |
| --- | --- |
|  |  |


joecowton1:
Should we close this one @marjisound ?

JamieB-gu:
We decided this wasn't needed because this type is standalone - it doesn't correspond to the `Format` type.

PR #1183, opened 2021-02-24
Wrapping third party embedded content in a ClickToView overlay

frj:
This change adds an overlay for third party embedded content that forces the user to give consent to being tracked by the third party before being able to see the third party content.

Not all third party embedded content tracks. Capi exposes a 'tracking' field on each element indicating whether the embedded content is considered to track. 

This is currently worked out by checking the html stored in the element for iframe or script tags and extracting a domain from the src attribute. This domain is compared to a white list of 'non-tracking' embed provider domain. 

## Why are you doing this?

This is being done as part of the Embeds Consent project which is addressing concerns about the privacy of users of our products posed by third party embedded content.

## Changes

- Adding ClickToView component. This component renders the consent overlay and handles the displaying of its child elements (the third party embedded content) in response to the 'Accept' button. This was copied over from DCR and it is the intention to move this to a shared library
-  Adding the EmbedComponentWrapper, this handles the logic around whether to display the overlay based on the tracking fields in CAPI. It also handles the storing of fields required to re-create the overlay and the embedded content client side during hydration/rendering. 
- Modifying rendering to output the EmbedComponentWrapper rather than the EmbedComponent its self.
- Modifying article.ts to enable the hydration of the overlay components so that it behaves correctly when clicked.

## Todos/Questions
- Ideally this feature would be behind a feature switch so this PR can be merged but the feature would remain 'invisible' until the decision is made to release it.

## Screenshots

| Before | After |
| --- | --- |
| Screenshot 2021-02-08 at 16 27 48 |  Screenshot 2021-02-10 at 14 26 48|

## Notes
- If you are in the process of creating a new Embed type you will need to do some work to get the ClickToView working, ie to hydrate the component on the client. A good example of how to do this is [here](https://github.com/guardian/apps-rendering/pull/1183/commits/b2ab7569edb42bb271e2f2d05ec74a178e3eb753), it shows the work done for the Instagram component.


src/components/ClickToView.tsx
```diff
@@ -0,0 +1,181 @@
...
+					`}
+				>
+					{source
+						? `Allow ${source} content?`
+						: 'Allow content provided by a third party?'}
+				</div>
+				<div
+					className={css`
+						${textSize}
+						a {
```
JamieB-gu, line 130: Are there any `a` tags in this `div`?

package.json
```diff
@@ -39,7 +39,7 @@
     "@guardian/apps-rendering-api-models": "^0.11.0",
     "@guardian/atoms-rendering": "2.0.7",
     "@guardian/bridget": "^1.7.0",
-    "@guardian/content-api-models": "^15.9.6",
+    "@guardian/content-api-models": "^15.9.16",
```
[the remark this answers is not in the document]
frj: Ill investigate this there is definitely something weird going on with my local npm setup

src/components/ClickToView.tsx
```diff
@@ -0,0 +1,181 @@
+import { Button } from '@guardian/src-button';
+import { space } from '@guardian/src-foundations';
+import { background, border } from '@guardian/src-foundations/palette';
+import { headline, textSans } from '@guardian/src-foundations/typography';
+import { SvgCheckmark } from '@guardian/src-icons';
+import { css } from 'emotion';
+import React, { useState } from 'react';
+
+type RoleType =
```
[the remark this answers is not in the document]
frj: The Role in guardian types is a numeric enum rather than strings, Ill do some investigation as to why this is necessary.

src/components/ClickToView.tsx
```diff
@@ -0,0 +1,181 @@
...
+	| 'inline'
+	| 'thumbnail'
+	| 'halfWidth';
+
+export type ClickToViewProps = {
+	children: React.ReactNode;
+	role?: RoleType;
+	onAccept?: () => void;
+	source?: string;
+	sourceDomain?: string;
```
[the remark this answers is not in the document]
frj: Currently the only use of this callback in DCR is for the generic 'unsafe' embeds, the callback is used to kick off the iframe resizing logic once the click to view overlay has been clicked and the iframe has been actually rendered.

So I think this will probably be optional but I think that will fall out of the work that is going on currently.

Ill make the changes to use the Option convention.

src/components/ClickToView.tsx
frj: Good point!

src/components/ClickToView.tsx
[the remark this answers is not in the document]
frj: Youre right there is no reason not to use Role from guardian/types Ive made that change.

src/components/embedWrapper.tsx
```diff
@@ -0,0 +1,243 @@
...
+			return {
+				kind: EmbedKind.Spotify,
+				src: embed.src,
+				width: embed.width.toString(),
+				height: embed.height.toString(),
+				...withDefault<Record<string, string>>({})(
+					map<string, Record<string, string>>((source) => {
+						return { source };
+					})(embed.source),
+				),
```
JamieB-gu, line 33: An alternative way to write this. If you find it less clear feel free to ignore 🙂:

```suggestion
				...pipe2(
					embed.source,
					map((source) => ({ source })),
					withDefault<Record<string, string>>({}),
				),
```

Since you're using this pattern a few times here it may be worth pulling out into a function?

src/components/embedWrapper.tsx
```diff
@@ -0,0 +1,243 @@
...
+		return andThen((value: string) => {
+			const parsedValue = Number.parseInt(value);
+
+			if (Number.isNaN(parsedValue)) {
+				return err(`${value} is not a integer`);
+			}
+
+			return ok(parsedValue);
+		})(requiredStringParam(container, parameterName));
+	};
```
JamieB-gu, line 122: Again, alternative way to write this, feel free to ignore if it doesn't look clearer to you:

```suggestion
	const requiredNumberParam = (
		container: Record<string, string>,
		parameterName: string,
	): Result<string, number> =>
		pipe(
			requiredStringParam(container, parameterName),
			andThen((value: string) => {
				const parsedValue = Number.parseInt(value);

				if (Number.isNaN(parsedValue)) {
					return err(`${value} is not a integer`);
				}

				return ok(parsedValue);
			}),
		);
```

src/components/embedWrapper.tsx
```diff
@@ -0,0 +1,243 @@
...
+			.getAttributeNames()
+			.filter((currentKey) => currentKey.startsWith('data-'))
+			.reduce((acc, currentKey) => {
+				const newKey = currentKey
+					.replace(/^data-/g, '')
+					.replace(/-./g, (x) => x.toUpperCase()[1]);
+				const currentValue = container.getAttribute(currentKey);
+				return { ...acc, [newKey]: currentValue };
+			}, {});
+	};
```
JamieB-gu, line 137: Does this provide a similar output?

```tsx
Object.assign({}, container.dataset);
```

src/components/embedWrapper.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+	const elementProps = parseDataAttributesFromElement(container);
+
+	switch (elementProps['kind'] as EmbedKind) {
```
JamieB-gu, line 141: Can we parse the `kind` field so we can guarantee its type and/or add a `default` to this `switch`? Just want to make sure we avoid an invalid value sneaking through and accidentally returning `undefined`.

src/components/embedWrapper.tsx
```diff
@@ -0,0 +1,243 @@
...
+	return h(ClickToView, {
+		source: embed.source,
+		sourceDomain: embed.sourceDomain,
+		children: h(EmbedComponent, { embed }),
+		role: none,
+		onAccept: none,
+	});
+};
+
+const withDatasetKeyFormat = (
```
JamieB-gu, line 209: Could we add a comment to this function just to explain what it's doing?

src/client/article.ts
```diff
@@ -436,6 +438,19 @@ function richLinks(): void {
 		});
 }
 
+function hydrateClickToView(): void {
+	document.querySelectorAll('.click-to-view-container').forEach((container) =>
```
JamieB-gu, line 442: Could we prefix this class with `js-` just to make clear its purpose is to be used by JavaScript?

```suggestion
	document.querySelectorAll('.js-click-to-view-container').forEach((container) =>
```

src/components/embedWrapper.tsx
```diff
@@ -0,0 +1,243 @@
...
+const EmbedComponentWrapper: FC<Props> = ({ embed }: Props) => {
+	if (
+		embed.tracking === EmbedTracksType.TRACKS ||
+		embed.tracking === EmbedTracksType.UNKNOWN
+	) {
+		return h(
+			'div',
+			{
+				...withDatasetKeyFormat(embedToDivProps(embed)),
+				className: 'click-to-view-container',
```
JamieB-gu, line 229: Companion suggestion to a comment I made above.

```suggestion
				className: 'js-click-to-view-container',
```

src/embed.ts
```diff
@@ -19,23 +19,29 @@ import type { DocParser } from 'types/parserContext';
 // ----- Types ----- //
 
 const enum EmbedKind {
-	Generic,
-	Spotify,
-	YouTube,
+	Generic = 'Generic',
+	Spotify = 'Spotify',
+	YouTube = 'YouTube',
```
JamieB-gu, line 24: Just curious why you decided to migrate these to `string`s, are you using the `string` values somewhere?

src/lib.ts
```diff
@@ -83,6 +84,41 @@ const parseIntOpt = (int: string): Option<number> => {
...
+	if (resultB.kind === ResultKind.Err) {
+		return resultB;
+	}
+
+	return ok(f(resultA.value, resultB.value));
+};
+
+const fold = <A, B>(f: (value: A) => B, ifNone: B) => (opt: Option<A>): B => {
+	return withDefault(ifNone)(map(f)(opt));
+};
```
JamieB-gu, line 121: This is nice 🚀 - similar signature to the `maybeRender` function above. Perhaps we should re-write that function to use this under the hood 🤔.

JamieB-gu (commented):
Some really great stuff here! Particularly like the use of `Result`/`Option` and the introduction of `fold` 🚀

Mostly small comments and nit-picks, very little in the way of major or blocking comments 👍.

src/components/ClickToView.tsx
```diff
@@ -0,0 +1,183 @@
...
+	onAccept,
+	source,
+	sourceDomain,
+}: ClickToViewProps): JSX.Element => {
+	const [isOverlayClicked, setIsOverlayClicked] = useState<boolean>(false);
+
+	const handleClick = (): void => {
+		setIsOverlayClicked(true);
+		if (onAccept.kind === OptionKind.Some) {
+			setTimeout(() => onAccept.value());
```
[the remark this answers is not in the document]
frj: This was done this way based on its usage in DCR, Im assuming the same use case is going to come up in AR.

This callback is used to kick off the code that resizes the iframes we wrap the generic embeds once they have been 'uncovered' by clicking on the accept button. 

The resize code can only be kicked off once the iframe has actually made it into the dom, hence running it in this timeout.

Ill leave it in for the moment, but will definitely revisit it before merging.

src/components/embedWrapper.tsx
frj: Nice :)

src/embed.ts
frj: So the string value appears as one of the 'data' attributes on the div generated by embedWrapper, the numeric version would work I just thought this might make things clearer if you were ever looking at the dom. 

Im happy either way?

src/components/embedWrapper.tsx
frj: Good point! there are some fiddly things around the types of the dom elements that sent me down this route.

I had another go and I think it does make sense to use the dataset attribute.

frj:
[Embed Overlay in Apps Rendering](https://trello.com/c/8xltWb9Q/1395-embed-overlay-in-apps-rendering)

package.json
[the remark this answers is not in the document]
frj: I was using the wrong version of npm which was messing up the package-lock.json, Ive reverted these changes.

src/embed.ts
```diff
@@ -18,31 +18,40 @@ import type { DocParser } from 'types/parserContext';
...
+	tracking: EmbedTracksType;
 }
 
 interface Instagram {
 	kind: EmbedKind.Instagram;
 	id: string;
 	caption: Option<string>;
+	source: Option<string>;
+	sourceDomain: Option<string>;
+	tracking: EmbedTracksType;
```
JamieB-gu, line 54: If we _know_ these are from YouTube, Spotify and Instagram don't we already know the `source` and `sourceDomain`?

JamieB-gu (commented):
Looks great! Couple of final comments/questions.

src/components/ClickToView.tsx
```diff
@@ -0,0 +1,189 @@
...
+				`}
+			>
+				<div
+					css={css`
+						${roleHeadlineSize(roleWithDefault)}
+						margin-bottom: ${remSpace[1]};
+					`}
+				>
+					{fold(
+						(source: string) => `Allow ${source} content?`,
```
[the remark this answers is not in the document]
frj: Thanks good spot.

src/embed.ts
frj: Yes its a good point, I guess there is a certain appeal to all of the embeds working in the same way.

Let me take a look and make sure this wouldnt complicate the code.

frj: Ive made this change, it adds a bit of complication to the embedWrapper component but I think it an improvement overall.

JamieB-gu (approved):

PR #1186, opened 2021-02-25
LIVE-1719: Gallery template story

faresite:
# Gallery template story

The Media article story was using gallery template but without the body, this PR adds the gallery body.

NB. Looks like there might be bug in the gallery template now where the text in the tablet storybook viewport is really at the edge, possibly a regression. I'll create a separate ticket to fix it.


joecowton1 (approved):
Nicely done 👍 

Just a minor thing but not sure if you're deliberately not assigning anyone to review these?

PR #1190, opened 2021-02-26
LIVE-1785: use enum lookups in mock data

joecowton1:
## Why are you doing this?

Adds enum lookups to mock data rather than using raw numbers.

### DIFF
--- a/src/fixtures/item.ts
+++ b/src/fixtures/item.ts
@@ -1,10 +1,23 @@
 // ----- Imports ----- //
 
-import { Design, Display, none, Pillar, some, toOption } from '@guardian/types';
+import {
+	Design,
+	Display,
+	none,
+	OptionKind,
+	Pillar,
+	ResultKind,
+	Role,
+	some,
+	toOption,
+} from '@guardian/types';
 import type { Option } from '@guardian/types';
 import type { Body } from 'bodyElement';
+import { ElementKind } from 'bodyElement';
 import { parse } from 'client/parser';
 import type { Contributor } from 'contributor';
+import type { MainMedia } from 'headerMedia';
+import { MainMediaKind } from 'headerMedia';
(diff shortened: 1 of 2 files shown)


src/fixtures/item.ts
```diff
@@ -98,56 +114,59 @@ const doc = docFixture();
...
 const body: Body = [
 	{
-		kind: 0,
+		kind: ResultKind.Ok,
 		value: {
-			kind: 0,
+			kind: ElementKind.Text,
 			doc,
 		},
 	},
```
JamieB-gu, line 122:
```ts
ok({ kind: ElementKind.Text, doc }),
```

src/fixtures/item.ts
```diff
@@ -98,56 +114,59 @@ const doc = docFixture();
...
+			credit: {
+				kind: OptionKind.Some,
+				value: 'Photograph: Sam Frost/The Guardian',
+			},
 			nativeCaption: {
-				kind: 0,
+				kind: OptionKind.Some,
 				value:
 					'Jane Giddins outside her home in Newton St Loe, Somerset. She is denied the legal right to buy the freehold because of an exemption granted to Prince Charles.',
 			},
```
JamieB-gu, line 148:
```ts
nativeCaption: some('Jane Giddins outside her home in Newton St Loe, Somerset. She is denied the legal right to buy the freehold because of an exemption granted to Prince Charles.'),
```

JamieB-gu (approved):
Nice!

One thing you _could_ do to simplify this is to use the `ok` and `some` helpers to create instances of `Result` and `Option`, which saves you writing out the full `OptionKind`/`ResultKind`/`kind`/`value` object. I've dropped in a couple of examples in comments, although I don't think it's a blocker.

PR #1199, opened 2021-03-02
LIVE-1834: move design.comment above display.showcase

joecowton1:
## Why are you doing this?

In Dotcom, users are able to add `display.showcase` on top of `design.comment` to trigger a different article header layout. However, in Editions we only have a single `comment` design so we need to ensure `design.comment` comes above `display.showcase` anywhere that's effected.

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/editions/headline/index.tsx
+++ b/src/components/editions/headline/index.tsx
@@ -171,16 +171,24 @@ const getHeadlineStyles = (
 ): SerializedStyles => {
 	const sharedStyles = getSharedStyles(format);
 
-	// Display.Immersive needs to come before Design.Interview
-	switch (format.display) {
-		case Display.Immersive:
-			return css(
-				sharedStyles,
-				getFontStyles('tight', 'bold'),
-				immersiveStyles,
-			);
-		case Display.Showcase:
-			return css(sharedStyles, getFontStyles('tight', 'bold'));
+	if (format.display === Display.Immersive) {
+		return css(
+			sharedStyles,
+			getFontStyles('tight', 'bold'),
+			immersiveStyles,
+		);
+	}
+
+	if (format.design === Design.Comment) {
+		return css(
+			sharedStyles,
+			getFontStyles('regular', 'light'),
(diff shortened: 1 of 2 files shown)


JamieB-gu (approved):

src/components/editions/headline/index.tsx
```diff
@@ -171,16 +171,24 @@ const getHeadlineStyles = (
...
-	switch (format.display) {
-		case Display.Immersive:
-			return css(
-				sharedStyles,
-				getFontStyles('tight', 'bold'),
-				immersiveStyles,
-			);
-		case Display.Showcase:
-			return css(sharedStyles, getFontStyles('tight', 'bold'));
+	if (format.display === Display.Immersive) {
```
mohammad-haque, line 174: any reason for not keeping the `switch` statements?

mohammad-haque (approved):
Looks good to me with a minor comment 👍

src/components/editions/headline/index.tsx
joecowton1: We need to sit the `format.design === Design.Comment` between the two `format.display` cases

mohammad-haque: Thanks, I thought you could do that ordering within the `switch cases`.

Also since the ordering is important it worth adding a comment so that future refactoring doesn't break this order

joecowton1: yeah we can't use both `design` and `display` in the same switch statement.

added a comment 👍

JamieB-gu: @mohammad-haque I think what Joe's saying is that we're no longer checking a single value (`display`), we're now checking both `display` _and_ `design`, which isn't possible with a `switch` (unless you use a tuple, but that's a bit awkward in JavaScript).

mohammad-haque: Cool, thank you 👍

PR #1201, opened 2021-03-02
Re-work Instagram Embeds

JamieB-gu:
## Why are you doing this?

Following on from #1169, this adds Instagram embeds to the new `Embed` model.

### iFrames And CSPs

Instead of using `iframe[srcDoc]`, as is the [case on DCR](https://github.com/guardian/dotcom-rendering/blob/08d6d52d6dd26f600c0794784f7268477c6be67a/src/web/components/elements/InstagramBlockComponent.tsx), it instead uses `iframe[src]`. This means that we no longer have to allow whatever scripts and styles Instagram uses (including inline ones) in our CSP, as `iframe[src]`s use a separate CSP to the main page. I'd appreciate any feedback on this approach, particularly any concerns @oliverlloyd @gtrufitt @liywjl @sndrs @adamnfish.

**Note:** I'm also going to start wrapping these embeds in `<figure>` tags in a follow-up PR.

## Changes

- Parsed Instagram elements to `Embed`s
- Added `Instagram` to `Embed` and included parser
- Removed specific rendering for Instagram elements
- Added Instagram to embed component
- Added `Instagram` component
- Removed Instagram script from CSP

### Auto-resizing To Fit

**Piece:** https://www.theguardian.com/travel/2020/apr/26/how-to-visit-amsterdam-while-stuck-at-home-in-lockdown-rijksmuseum

This embed auto-resizes to fit the height of the `iframe` (a feature of the Instagram embed).

| Before | After |
| - | - |
|  before | after |

### The Whitespace Issue

**Piece:** https://www.theguardian.com/world/2016/dec/16/australian-thieves-paraded-with-signs-on-indonesian-streets

This one doesn't resize itself quite as well, and leaves some whitespace. On the plus side, before this change it didn't render properly at all.

| Before | After |
| - | - |
| before-2 | after-2 |


joecowton1 (approved):
LGTM 👍

marjisound (approved):

PR #1176, opened 2021-02-19
LIVE-1740: Update types to 4.0.0

dskamiotis:
## Why are you doing this?

Adds `Letter` and `Labs` types. 

Currently parked waiting on the following:
- Image-Rendering types bump to 4.0.0
- changes to `relatedItemTypes` thrift https://github.com/guardian/apps-rendering/pull/1180

## Changes

- Upgraded to Types 4.0.0
- Changed use of `Design.AdvertisementFeature` to `Special.Labs`
- Used `LABS` in `relatedItemType`

## Screenshots

| Before | After |
| --- | --- |
|  |  |

PR #1203, opened 2021-03-03
LIVE-1834: add GuardianView / Editorial article type

joecowton1:
## Why are you doing this?

Production feedback has flagged that the `Editorial` designs are not being reflected.

## Changes

- Add Editorial story
- add `if` case for `Design.GuardianView`
- add `Design.GuardianView` to article rendering list

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/__snapshots__/storyshots.test.ts.snap
+++ b/src/__snapshots__/storyshots.test.ts.snap
@@ -3882,6 +3882,1155 @@ exports[`Storyshots Editions/Article Default 1`] = `
 </main>
 `;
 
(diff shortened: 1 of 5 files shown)


JamieB-gu:
FYI this was renamed to `Editorial` in https://github.com/guardian/types/pull/68. If this PR is merged before #1176 (which is likely) @dskamiotis will need to tweak this check.

abeddow91 (approved):
👍 LGTM

PR #1205, opened 2021-03-03
LIVE-1692: video.ts test

marjisound:
## Why are you doing this?
subtask of [increase coverage ticket](https://theguardian.atlassian.net/browse/LIVE-1684)

### Test coverage
| | Before | After |
| - | ------ | ----- |
| Statements | 60.54%  | 61.67% (+0.24%)|
| Branches | 55.09% | 57.6% (0%)| 
| Functions | 56.74% | 57.8% (+0.87%) |
| Lines | 59.73% | 60.82% (+0.2%) |

### DIFF
--- a/src/video.test.ts
+++ b/src/video.test.ts
@@ -1,0 +1,44 @@
+import { AssetType } from '@guardian/content-api-models/v1/assetType';
+import { ElementType } from '@guardian/content-api-models/v1/elementType';
+import { AtomType } from '@guardian/content-atom-model/atomType';
+import { DisplayType } from '@guardian/content-atom-model/explainer/displayType';
+import { none } from '@guardian/types';
+import { parseVideo } from 'video';
+
+describe('parseVideo', () => {
+	const blockElement = {
+		type: ElementType.VIDEO,
+		assets: [
+			{
+				type: AssetType.VIDEO,
+			},
+		],
+	};
+	test('returns none given no atoms', () => {
+		expect(parseVideo(blockElement)).toEqual(none);
+	});
+
+	const atoms = {
+		media: [
+			{
+				id: 'atomId',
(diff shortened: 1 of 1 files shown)


joecowton1 (approved):

src/video.test.ts
```diff
@@ -0,0 +1,128 @@
...
+		};
+	});
+
+	test('returns video', () => {
+		const expected = some({
+			posterUrl: 'poster-url',
+			videoId: 'asset-id',
+			duration: undefined,
+			atomId: 'atomId2',
+			title: 'mediaTitle',
```
faresite, line 79: It's a nit but I'd reference these values as common variables so they live in one place?
Something like top level:
```
// in the same scope as `let blockElement: BlockElement;`
const atomId = 'atomId2'

const expected = some({
	...
	atomId,
}
```
If that makes sense?

faresite:
Nice one @marjisound 👍  just a small nit I've added.

src/video.test.ts
marjisound: Yes that makes sense. Will do the change and let you know :)

src/video.test.ts
```diff
@@ -30,7 +35,7 @@ describe('parseVideo', () => {
 				},
 			],
 			contentAtomTypeData: {
-				atomId: 'atomId2',
+				atomId: atomId,
```
faresite, line 38: If the object key has the same name as the value variable you can shorthand it like so:
```
contentAtomTypeData: {
	atomId,
	atomType: 'someAtomType',
}
```

faresite (approved):
Nice one @marjisound 👍

src/video.test.ts
marjisound: Thanks Fares for pointing it out

PR #1206, opened 2021-03-03
LIVE-1848: ER headline colour

faresite:
The Editions headline was using dark mode text colour even though Editions doesn't support it. 

| Before (in dark mode) | After (in dark mode) |
| --- | --- |
| Screenshot 2021-03-03 at 15 32 28 | Screenshot 2021-03-03 at 15 32 35 |


joecowton1 (approved):
Looks good once linter error is fixed 👍

PR #1207, opened 2021-03-03
Fix main's lint error and snapshots

faresite:
A lint error has sneaked into main. And snapshots needed updating.


joecowton1 (approved):
Sneaky.

PR #1197, opened 2021-03-01
LIVE-1718: Emotion 11

faresite:
Draft: Emotion 10 to 11 update. Opening draft for visibility

PR #1210, opened 2021-03-04
Bump atoms-rendering To v9.0.0

JamieB-gu:
## Why are you doing this?

We'd like to be on a more recent version of atoms-rendering. This also helps us incrementally work our way towards the Emotion 11 upgrade in #1197.

A few things to note:

- I _think_ I've configured the answer buckets changes correctly but would appreciate a double-check @liywjl 🙏
- `atoms-rendering` no longer exports the types for some of the quiz atom fields, so I've had to duplicate them here - I'll look at making a change there to start exporting them again
- The typing in the client-side code is a bit loose, there's an `any` type sneaking in - I'd like to tighten this up if possible
- The alias for `preact-render-to-string` allows it to compile but _will not work_ at runtime because the API is different to `react-dom/server`. This isn't a problem for now because it only affects Interactive Atoms, and we don't use atoms-rendering for those yet. However, I'd like to move `atoms-rendering` over to importing `react-dom/server`, and have consumers that are using Preact alias that instead (discussed this offline with @gtrufitt who mentioned DCR ran into problems trying this?).

## Changes

- Temporary alias for `preact-render-to-string`
- Split parsing of quiz atoms in Knowledge and Personality atoms
- Modelled Personality and Knowledge atoms
- Split rendering of quiz atoms into Personality and Knowledge atoms
- Added temporary `themeToPillar` function


liywjl (approved):
👍  this update seems to have covered most bases. Looks good to me

It has pointed out some isssues on our end (using `preact-render-to-string` instead of `react-dom/server` and not adding `-Type` suffix to quiz types)

faresite:
Looks good 👍  a few failing tests now, would be nice to add a couple of test cases for knowledge and personality quizzes? (there's also some lint errors btw)

src/themeStyles.ts
```diff
@@ -111,6 +111,21 @@ function themeToPillar(theme: Theme): string {
 	}
 }
 
+function themeToPillar(theme: Theme): Pillar {
+	switch (theme) {
```
faresite, line 115: I think this is why you have the `Operands of '+' operation must either be both strings or both numbers.` error. This function used to return a string and now it returns an enum number.

src/client/article.ts
```diff
@@ -370,10 +373,30 @@ function hydrateQuizAtoms(): void {
 		const props = atom.querySelector('.js-quiz-params')?.innerHTML;
 		try {
 			if (props) {
-				const quizProps: unknown = JSON.parse(
+				const quizProps = (JSON.parse(
 					props.replace(/&quot;/g, '"'),
-				);
-				ReactDOM.hydrate(h(QuizAtom, quizProps as QuizAtomType), atom);
+				) as unknown) as { quizType: string };
```
JamieB-gu, line 378: Longer term we should probably parse this, rather than casting it, to avoid runtime errors. For now atoms aren't available to users, and this gets us closer to Emotion 11, so I think it's ok.

faresite: Agreed 👍

PR #1212, opened 2021-03-04
LIVE-1889: discussion-rendering v4.6.1

faresite:
[Jira ticket](https://theguardian.atlassian.net/browse/LIVE-1889)
Precursor to the emotion 11 updates. 

Updating to discussion-rendering 4.6.1 just before a major types update. 
Once discussion-rendering was updated, the `@guardian/types` were clashing with the main dependency and  image-rendering's dependency.
`image-rendering` is declaring `@guardian/types` as both dependency and as a peer dependency, which was causing a ts compiler conflict (even though the types were identical 🤔). 
This PR uses this commit https://github.com/guardian/image-rendering/commit/6ad32f87e66a794c2d385cf9e82d605fb9e8b35c which is based off image-rendering 5.0.2 but removing `types` as a peer dependency. 
This update is intermediary so both discussion and image rendering should be updated further up with the emotion 11 update work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2221,9 +2221,9 @@
       }
     },
     "@guardian/discussion-rendering": {
-      "version": "3.2.1",
-      "resolved": "https://registry.npmjs.org/@guardian/discussion-rendering/-/discussion-rendering-3.2.1.tgz",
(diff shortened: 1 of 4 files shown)


joecowton1 (approved):

JamieB-gu (approved):
Reasonable step to get us to Emotion 11 - still a bit puzzled though 🤔

PR #1215, opened 2021-03-04
Webpack 5 And Working Storybook

JamieB-gu:
## Why are you doing this?

Most of this is @marjisound's work from #1211 to upgrade to Webpack 5. I've just tweaked the way `terser-webpack-plugin` is installed based on [a suggestion](https://github.com/storybookjs/storybook/issues/13893#issuecomment-787841568) from a storybook issue. I've also jumped to the `6.2.0` beta build of Storybook and used the Webpack 5 builder as recommended in this issue: https://github.com/storybookjs/storybook/issues/14170

**Note:** We could think about removing `@types/webpack` as a direct dependency because a) Webpack 5 carries its own type definitions, and b) anything else that depends on it can pull it in as a transitive dependency.

## Changes

Everything from #1211, plus:

- Tweaked the dependency tree for `terser-webpack-plugin` (which I did naively by running `npm i terser-webpack-plugin && npm remove terser-webpack-plugin`)
- Updated storybook libraries to 6.2.0 beta
- Added `builder-webpack5` for Webpack 5 support
- Configured Storybook to use Webpack 5 builder
- Handled lack of `crypto` in browser


joecowton1 (approved):
LGTM

marjisound (approved):

PR #1218, opened 2021-03-08
LIVE-1880 : Header Media Lightbox

abeddow91:
## Why are you doing this?
~This PR lifts the class name top the parent component of header media so that we can trigger a lightbox on click~

This PR removes the pointer event from the image caption and adds it to summary to unblock the image component.

## Changes

- Disable caption pointer event to unblock image
- Disable dark mode as editions does not currently support dark mode

### DIFF
--- a/src/__snapshots__/storyshots.test.ts.snap
+++ b/src/__snapshots__/storyshots.test.ts.snap
@@ -379,6 +379,7 @@ exports[`Storyshots Editions/Article Analysis 1`] = `
 }
 
 .emotion-7 {
+  pointer-events: none;
   position: absolute;
   left: 0;
   right: 0;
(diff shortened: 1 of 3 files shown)


joecowton1 (approved):

faresite:
Hey @abeddow91 just for some context, the image should trigger the image when clicked without moving the classname up? It does work on a regular browser, is it not triggering inside RN's webview?

abeddow91:
@faresite The image classname works fine for images in the body but not for the header image as there is a caption component stacked on top. Testing on my side it doesn't work on browser on in RN without moving the class name up. Perhaps I'm missing something though?

faresite:
@abeddow91 you're right, forgot the caption now takes full width and height. 
May I suggest in this case to use `pointer-events: none;` on the `<figcaption />` and add `pointer-events: auto;` to the button inside (I think it's the `<summary />`). What do you think?
I'm just hesitant for us to break out of the image-rendering's lightbox API if we don't have to.

abeddow91:
@faresite Much more elegant 👍 - I'll update accordingly :)

joecowton1 (approved):
Nicely done 👍

faresite (approved):
💡📦

PR #1220, opened 2021-03-08
Fix: Standfirst Spacing

abeddow91:
## Why are you doing this?

This is a fix for a styling bug that positions the share button below the standfirst text.

## Changes

- Change 1
- Change 2

## Screenshots

| Before | After|
| --- | --- |
|  |  |


faresite (approved):
Thanks for fixing @abeddow91  🙌

PR #1222, opened 2021-03-09
LIVE-1895: Body Spacing

abeddow91:
## Why are you doing this?
Adds missing vertical padding to body on devices below tablet.

## Screenshots

| Before | After |
| --- | --- |
|  |  |


faresite (approved):
LGTM 👍
I'm surprised the chromatic diff doesn't show this change but instead shows some pixel changes in the header image itself which is obviously not related to this PR  🤔

PR #1224, opened 2021-03-09
LIVE-1919: fix broken caption icon

faresite:
## Why are you doing this?
In Chrome 89, it seems there must have been a change that realigns with the spec. This in turn broke the caption icon, the way it has been broken on firefox for ages.

This PR fixes it so it works in chrome (older versions and 89) and safari (thanks @mxdvl)
Also I noticed there was some strange padding in the editions caption on wider viewports which I don't understand so I removed it. If anyone knows why it was there in the first place, do give a shout.

### icon fix
| before | after |
| -- | -- |
|Screenshot 2021-03-09 at 12 38 20 | Screenshot 2021-03-09 at 12 38 30 |

### editions caption padding
| before | after |
| -- | -- |
| Screenshot 2021-03-09 at 12 39 05 | Screenshot 2021-03-09 at 12 39 30 |

### DIFF
--- a/src/components/editions/headerImageCaption.tsx
+++ b/src/components/editions/headerImageCaption.tsx
@@ -1,6 +1,7 @@
 import type { SerializedStyles } from '@emotion/core';
 import { css } from '@emotion/core';
 import { remSpace } from '@guardian/src-foundations';
+import { from } from '@guardian/src-foundations/mq';
 import { brandAlt, neutral } from '@guardian/src-foundations/palette';
 import { textSans } from '@guardian/src-foundations/typography';
 import { SvgCamera } from '@guardian/src-icons';
@@ -12,6 +13,7 @@ import type { FC, ReactElement } from 'react';
 const captionId = 'header-image-caption';
 
 const HeaderImageCaptionStyles = (
+	isFullWidthImage: boolean,
 	iconBackgroundColor?: string,
 ): SerializedStyles => css`
 	summary {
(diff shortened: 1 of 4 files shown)


src/components/editions/headerImageCaption.tsx
```diff
@@ -46,14 +46,6 @@ const HeaderImageCaptionStyles = (
...
 		${textSans.small()};
 		box-sizing: border-box;
-
-		${from.tablet} {
-			padding-left: ${remSpace[6]};
-		}
-
-		${from.desktop} {
-			padding-left: 9rem;
-		}
```
faresite, line 56: @joecowton1 do you know why this was here in the first place? maybe there's a good reason for it and I shouldn't remove it 🤔

mxdvl:
For some context, [MDN Docs on the `summary` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary):

> Per the HTML specification, the default style for <summary> elements includes display: list-item. This makes it possible to change or remove the icon displayed as the disclosure widget next to the label from the default, which is typically a triangle.
>
> **You can also change the style to `display: block` to remove the disclosure triangle.**

(emphasis mine)

abeddow91 (approved):

src/components/editions/headerImageCaption.tsx
joecowton1: @faresite it was to keep the caption inline with the left margin on full screen images, but i get the feeling it shouldn't apply to all images, just full screen ones. Does that make sense?

joecowton1 (approved):
Nicely done 👍

PR #1225, opened 2021-03-09
LIVE-1894: Add useDarkMode prop for link styling

abeddow91:
## Why are you doing this?
The live apps supports dark mode but Editions doesn't. This PR adds a useDarkMode prop to allow us to disable dark mode styles on body anchor links.

## Screenshots

| Before | After |
| --- | --- |
|  |  |

| AR | 
| --- | 
|  |

### DIFF
--- a/src/components/anchor.tsx
+++ b/src/components/anchor.tsx
@@ -17,12 +17,14 @@ interface Props {
 	children?: ReactNode;
 	format: Format;
 	className?: SerializedStyles;
+	useDarkMode?: boolean;
 }
 
-const styles = css`
+const styles = (useDarkMode: boolean) => css`
 	text-decoration: none;
 
-	${darkModeCss`
+	${useDarkMode &&
+	darkModeCss`
         color: ${neutral[86]};
         border-color: ${neutral[46]};
     `}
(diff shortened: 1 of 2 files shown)


joecowton1 (approved):
Great stuff 👍

src/components/anchor.tsx
```diff
@@ -17,12 +17,13 @@ interface Props {
 	children?: ReactNode;
 	format: Format;
 	className?: SerializedStyles;
+	useDarkMode?: boolean;
 }
 
-const styles = css`
+const styles = (useDarkMode: boolean): SerializedStyles => css`
```
faresite, line 23: Nice one @abeddow91 , can I suggest we name it something else, `useDarkMode` sounds like a React hook. Maybe `supportsDarkMode` or something along those lines?

abeddow91: Makes sense, will do :)

src/renderer.ts
```diff
@@ -757,7 +764,7 @@ const renderEditions = (format: Format, excludeStyles = false) => (
 ): ReactNode => {
 	switch (element.kind) {
 		case ElementKind.Text:
-			return textRenderer(format, excludeStyles, element);
+			return textRenderer(format, excludeStyles, element, false);
```
faresite, line 767: Just entertaining a thought here, maybe we should use an object syntax when the function arity is above 2 or 3? So it's easier to read the arguments. Not necessarily for this PR but speaking in general as a style guide. It's a shame JS (and in consequence TS) don't support named parameters like a lot of other languages.
What are your thoughts? @abeddow91 @joecowton1 @JamieB-gu

faresite: Although with TypeScript and language tools it's easy to read the signature of the function .. so maybe not worth the cost of creating an object.

faresite (approved):
👍

PR #1232, opened 2021-03-19
LIVE-1687-date.ts test

marjisound:
## Changes

- Change 1
- Change 2

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/date.test.ts
+++ b/src/date.test.ts
@@ -1,6 +1,8 @@
 // ----- Imports ----- //
 
-import { formatDate } from './date';
+//import { act } from 'react-dom/test-utils';
+import { some } from '@guardian/types';
+import { formatDate, formatSeconds, fromString, isValidDate } from './date';
 
 // ----- Tests ----- //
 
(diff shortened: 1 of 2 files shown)


config/jestglobalSetup.js
```diff
@@ -0,0 +1,3 @@
+export default () => {
+	process.env.TZ = 'Europe/London';
```
faresite, line 2: I'd add a comment to why you need that here and mention `date.test.ts`

src/date.ts
```diff
@@ -121,7 +121,11 @@ const localMonth = (date: Date): string => months[date.getMonth()];
 
 function fromString(date: string): Option<Date> {
 	try {
-		return some(new Date(date));
+		const maybeDate = new Date(date);
+		if (!isValidDate(maybeDate)) {
+			throw new Error('not a valid date');
```
faresite, line 126: I can't remember why this throw has been added .. does it matter if we're throwing anything here since we're catching it and returning a `none` anyway?

src/date.test.ts
```diff
@@ -17,3 +25,160 @@ describe('formatDate', () => {
...
+	});
+
+	test('returns none given a non valid date string', () => {
+		expect(fromString((undefined as unknown) as string)).toEqual(none);
+		expect(fromString('random')).toEqual(none);
+	});
+});
+
+describe('formatLocal', () => {
+	test('returns correct local time zone', () => {
```
faresite, line 72: Could mention the `TZ` env setup here too?

src/date.test.ts
```diff
@@ -17,3 +25,160 @@ describe('formatDate', () => {
...
+		expect(formatSeconds('3600')).toEqual(some('1:00:00'));
+		expect(formatSeconds('3620.67')).toEqual(some('1:00:20'));
+	});
+});
+
+describe('makeRelativeDate', () => {
+	beforeEach(() => {
+		const fakeCurrentTime = new Date('2019-02-16T00:00:00');
+		jest.useFakeTimers('modern');
+		jest.setSystemTime(fakeCurrentTime);
```
faresite, line 108: Very nice 👍

faresite (commented):
Really nice @marjisound 👍 just a couple of comments.

src/date.ts
marjisound: we added this because the parameter could be an invalid date string like `(undefined as unknown) as string` or `'random'` but it still doesn't throw any exception and maybeDate gets a value of `Invalid Date`. And the result of the function would become some(Invalid Date)

I actually couldn't find any scenario that would throw an exception.

faresite (approved):
That looks great @marjisound 👍 testing dates is hard and fiddly!

PR #1236, opened 2021-03-22
Render All Footer Styles On Server

JamieB-gu:
## Why are you doing this?

The merge of #1229 triggered a latent issue in the Footer, caused by its use of dynamically generated client-side styles. The fix is to make sure all styles are rendered on the server, and only generate HTML on the client. I've done this by splitting it into a wrapper component, `Footer`, that handles all the styles and only gets rendered on the server, and an inner component, `FooterContent`, that doesn't own any styles and gets rendered on both the server and the client.

**Note:** This also stops exempting Android from the client-side footer rendering code, as this is now supported: https://github.com/guardian/android-news-app/pull/6648.

## Changes

- Use `js-footer` to select footer
- Use `logger` on client
- Moved `Footer` to main `components` directory
- Moved link styles to footer wrapper
- Created inner `FooterContent` and outer `Footer`
- Updated imports
- Stopped exempting Android from footer render

## Screenshots

| Before | After (via @GiuliaAriu) |
| - | - |
| before | after |


ab-gnm (commented):
Thanks for the changes Jamie! Great that you could resolve the "no footer on Android" issue as well.

Rest of the changes all look good to me, but I'm not sure I understand the `storyshots.test.ts.snap`. I hope someone from MSS can approve.

marjisound (approved):

PR #1241, opened 2021-03-29
LIVE-2093: Fix epic styles

faresite:
## Why are you doing this?

Since merging emotion 11, client runtime styles are being blocked by CSP, this affected the footer and epic so far. The big mystery is why wasn't emotion blocking them before 🤔 Or maybe something else is happening, which we will need to investigate, but for now we need to fix the styles as they're user facing (even if just for beta users).

The fix here is not very nice and will need to think of something better .. Moving the styles up to the wrapper (like in https://github.com/guardian/apps-rendering/pull/1236). This mostly worked, except for the buttons. The `Epic` component uses `Source`'s `Button` component which is rendered client side.
I've kept the Source button for now however copied its styles up so they can be rendered server side, while we think of a better solution.

## Changes

- changed `#epic-placeholder` to `#js-epic-placeholder`
- moved `Epic` to `EpicContent`
- `Epic` is now the wrapper component.

### DIFF
--- a/src/client/article.ts
+++ b/src/client/article.ts
@@ -17,7 +17,7 @@ import {
 } from 'client/nativeCommunication';
 import setup from 'client/setup';
 import FooterContent from 'components/footerContent';
-import Epic from 'components/shared/epic';
+import EpicContent from 'components/shared/epicContent';
 import { formatDate, formatLocal, isValidDate } from 'date';
 import { handleErrors, isObject } from 'lib';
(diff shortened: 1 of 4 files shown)


aoifemcl15:
Tested on iOS when pointing to CODE and can see the epic appearing as expected on SSR articles in light and dark mode.

marjisound (approved):

PR #1244, opened 2021-04-07
LIVE-1740: Update @guardian/types to 6.0.0

faresite:
## Why are you doing this?

We need to be on `@guardian/types` v4.0.0 so we can work on the Letter templates, however types is already on v6.0.0 so updating this in one go.

Initially we thought we will have to wait until these v5.1.0 (opened before 6 was released) are merged:
- https://github.com/guardian/atoms-rendering/pull/232 (merged now)
- https://github.com/guardian/discussion-rendering/pull/479
- https://github.com/guardian/image-rendering/pull/253

I opened this PR as a draft so we're ready to go, but looks like it that the build doesn't need these PRs merged after all 🤷‍♂️ .. possibly because nothing is conflicting in our codebase between v3 and v6 🎉 Therefore I'm publishing this PR to unblock the team.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2519,9 +2519,9 @@
       }
     },
     "@guardian/types": {
-      "version": "3.0.0",
-      "resolved": "https://registry.npmjs.org/@guardian/types/-/types-3.0.0.tgz",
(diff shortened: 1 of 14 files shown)


joecowton1 (approved):
Great work Fares 🥳

PR #1248, opened 2021-04-08
Live 1182 remove editions embeds

joecowton1:
## Why are you doing this?

Removing atoms and embeds as part of Editions embed fixes.

### Changes
- remove Editions audio atom
- remove Editions callout
- remove Editions quiz atom
- remove Editions Spotify embed
- remove Editions chart atom

### DIFF
--- a/src/components/embed.tsx
+++ b/src/components/embed.tsx
@@ -13,22 +13,23 @@ import type { FC } from 'react';
 
 interface Props {
 	embed: Embed;
+	editions?: boolean;
 }
 
 /**
  * Handles rendering of all third-party embeds.
  * See the `Embed` type for more information.
  */
-const EmbedComponent: FC<Props> = ({ embed }) => {
+const EmbedComponent: FC<Props> = ({ embed, editions }) => {
 	switch (embed.kind) {
 		case EmbedKind.Spotify:
-			return (
+			return !editions ? (
 				<Audio
 					src={embed.src}
 					width={embed.width}
 					height={embed.height}
 				/>
-			);
+			) : null;
 
 		case EmbedKind.YouTube:
 			return (
--- a/src/components/embed.tsx
+++ b/src/components/embed.tsx
@@ -13,7 +13,7 @@ import type { FC } from 'react';
 
 interface Props {
 	embed: Embed;
-	editions?: boolean;
+	editions: boolean;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


abeddow91:
`http://localhost:8080/music/2021/apr/07/baker-boy-hiatus-kaiyote-montaigne-and-others-australias-best-new-music-for-april?editions` still shows a Spotify embed

joecowton1:
> `http://localhost:8080/music/2021/apr/07/baker-boy-hiatus-kaiyote-montaigne-and-others-australias-best-new-music-for-april?editions` still shows a Spotify embed (assuming this is a spotify embed)

This is because although `EmbedKind.Spotify` is no longer rendering for Editions, not all Spotify Embeds are being correctly parsed and some are instead rendered as generic embeds.

JamieB-gu:
> not all Spotify Embeds are being correctly parsed and some are instead rendered as generic embeds as in this case.

Just to clarify, there are two ways Spotify can end up in the page. One is as a formal "audio" embed, which we treat as a distinct element and which Joe has filtered out for Editions. The other is as a "generic" embed, which is a catch-all for all sorts of different embed types and is sometimes used (perhaps incorrectly) to include Spotify.

I wouldn't say these are being parsed incorrectly as such - it's a conscious decision to [only trigger the audio parsing code](https://github.com/guardian/apps-rendering/blob/f803ffa1797a0cf17d3b4b9df5f16f6a0822b639/src/bodyElement.ts#L405) when encountering an audio element. We _could_ decide to start looking for certain patterns in generic embeds and convert them to the more formal type. Thus far we haven't done this because that'll cause a difference in how they're rendered and thus might surprise the people who chose this method of embedding.

src/components/embed.tsx
```diff
@@ -13,22 +13,23 @@ import type { FC } from 'react';
 
 interface Props {
 	embed: Embed;
+	editions?: boolean;
```
JamieB-gu, line 16: Could we make this explicit?

```suggestion
	editions: boolean;
```

JamieB-gu (approved):
Nice, surprisingly few changes required!

One question.

src/components/embed.tsx
joecowton1: Good call, done.

JamieB-gu (approved):

PR #1249, opened 2021-04-08
LIVE-2208: fix bridget client / server

faresite:
## Why are you doing this?
We have introduced webpack 5 in https://github.com/guardian/apps-rendering/pull/1215 
Webpack 5 strips away any automatic nodejs polyfills, including Buffer (you can see the list of node libs for browser [here](https://github.com/webpack/node-libs-browser)). 
Bridget (and more technically Thrift) relies on Buffer, hence webpack 5 broke features like following authors or launching the lightbox. We need to manually polyfill node libs that we want now, so I'm introducing `feross/buffer` which is the same buffer polyfill webpack v4 was using.

## Changes

- add `buffer` as a dependency
- update webpack config with `resolve.alias` and the Provide plugin.

### DIFF
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -34,8 +34,11 @@ class LaunchServerPlugin {
 
 // ----- Shared Config ----- //
 
-function resolve(loggerName: string): ResolveOptions {
-	return {
+function resolve(
+	loggerName: string,
+	isClient: boolean = false,
+): ResolveOptions {
+	const resolveOptions: ResolveOptions = {
 		extensions: ['.ts', '.tsx', '.js'],
 		modules: [path.resolve(__dirname, 'src'), 'node_modules'],
 		alias: {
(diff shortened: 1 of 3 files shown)


webpack.config.ts
```diff
@@ -54,6 +54,9 @@ function resolve(loggerName: string): ResolveOptions {
 			// `react-dom/server`, and for DCR to alias that to `preact-render-to-string`.
 			// Then we can get rid of this line.
 			'preact-render-to-string': 'react-dom/server',
+			// Webpack 5 removed a lot of the nodejs polyfills including Buffer
+			// We rely on Buffer for our thrift client/server
+			Buffer: 'buffer',
```
JamieB-gu, line 59: Will this also resolve `Buffer` to `buffer` in our server-side bundle? Should we maybe make this resolution apply for the client bundle only, since we probably want to use Node's `Buffer` server-side?

faresite: Good point, I couldn't find places where we're using Buffer on the server but good to protect ourselves either way 👍

JamieB-gu (approved):

PR #1250, opened 2021-04-08
LIVE-2132: fix editions instagram

joecowton1:
## Why are you doing this?

Instagram embeds were previously broken in Editions. 

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/server/csp.ts
+++ b/src/server/csp.ts
@@ -137,6 +137,9 @@ function cspEditions(
 			? 'https://platform.twitter.com https://cdn.syndication.twimg.com'
 			: ''
 	};
+	frame-src https://www.theguardian.com ${
+		thirdPartyEmbed.instagram ? 'https://www.instagram.com' : ''
+	};
     frame-src https://www.theguardian.com https://embed.theguardian.com https://interactive.guim.co.uk ${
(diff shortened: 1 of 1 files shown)


faresite (approved):
👍

src/server/csp.ts
```diff
@@ -137,6 +137,9 @@ function cspEditions(
 			? 'https://platform.twitter.com https://cdn.syndication.twimg.com'
 			: ''
 	};
+	frame-src https://www.theguardian.com https://www.scribd.com ${
```
faresite, line 140: maybe remove scribd.com until we know that we need it?!

joecowton1: removed :)

src/server/csp.ts
```diff
@@ -137,6 +137,9 @@ function cspEditions(
 			? 'https://platform.twitter.com https://cdn.syndication.twimg.com'
 			: ''
 	};
+	frame-src https://www.theguardian.com ${
```
JamieB-gu, line 140: You may want to add this to the existing `frame-src` rather than creating a second one, I'm not sure what the consequences of that might be 🤔.

joecowton1: already merged but actioned in https://github.com/guardian/apps-rendering/pull/1251

PR #1251, opened 2021-04-08
LIVE-2132: use same frame-src script 

joecowton1:
## Why are you doing this?

@JamieB-gu commented on a previously merged PR (https://github.com/guardian/apps-rendering/pull/1250) with a suggestion of using the same `frame-src` script rather than a new one, this PR implements that suggestion.

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/server/csp.ts
+++ b/src/server/csp.ts
@@ -137,7 +137,6 @@ function cspEditions(
 			? 'https://platform.twitter.com https://cdn.syndication.twimg.com'
 			: ''
 	};
-	frame-src https://www.theguardian.com 
     frame-src https://www.theguardian.com https://embed.theguardian.com https://interactive.guim.co.uk ${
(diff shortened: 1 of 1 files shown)


src/server/csp.ts
```diff
@@ -137,16 +137,14 @@ function cspEditions(
 			? 'https://platform.twitter.com https://cdn.syndication.twimg.com'
 			: ''
 	};
-	frame-src https://www.theguardian.com ${
-		thirdPartyEmbed.instagram ? 'https://www.instagram.com' : ''
-	};
+	frame-src https://www.theguardian.com
```
JamieB-gu, line 140:
```suggestion

```

joecowton1:
Thanks @JamieB-gu , missed that somehow :)

JamieB-gu (approved):

PR #1254, opened 2021-04-09
LIVE-1690 - Tests for relatedContent.ts

marjisound:
## Why are you doing this?
subtask of [increase coverage ticket](https://theguardian.atlassian.net/browse/LIVE-1684)

### Test coverage
| | Before | After |
| - | ------ | ----- |
| Statements | 66.55%  | 70.35% (+3.8%)|
| Branches | 59.86% | 66.53% (+6.67%)| 
| Functions | 61.84% | 66.08% (+6.22%) |
| Lines | 65.7% | 69.5% (+3.8%) |


src/relatedContent.ts
```diff
@@ -47,7 +55,7 @@ const parseHeaderImage = (content: Content): Image | undefined => {
 					url: asset.file ?? '',
 					height: asset.typeData?.height ?? 360,
 					width: asset.typeData?.width ?? 600,
-					alt: data?.alt,
+					altText: data?.alt,
```
marjisound, line 58: I changed the property name because the type `Image` has `altText` as a property. And we were setting the `alt` instead which doesn't exist in `Image`

src/relatedContent.test.ts
```diff
@@ -0,0 +1,337 @@
...
+			lastModified: capiDateTime,
+		};
+		defaultContent.fields = contentFields;
+
+		const actual = parseRelatedContent([defaultContent]);
+
+		expect(actual.relatedItems[0].lastModified).toEqual(capiDateTime);
+	});
+
+	it('returns the first 4 related content given more than 4 content items', () => {
```
faresite, line 99: This test doesn't check if the 4 items are actually the first 4, it only checks that the it returns 4 items.

src/relatedContent.test.ts
```diff
@@ -0,0 +1,337 @@
...
+
+	it('returns Type FEATURE given a content with feature tone tag and any other tag', () => {
+		addTagToTagsList('tone/features', TagType.TONE);
+		addTagToTagsList('type/video', TagType.TYPE);
+
+		const actual = parseRelatedContent([defaultContent]);
+		expect(actual.relatedItems[0].type).toEqual(RelatedItemType.FEATURE);
+	});
+
+	const addTagToTagsList = (tagId: string, tagType: TagType) => {
```
faresite, line 217: Nit, can we move this block up before the tests that use it?

src/relatedContent.ts
faresite: Good catch .. I'm surprised typescript wasn't complaining before .. I wonder why 🤔

faresite (approved):
Just a couple of nitpicking comments. Great work 👍

PR #1256, opened 2021-04-09
LIVE-2148: interactive atoms

joecowton1:
## Why are you doing this?

We're looking to include additional atoms and embeds in Editions where possible. This PR brings together the AR and ER approaches to CSP and allows us to render `interactive` atoms.

See this article for an example:
http://localhost:8080/world/2020/jun/10/uk-coronavirus-lockdown-20000-lives-boris-johnson-neil-ferguson?editions

## Changes

- Use `meta` component for Editions header
- share `csp` component between editions and AR
- add additional fields (rules?) to csp functions 

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/server/editionsPage.tsx
+++ b/src/server/editionsPage.tsx
@@ -7,7 +7,7 @@ import type { EmotionCritical } from '@emotion/server/create-instance';
 import type { RenderingRequest } from '@guardian/apps-rendering-api-models/renderingRequest';
 import type { Option } from '@guardian/types';
 import { map, none, some } from '@guardian/types';
-import { getThirdPartyEmbeds, requiresInlineStyles } from 'capi';
+import { getThirdPartyEmbeds } from 'capi';
 import type { ThirdPartyEmbeds } from 'capi';
 import { atomCss, atomScript } from 'components/atoms/interactiveAtom';
 import Article from 'components/editions/article';
@@ -70,12 +70,8 @@ function renderHead(
 
 	return `
         ${renderToString(meta)}
-        <link rel="stylesheet" type="text/css" href="/fontSize.css">
         <style>${generalStyles}</style>
         <style data-emotion-css="${emotionIds.join(' ')}">${itemStyles}</style>
-        <script id="targeting-params" type="application/json">
-            ${JSON.stringify(request.targetingParams)}
-        </script>
     `;
 }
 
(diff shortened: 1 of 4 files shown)


src/server/editionsPage.tsx
```diff
@@ -34,37 +37,47 @@ const docParser = JSDOM.fragment.bind(null);
...
-		<meta name="viewport" content="initial-scale=1" />
-		<meta http-equiv="Content-Security-Policy" content="${cspString}" />
-        <style>${styles}</style>
+        ${renderToString(meta)}
+        <link rel="stylesheet" type="text/css" href="/fontSize.css">
+        <style>${generalStyles}</style>
         <style data-emotion-css="${emotionIds.join(' ')}">${itemStyles}</style>
+        <script id="targeting-params" type="application/json">
+            ${JSON.stringify(request.targetingParams)}
+        </script>
```
JamieB-gu, line 78: Is Editions starting to use the targeting params?

src/server/editionsPage.tsx
```diff
@@ -102,7 +115,16 @@ function render(
...
-	const head = renderHead(request, thirdPartyEmbeds, body.css, body.ids);
+	const inlineStyles = requiresInlineStyles(request.content);
+
+	const head = renderHead(
+		item,
+		request,
+		thirdPartyEmbeds,
+		body.css,
+		body.ids,
+		inlineStyles,
```
JamieB-gu, line 126: I think you can just pass this as `false`, the Editions version of the CSP ignores it anyway.

src/server/editionsPage.tsx
```diff
@@ -34,37 +37,47 @@ const docParser = JSDOM.fragment.bind(null);
...
+	const meta = h(Meta, { title: request.content.webTitle, cspString });
 
 	return `
-		<meta charSet="utf-8" />
-		<title>${request.content.webTitle}</title>
-		<meta name="viewport" content="initial-scale=1" />
-		<meta http-equiv="Content-Security-Policy" content="${cspString}" />
-        <style>${styles}</style>
+        ${renderToString(meta)}
+        <link rel="stylesheet" type="text/css" href="/fontSize.css">
```
JamieB-gu, line 73: Does Editions support the same scaling mechanism as Android? If not I think you can probably ignore this.

JamieB-gu (commented):
Nice. Couple comments/questions.

src/server/editionsPage.tsx
joecowton1: No don't think so, good call.

src/server/editionsPage.tsx
joecowton1: Cool, fixed.

src/server/editionsPage.tsx
joecowton1: Removed 👍

JamieB-gu (approved):

PR #1265, opened 2021-04-13
LIVE-2179: Send targeting params on init

faresite:
## Why are you doing this?

We need to send targeting params to the native apps on init. 
Bridget was updated to 1.9.0 to use the new service (https://github.com/guardian/bridget/pull/72)

## Changes

- Bump to Bridget 1.9.0
- Create the analytics Bridget client
- Use the client to send targeting params on init

### DIFF
--- a/src/client/nativeCommunication.ts
+++ b/src/client/nativeCommunication.ts
@@ -8,11 +8,11 @@ import { errorToString, isObject, memoise } from 'lib';
 import { logger } from 'logger';
 import {
 	acquisitionsClient,
+	analyticsClient,
 	commercialClient,
 	galleryClient,
 	userClient,
 	videoClient,
-	analyticsClient,
 } from '../native/nativeApi';
 
 type Slot = AdSlot | VideoSlot;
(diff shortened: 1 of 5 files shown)


src/client/nativeCommunication.ts
```diff
@@ -264,4 +265,15 @@ function reportNativeElementPositionChanges(): void {
 	}
 }
 
-export { ads, slideshow, videos, reportNativeElementPositionChanges };
+function sendTargetingParams(): void {
+	const targetingParams = getTargetingParams();
+	void analyticsClient.sendTargetingParams(targetingParams);
```
JamieB-gu, line 270: Will this crash on devices on an older version of Bridget?

JamieB-gu (approved):

src/client/nativeCommunication.ts
faresite: I think the thrift client/server is built to deal with this.. When I tested it in a browser without any bridget, any other calls still work. Do you think we should start "if versioning" already?
I'm gonna double check it doesn't crash in the simulator too.

faresite: Just tested 👍 All works well .. didn't crash or break any subsequent bridget calls. Good to know this for when we introduce new methods in newer bridget versions.

PR #1266, opened 2021-04-13
LIVE-2147: article video embed

joecowton1:
## Why are you doing this?

Article body video embeds weren't rendering correctly. This PR adds an iframe based `Video` component which takes an `atomId` and `title`.  This can now be used in the `headerMedia` component too. 

## Changes

- Create Editions `Video` component
- Update `MediaAtom` type to include atomId
- add embed url to csp config
- add storybook story

**Storybook doesn't seem to be working at the moment**

## Screenshots

| Before | After |
| --- | --- |
|  |  |


abeddow91 (approved):
Nice one!

src/components/editions/video/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+const Video: FC<Props> = ({ title, atomId }) => {
+	return (
+		<div css={videoWrapperStyles}>
+			<iframe
+				title={title}
+				css={videoStyles}
+				frameBorder="0"
+				scrolling="no"
+				allowFullScreen
+				src={`https://embed.theguardian.com/embed/atom/media/${atomId}#noadsaf`}
```
JamieB-gu, line 37: Question for @paperboyo and @akash1810, is this the correct URL to use for embedding media atoms?

JamieB-gu (commented):
Code all looks sensible to me. Just had a question about the correct URL to use for the `iframe` source - dotcom is using `youtube.com` for these.

> Storybook doesn't seem to be working at the moment

Uh oh...

faresite:
> Code all looks sensible to me. Just had a question about the correct URL to use for the `iframe` source - dotcom is using `youtube.com` for these.
> 
> > Storybook doesn't seem to be working at the moment
> 
> Uh oh...

If we used `youtube.com` wouldn't we be losing the benefits of having things like the poster image and the button?

JamieB-gu:
> If we used `youtube.com` wouldn't we be losing the benefits of having things like the poster image and the button?

Don't think so. Dotcom manage that [just fine](https://www.theguardian.com/world/video/2021/apr/15/the-mexican-women-who-kicked-out-the-cartels-video).

It's all [handled by](https://github.com/guardian/atoms-rendering/blob/main/src/YoutubeAtom.tsx) `atoms-rendering`. @nitro-marky may be able to elaborate.

PR #1267, opened 2021-04-15
LIVE-1379: 🗑 Delete media.js from client bundle

faresite:
## Why are you doing this?

[JIRA ticket](https://theguardian.atlassian.net/browse/LIVE-1379?atlOrigin=eyJpIjoiODZhNDczMmRiM2UwNGRkZGJkYTE1OTFkMTU2YmQwOTEiLCJwIjoiaiJ9)

## Changes

- Delete `media.js` and direct script source of non liveblog articles to the `article.js` file.
- Adds article `Format` as a json script in the html
- parses it in the client side bundle
- init client side functions based on the format.


src/client/article.ts
```diff
@@ -451,20 +454,40 @@ function richLinks(): void {
...
+		footerInit();
+		insertEpic();
+		callouts();
+		renderComments();
+		hasSeenCards();
+		initAudioAtoms();
+		hydrateQuizAtoms();
+		localDates();
+		richLinks();
+	}
```
JamieB-gu, line 490: Do we not want to do these things for media articles?

src/client/nativeCommunication.ts
```diff
@@ -49,6 +51,22 @@ function getRect(slotPosition: DOMRect): Rect {
 	});
 }
 
+const getFormat: () => Option<Format> = () => {
```
JamieB-gu, line 54: Why the `nativeCommunication` module out of curiosity? Do you anticipate that we'll start sending this to the native layers at some point?

src/client/article.ts
faresite: Maybe we do? I was just refactoring it the way it is currently. The `media.ts` currently only calls `ads` and `slideshow`.. maybe it's a legacy thing. Shall we call these for media articles too?

src/client/nativeCommunication.ts
faresite: I don't think we will need to send format at some point since we're the ones handling all rendering .. the reason I needed this is to distinguish media articles so I can call the corresponding methods. But if we're ok to call the same methods for media articles (as per the comments above), we don't need any of this really.

src/client/article.ts
JamieB-gu: I think so. The reason we want to remove `media.js` is because it's going out of date and missing out on all the changes being made to the main client script 🙂.

There's a possible question mark for ads on media articles. How do ads normally work in galleries?

faresite: I can't get an SSR gallery but in the templates there's only one ad after the 3rd image. Looking at the code here https://github.com/guardian/apps-rendering/blob/815ec3af5c631bc12c1db926caca145f69b0f439/src/ads.tsx#L7, my guess is that we're not handling it the same way in SSR currently and we're gonna have more than one add in the gallery articles. 🤔

JamieB-gu (approved):

PR #1268, opened 2021-04-15
LIVE-2183: Bridget footer links 🦶🔗

faresite:
## Why are you doing this?

At present the 'privacy settings' and 'privacy policy' in the article footers are links. We would like these to call Bridget functions, which the native layer can then use to open the appropriate settings section.
[JIRA ticket](https://theguardian.atlassian.net/browse/LIVE-2183)

## Changes
- create the Navigation bridget client
- add dom element ids to the footer link
- trigger bridget calls on link click
- Bump Bridget to `1.10.0` from https://github.com/guardian/bridget/pull/73 

iOS PR https://github.com/guardian/ios-live/pull/5660

### DIFF
--- a/src/client/article.ts
+++ b/src/client/article.ts
@@ -217,8 +217,8 @@ function renderComments(): void {
 }
 
 function footerLinks(): void {
-	const privacySettingsLink = document.getElementById('ar-privacy-settings');
-	const privacyPolicyLink = document.getElementById('ar-privacy-policy');
+	const privacySettingsLink = document.getElementById('js-privacy-settings');
+	const privacyPolicyLink = document.getElementById('js-privacy-policy');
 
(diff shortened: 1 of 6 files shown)


src/client/article.ts
```diff
@@ -215,12 +216,28 @@ function renderComments(): void {
 	}
 }
 
+function footerLinks(): void {
+	const privacySettingsLink = document.getElementById('ar-privacy-settings');
+	const privacyPolicyLink = document.getElementById('ar-privacy-policy');
```
JamieB-gu, line 221: Would recommend using the `js-` prefix convention.

JamieB-gu (approved):
Nice. One suggestion.

src/client/article.ts
faresite: ah yeah, true! thanks 👍

PR #1270, opened 2021-04-16
Live 2132 remove instagram embed er

abeddow91:
## Why are you doing this?

Editions have decided not to render Instagram embeds in articles.

### DIFF
--- a/src/server/csp.ts
+++ b/src/server/csp.ts
@@ -143,9 +143,7 @@ function buildCspEditions(
 			? 'https://platform.twitter.com https://cdn.syndication.twimg.com'
 			: ''
 	};
-	frame-src https://www.theguardian.com ${
-		thirdPartyEmbed.instagram ? 'https://www.instagram.com' : ''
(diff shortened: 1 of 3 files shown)


src/server/csp.ts
```diff
@@ -143,9 +143,7 @@ function buildCspEditions(
 			? 'https://platform.twitter.com https://cdn.syndication.twimg.com'
 			: ''
 	};
-	frame-src https://www.theguardian.com ${
-		thirdPartyEmbed.instagram ? 'https://www.instagram.com' : ''
-	} https://www.facebook.com https://www.tiktok.com https://interactive.guim.co.uk ${
+	frame-src https://www.facebook.com https://www.tiktok.com https://interactive.guim.co.uk ${
```
JamieB-gu, line 146: Did you mean to drop `www.theguardian.com` from the list of iframeable domains?

JamieB-gu (approved):

src/server/csp.ts
abeddow91: Ah no I didn't, thanks for flagging -  added back in!

PR #1271, opened 2021-04-19
LIVE-2318 : ER Quiz

abeddow91:
## Why are you doing this?

Adds `Design.Quiz` to the Editions article render so it renders as a `standard` article.

## Screenshots

| Before | After |
| --- | --- |
|  |  |


groakland (approved):
LGTM

PR #1272, opened 2021-04-20
Restrict Support To iOS 13 Or Later

JamieB-gu:
## Why are you doing this?

The iOS App only supports iOS 13 or later, so it's not worth targeting anything lower than that in our builds. The [`browserslist`](https://github.com/browserslist/browserslist) is used to inform our build tools which browsers we support.

### Questions

- What does this mean for Editions?

## Changes

- Set `browserslist` to iOS 13 and above

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "server/server.ts",
   "browserslist": [
     "android >= 5",
-    "ios >= 11"
+    "ios >= 13"
   ],
   "scripts": {
     "postinstall": "is-ci || husky install",


marjisound (approved):

faresite:
Last I remember Editions was supporting iOS 9 but that was like half a year back, maybe this is changed now .. @abeddow91 @joecowton1 ?

JamieB-gu:
Closing until we determine support requirements for Editions.

PR #1277, opened 2021-04-26
LIVE-2348: Prefix related item link with theguardian.com

faresite:
## Why are you doing this?

We want the related content items in SSR articles to open in a native view, since there's already a built-in guardian link parser that will open the item in a native view controller, we can use this as we go into prod. in the future we will need to do something about it (Bridget probably), as this means all related items in an SSR article will open up in the legacy templates.

| before | after |
| -- | -- |
|  |  |

### DIFF
--- a/src/relatedContent.test.ts
+++ b/src/relatedContent.test.ts
@@ -54,7 +54,7 @@ describe('parseRelatedContent', () => {
 					title: 'contentTitle',
 					lastModified: undefined,
 					headerImage: undefined,
-					link: '/contentId',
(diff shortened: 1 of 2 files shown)


JamieB-gu (approved):

JamieB-gu:
FYI @SLambrakis @aoifemcl15 @jordanterry

aoifemcl15:
@JamieB-gu  I've created a ticket for us to open apps rendered articles from related items in the future as part of 100% apps rendering [here](https://theguardian.atlassian.net/browse/LIVE-2354).

PR #1280, opened 2021-04-28
LIVE-2138: add conditional

joecowton1:
## Why are you doing this?

Sometimes comment pieces are missing a `contributor` which causes the template not to load, this PR adds a conditional to check for this.

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/editions/avatar/index.tsx
+++ b/src/components/editions/avatar/index.tsx
@@ -29,6 +29,10 @@ const Avatar: FC<Props> = ({ item }) => {
 	const [contributor] = item.contributors;
 	const format = getFormat(item);
 
+	if (!contributor) {
+		return null;
+	}
+
 	return pipe2(
 		contributor.image,
 		map((image) => (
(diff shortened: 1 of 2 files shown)


src/components/editions/avatar/index.tsx
```diff
@@ -29,6 +29,10 @@ const Avatar: FC<Props> = ({ item }) => {
 	const [contributor] = item.contributors;
 	const format = getFormat(item);
 
+	if (!contributor) {
+		return null;
+	}
```
JamieB-gu, line 34: Thanks ESLint... that's not actually true. Although tbf it's actually TS that's being economical with the truth here.

@joecowton1 you might find this useful here:

```ts
const headOption = <A>(arr: A[]): Option<A> =>
  arr.length > 0 ? some(arr[0]) : none;
```

joecowton1: Ah thanks Jamie! Was just scratching my head for a moment there.

JamieB-gu (approved):

PR #1281, opened 2021-04-28
LIVE-2138: add recipe article type to Editions as standard template

joecowton1:
## Why are you doing this?

Recipe article types were being used by production but not included in Editions Rendering. This PR adds `Design.Recipe` to the list of rendered articles to fix that.

## Screenshots

| Before | After |
| --- | --- |
|  |  |


mohammad-haque:
@joecowton1 some of the screenshots looks broken in the `before > after` stage, is that expected?

joecowton1:
> @joecowton1 some of the screenshots looks broken in the `before > after` stage, is that expected?

Yeah they're just being very flakey. Not caused by this PR.

mohammad-haque (approved):

PR #1282, opened 2021-04-28
Live 2351 football scores

joecowton1:
## Why are you doing this?

When developing, we were missing football match report data from the CAPI response as that data is provided separately from PA. 

This PR adds a  'show-references' parameter to the CAPI request to return football club and competition ids, which we can then parse into a new api request to PA, which provides us with match data for the corresponding teams.

We then pass that response into the apps rendering flow with the rest of the content.

We then use the new football scores data to populate a `FootballScores` component that lives in the Editions article header.

* for Editions styles need updating to match new Editions designs (😖 ) but this should provide all necessary functionality.

## Screenshots

| Before | After |
| --- | --- |
|  |  |


abeddow91 (approved):
Looks great!

PR #1288, opened 2021-05-04
LIVE-1249: update layout to match new designs

joecowton1:
## Why are you doing this?

We've recently added the `FootballScores` component to the Editions match report header. This PR creates new Editions specific `FootballScores` components in order to accommodate the different layouts.

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/genericEmbed.tsx
+++ b/src/components/genericEmbed.tsx
@@ -30,19 +30,22 @@ interface Props {
 	embed: Generic;
 }
 
-const GenericEmbed: FC<Props> = ({ embed }) => (
-	<figure css={styles}>
-		<iframe
-			srcDoc={embed.html}
-			title={withDefault('Embed')(embed.alt)}
-			// Prevents scrollbars: covers body margin and random extra 6px
-			height={embed.height + 22}
-		/>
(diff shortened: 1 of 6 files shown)


src/components/genericEmbed.tsx
```diff
@@ -30,19 +30,22 @@ interface Props {
...
-			// Prevents scrollbars: covers body margin and random extra 6px
-			height={embed.height + 22}
-		/>
-		{maybeRender(embed.alt, (alt) => (
-			<figcaption css={captionStyles}>{alt}</figcaption>
-		))}
-	</figure>
-);
+const GenericEmbed: FC<Props> = ({ embed }) => {
+	console.log(embed);
```
abeddow91, line 34: console.log should go

abeddow91 (changes requested):
One console log and a couple of broken tests but otherwise good to go! Nice one!

src/components/genericEmbed.tsx
joecowton1: ooooops

abeddow91 (approved):

PR #1295, opened 2021-05-06
Use Dotcom Links For Related Comment Cards

JamieB-gu:
## Why are you doing this?

An oversight in #1284 meant that "Comment" variations of the related content cards weren't updated to use dotcom links (because they live in a separate component). This remedies that.

## Changes

- Also use dotcom links for comment cards


frankie297 (approved):
Perfect

PR #1297, opened 2021-05-06
LIVE-2368: add new EmailSignup embed type

joecowton1:
## Why are you doing this?

Editions was rendering email-signups as generic embeds. We don't want email-signups in Editions but we need to keep generic embeds - this PR separates out email signups into their own embed type so we can mute them in Editions.

## Changes

- add `EmailSignup` type to `Embeds`
- parse generic embed html to check if iframe `data-component` attribute contains `email-embed`
- mute `EmailSignup` embeds in Editions.

## Screenshots

| Before | After |
| --- | --- |
|  |  |


abeddow91 (approved):
LGTM!

PR #1298, opened 2021-05-06
LIVE-2394: add conditional

joecowton1:
## Why are you doing this?

The`FootballScore` component crashes if we have an empty `scores` field. This PR guards against that, and disables eslint for the relevant lines.

** Ideally we should probably change the types but this is a fix in the meantime


marjisound (approved):

abeddow91 (approved):

joecowton1:
Closing as not needed.

PR #1305, opened 2021-05-11
LIVE-2395: add tiktok to embed types 

joecowton1:
## Why are you doing this?

`TikTok` embeds were lacking a distinct type and always rendered as `Generic` embeds. This PR parses `generic` embeds and separates `TikTok` embeds into their own newly created type.

  
## Changes

- create new `TikTok` embed type
- parse `Generic` embed types
- block embed overlays on `TikTok` embed type

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/embedWrapper.tsx
+++ b/src/components/embedWrapper.tsx
@@ -16,7 +16,15 @@ import {
 import { ClickToView } from 'components/ClickToView';
 import EmbedComponent from 'components/embed';
 import { EmbedKind } from 'embed';
-import type { Embed, Generic, Instagram, Spotify, YouTube } from 'embed';
+import type {
+	EmailSignup,
+	Embed,
+	Generic,
+	Instagram,
+	Spotify,
+	TikTok,
+	YouTube,
+} from 'embed';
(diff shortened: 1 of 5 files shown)


src/components/embedWrapper.tsx
```diff
@@ -332,6 +383,12 @@ const getSourceDetailsForEmbed = (embed: Embed): SourceDetails => {
 	}
 };
 
+const isblockedEditionsEmbed = (embed: Embed): boolean =>
+	embed.kind === EmbedKind.TikTok || embed.kind === EmbedKind.Instagram;
+
+const renderOverlay = (embed: Embed, editions: boolean): boolean =>
+	editions && !isblockedEditionsEmbed(embed);
```
JamieB-gu, line 390: Will this return `false` whenever the product is _not_ Editions, i.e. won't it always return `false` on the live apps? Does this mean the overlay will never render on live apps?

joecowton1: Yeah good point.

abeddow91 (approved):
Great work!!

PR #1228, opened 2021-03-10
LIVE-1126: Emotion 11

faresite:
# Emotiona11

## Changes:
- emotion 11 renames (mostly using the codemod)
- update to image-rendering on `@guardian/types@3`
- changes related to the types 3 update
- updating atoms-rendering to `v12.2.0`
- using custom emotion cache instance

PR #1306, opened 2021-05-12
Stop Using Babel React Transforms

JamieB-gu:
## Why are you doing this?

They're not doing anything any more. As of #1228 we're using the [TypeScript compiler](https://github.com/guardian/apps-rendering/blob/dddeb4ae1f85fa1ec16f31cb8d9ed2d89605707d/tsconfig.json#L9) to handle our React and Emotion transpilation steps. Therefore there's no need to also do this with Babel.

For more information see [the `tsconfig` docs](https://www.typescriptlang.org/tsconfig#jsx), these [other `tsconfig` docs](https://www.typescriptlang.org/tsconfig/#jsxImportSource), and the [React blog post](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) on the new JSX transform.

Thanks to @tjmw for helping me discover this! 🙈

## Changes

- Removed `@babel/preset-react` usage from webpack
- Removed `@babel/preset-react` and `@emotion/babel-plugin` from deps

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,8 +84,6 @@
   "devDependencies": {
     "@babel/core": "^7.12.10",
     "@babel/preset-env": "^7.12.1",
-    "@babel/preset-react": "^7.12.10",
-    "@emotion/babel-plugin": "^11.1.2",
     "@emotion/jest": "^11.2.0",
     "@guardian/eslint-config-typescript": "^0.4.2",
     "@guardian/prettier": "^0.5.0",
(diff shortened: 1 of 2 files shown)


joecowton1:
Great shout, was talking to @sndrs about this earlier in the week (in a wider remit, not just on AR).

joecowton1 (approved):

PR #1307, opened 2021-05-13
LIVE-2148: map generic instagram to structured instagram embeds

joecowton1:
## Why are you doing this?

Ideally we'd like all `Instagram` embeds to render as structured `Instagram` embeds regardless of whether they're inputted as `Generic` embeds or structured embeds. This PR extracts the Instagram post id from a generic embed and creates a structured embed.

Two questions:
- These new embeds don't pass CSP on AR for some reason - why? From what I can tell the embed is exactly as it should be.
- the `parseGenericInstagram` function is currently returning an empty string to appease the `Instagram` type expectation - is there a better way of doing this?

## Changes

- add condition to parse `Instagram` embeds separate from `Generic` embeds
- extract `id` from generic embed `html` field
- map to new structured `Instagram` embed

## To test
- running apps rendering go to `http://localhost:8080/culture/2021/mar/18/nikki-britton-the-10-funniest-things-i-have-ever-seen-on-the-internet`
- page should contain instagram embeds
- go to `http://localhost:8080/culture/2021/mar/18/nikki-britton-the-10-funniest-things-i-have-ever-seen-on-the-internet?editions`
- page should not contain instagram embeds

### DIFF
--- a/src/capi.ts
+++ b/src/capi.ts
@@ -112,11 +112,23 @@ const noThirdPartyEmbeds: ThirdPartyEmbeds = {
 	spotify: false,
 };
 
+const getGenericThirdPartyEmbeds = (
+	thirdPartyEmbeds: ThirdPartyEmbeds,
+	element: BlockElement,
+): ThirdPartyEmbeds => {
+	if (element.embedTypeData?.source === 'Instagram') {
+		return { ...thirdPartyEmbeds, instagram: true };
+	}
+	return thirdPartyEmbeds;
+};
+
 const checkForThirdPartyEmbed = (
 	thirdPartyEmbeds: ThirdPartyEmbeds,
 	element: BlockElement,
 ): ThirdPartyEmbeds => {
 	switch (element.type) {
+		case ElementType.EMBED:
+			return getGenericThirdPartyEmbeds(thirdPartyEmbeds, element);
 		case ElementType.INSTAGRAM:
 			return { ...thirdPartyEmbeds, instagram: true };
 		case ElementType.TWEET:
--- a/src/embed.ts
+++ b/src/embed.ts
@@ -327,9 +327,7 @@ const parseGenericEmbedKind = (parser: DocParser) => (
 
 const extractIdFromInstagramUrl = (url: string): string => {
 	const splitUrl = url.split('/');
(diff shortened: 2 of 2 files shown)


dskamiotis (approved):
Instagram embed successfully loads on AR (with 'blocked content' image) and not on Editions.

JamieB-gu:
> These new embeds don't pass CSP on AR for some reason - why?

Are you talking about Editions specifically? They're permitted in [the `frame-src` directive](https://github.com/guardian/apps-rendering/blob/fb7f33dcd5e83e87ca216a9ee63d14c371a14a17/src/server/csp.ts#L108) for normal AR articles, but not [on Editions](https://github.com/guardian/apps-rendering/blob/fb7f33dcd5e83e87ca216a9ee63d14c371a14a17/src/server/csp.ts#L146) atm. You should just be able to apply the same rule there.

JamieB-gu:
> the `parseGenericInstagram` function is currently returning an empty string to appease the `Instagram` type expectation - is there a better way of doing this?

I think you may want to return an `Err` in the case where the instagram id can't be extracted. That way nothing will be rendered.

joecowton1:
> > the `parseGenericInstagram` function is currently returning an empty string to appease the `Instagram` type expectation - is there a better way of doing this?
> 
> I think you may want to return an `Err` in the case where the instagram id can't be extracted. That way nothing will be rendered.

do you mean like this:
```
const extractInstagramId = (parser: DocParser) => (html: string): string => {
	const instagramID = parseInstagramHTML(parser)(html);

	if (instagramID.kind === ResultKind.Err) {
		return err("I can't parse this html, it has no 'id' field");
	}

	return extractIdFromInstagramUrl(instagramID.value.id);
};
```
I'm getting an `Type 'Err<string>' is not assignable to type 'string'.ts(2322)`, I assume because the return type needs to change now but i'm not sure what it should be?

JamieB-gu:
> i'm not sure what it should be?

`Result<string, string>`

joecowton1:
@JamieB-gu added generic Instagram embeds to the `checkForThirdPartyEmbed` function

JamieB-gu (approved):

PR #1312, opened 2021-05-21
Chromatic: ignore instagram embed iframe

faresite:
## Why are you doing this?

The instagram embed story is very attention seeking and needs approving all the time. Trying this to see if this helps before removing this story.


faresite:
Closing as it's not as simple as that 🤷‍♂️

PR #1322, opened 2021-05-27
LIVE-2490: remove margin

joecowton1:
## Why are you doing this?

We have a margin we don't need on the Editions Rendering templates.

## Changes

- Remove margin


abeddow91 (approved):
LGTM!

PR #1324, opened 2021-05-28
SPIKE: add format params

joecowton1:
## Why are you doing this?

Adding the ability to override `Format` values using query parameters.

This will be required as part of the `4mats` project lead by @faresite https://github.com/guardian/4mats

## Changes

- Add check for `format` query parameter
- Add `formatOverride` function to apply params to capi item responses.

### DIFF
--- a/src/server/formatOverride.ts
+++ b/src/server/formatOverride.ts
@@ -1,0 +1,15 @@
+import type { Item } from 'item';
+
+export const overideFormatValues = (
+	item: Item,
+	formatParams?: string[],
+): Item => {
+	if (formatParams) {
+		const design = parseInt(formatParams[0]);
+		const display = parseInt(formatParams[1]);
+		const theme = parseInt(formatParams[2]);
+
+		return { ...item, design, display, theme };
+	}
+	return item;
+};
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -124,6 +124,7 @@ async function serveArticle(
 	request: RenderingRequest,
 	res: ExpressResponse,
 	isEditions = false,
+	formatParams?: string[],
 ): Promise<void> {
 	const imageSalt = await getConfigValue('apis.img.salt');
 
@@ -136,6 +137,7 @@ async function serveArticle(
 		imageSalt,
 		request,
 		getAssetLocation,
+		formatParams,
 	);
 
 	res.set('Link', getPrefetchHeader(resourceList(clientScript)));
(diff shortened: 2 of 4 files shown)


JamieB-gu:
@joecowton1 is this still going to happen?

joecowton1:
> @joecowton1 is this still going to happen?

Ideally I think so, yes. 

We we interested in whether you had any objections? Sorry, meant to put that in the description.

abeddow91 (approved):
Nice one! This looks great

abeddow91:
@JamieB-gu do you have any objections or opinions on this PR?

src/server/server.ts
```diff
@@ -178,11 +180,12 @@ async function serveArticlePost(
 		const renderingRequest = await mapiDecoder(req.body);
 		const richLinkDetails = req.query.richlink === '';
 		const isEditions = req.query.editions === '';
+		const formatParams = req.query.format as string[] | undefined;
```
JamieB-gu, line 183: Prefer parsing over type casting if possible.

src/server/server.ts
```diff
@@ -237,11 +241,17 @@ async function serveArticleGet(
 				};
 
 				const richLinkDetails = req.query.richlink === '';
+				const formatParams = req.query.format as string[] | undefined;
```
JamieB-gu, line 244: Same comment about type casting.

src/server/formatOverride.ts
```diff
@@ -0,0 +1,15 @@
+import type { Item } from 'item';
+
+export const overideFormatValues = (
+	item: Item,
+	formatParams?: string[],
+): Item => {
+	if (formatParams) {
+		const design = parseInt(formatParams[0]);
+		const display = parseInt(formatParams[1]);
+		const theme = parseInt(formatParams[2]);
```
JamieB-gu, line 10: These could be `NaN`. What happens in that case?

JamieB-gu (commented):
Seems reasonable, but could you elaborate on what these changes are for? Are they only to be used in conjunction with the 4mats project, not in PROD? What does 4mats need them for?

FYI you may run into issues overriding the parsing code like this. For example, `Review` [is typed such](https://github.com/guardian/apps-rendering/blob/fa189bc9040a8530c31913cfcc574f7a14acd297/src/item.ts#L86) that it's only valid if the `starRating` field exists. If you take a normal article and override it with `Design.Review` strange things may happen (including runtime crashes).

joecowton1:
Closing for now at least.

PR #1325, opened 2021-06-02
LIVE-1905: Rich link hyphenation bug on Android

faresite:
## Why are you doing this?

A bug has been found on Android where the rich link is overlapping with the content because of a long word. As shown in this screenshot.

Screenshot 2021-06-02 at 16 12 54

The long word is not at fault here but a hyphenation issue on Chrome for Android that doesn't work if the word is attached to a special character like a colon `:` . Without the special character, the hyphenation works fine as shown here:
Screenshot 2021-06-02 at 16 04 11

I have found an Android chromium issue that seems to be related where hyphenation wasn't working at all https://bugs.chromium.org/p/chromium/issues/detail?id=813414

My feeling is that this is a chromium bug that needs to be fixed and slapping an awful hack to fix it wouldn't be the best option, assuming this doesn't happen very often. Although I'm happy to be challenged at that. As a fix I have introduced `word-wrap: break-word` which gets applied when the hyphenation doesn't work, like in the following screenshot:
Screenshot 2021-06-02 at 16 03 38

As you can see, the hyphenation doesn't work but it does break the word when, leaving other hyphenation working correctly. This is not pretty, but when the chromium bug gets fixed this should work as expected. if I don't find it raised already, I'll try and raise it myself.

Before committing to this solution however I'd welcome some thoughts or better solutions. 

## Changes

- Adds the word-wrap css rule and clamps max width of the headline at 100%.

### DIFF
--- a/src/__snapshots__/storyshots.test.ts.snap
+++ b/src/__snapshots__/storyshots.test.ts.snap
@@ -18245,6 +18245,8 @@ exports[`Storyshots Rich Link Default 1`] = `
   -webkit-text-decoration: none;
   text-decoration: none;
   color: #121212;
+  max-width: 100%;
+  word-wrap: break-word;
 }
 
 .emotion-1 a h1 {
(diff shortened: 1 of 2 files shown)


paperboyo:
Woah! Good to know, great job!

abeddow91 (approved):
LGTM!

PR #1326, opened 2021-06-03
Live 2477 :  Editions Inline JS

abeddow91:
## Why are you doing this?
Editions articles are currently served and downloaded without its js file. There is a [larger body of work](https://theguardian.atlassian.net/browse/LIVE-2483?atlOrigin=eyJpIjoiY2M4ZGFiNjQ3NzY1NDVkMzliODhkNTYyYWUxNzdlOWEiLCJwIjoiaiJ9) going on to bundle the .js with the .html but as a temporary work around we are inlining the js into the article body. As the .js file has a hashed name, we have also introduced a webpack plugin to generate a key/value object from the hashed file names so we can refer to the file dynamically.

### DIFF
--- a/src/server/editionsPage.tsx
+++ b/src/server/editionsPage.tsx
@@ -94,7 +94,6 @@ const buildHtml = (
 	head: string,
 	body: string,
 	scripts: ReactElement,
-	inlineJS: string,
 ): string => `
     <!DOCTYPE html>
     <html lang="en">
@@ -105,7 +104,6 @@ const buildHtml = (
         <body>
 			${body}
 			${renderToString(scripts)}
-			<script>${inlineJS}</script>
         </body>
     </html>
 `;
@@ -136,11 +134,12 @@ function render(
 		<Scripts
 			clientScript={clientScript}
 			twitter={thirdPartyEmbeds.twitter}
+			inlineClientJS={inlineJS}
 		/>
 	);
 
 	return {
-		html: buildHtml(head, body.html, scripts, inlineJS),
+		html: buildHtml(head, body.html, scripts),
 		clientScript: none,
 	};
 }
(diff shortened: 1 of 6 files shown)


src/server/server.ts
```diff
@@ -131,11 +135,25 @@ async function serveArticle(
 		throw new Error('Could not get image salt');
 	}
 
+	let inlineJS = '';
+	if (isEditions) {
+		const hashResponse = await fetch(
+			`https://mobile.code.dev-guardianapis.com/assets/hashed-names.json`,
```
joecowton1, line 141: Do we need to update this to use `mobile.guardianapis.com` ?

webpack.config.ts
```diff
@@ -235,11 +236,20 @@ const clientConfigProduction = {
...
+					jsBundleName: data.assetsByChunkName.editions,
+				});
+			},
+		}),
 	],
 	output: {
 		path: path.resolve(__dirname, 'dist/assets'),
 		publicPath: '',
 		filename: '[name].[contenthash].js',
+
```
joecowton1, line 252: Probably don't need this space

src/server/editionsPage.tsx
```diff
@@ -103,6 +105,7 @@ const buildHtml = (
         <body>
 			${body}
 			${renderToString(scripts)}
+			<script>${inlineJS}</script>
```
joecowton1, line 108: Is it better to use `renderToString` here?  (I don't know the answer to that)

abeddow91: Not sure either, will take a look and see the pros/cons

webpack.config.ts
abeddow91: Removed

src/server/server.ts
abeddow91: Yep, updated!

src/server/editionsPage.tsx
abeddow91: This has been updated so inline js is an optional param for the Scripts component and therefore uses renderToString

PR #1330, opened 2021-06-07
Live 2540 : ER font and script paths

abeddow91:
## Why are you doing this?

ER will be bundling fonts and client js scripts from static S3 files with the .html. As such, the paths need to be unhashed and within the same directory. This PR removes the hash from the script file path and creates a new edition page fonts with the with the correct path.


mohammad-haque (approved):
Looks good to me, lets' deploy to CODE and test the output 👍

PR #1334, opened 2021-06-08
Fix: updated editions.js path

mohammad-haque:
## Why are you doing this?
Editions have a JS file that lives within an asset/js folder, this PR just updates the path to reflect that.

From: `editions.js`
To: `assets/js/editions.js`


abeddow91 (approved):
LGTM

mohammad-haque:
Thanks @abeddow91

PR #1339, opened 2021-06-15
LIVE-2551: fix font size across caption rendering

joecowton1:
## Why are you doing this?

Image captions were incorrect font size.

## Changes

- add correct stying to `figcaption` component
- add correct styling to `credit` component

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/credit.tsx
+++ b/src/components/credit.tsx
@@ -15,9 +15,13 @@ interface Props {
 	format: Format;
 }
 
-const styles = css`
+const mediaStyles = css`
 	${textSans.xsmall()}
 	margin: ${remSpace[1]} 0;
+`;
+
+const defaultStyles = css`
+	${textSans.xxsmall()}
 `;
 
 const Credit: FC<Props> = ({ format, credit }) =>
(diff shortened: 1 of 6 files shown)


src/components/credit.tsx
```diff
@@ -15,20 +15,24 @@ interface Props {
...
 	pipe(
 		credit,
 		map((cred) => {
 			switch (format.design) {
 				case Design.Media:
-					return <p css={styles}>{cred}</p>;
+					return <p css={mediaStyles}>{cred}</p>;
 				default:
-					return <> {cred}</>;
+					return <span css={defaultStyles}> {cred}</span>;
```
JamieB-gu, line 35: What's the reason that one is a `span` and the other a `p`?

JamieB-gu (approved):

src/components/credit.tsx
joecowton1: The existing one was a `p` so i thought there must be a reason for it but I think a `span` makes more sense?

PR #1340, opened 2021-06-17
LIVE-2606: add fixed heights

joecowton1:
## Why are you doing this?

Header media images for Immersive / Interview templates where rendering at the wrong height. 

## Changes

- add fixed height for mobile images
- add image sizes to media queries

## Screenshots

| Before | After |
| --- | --- |
|  |  |


abeddow91 (approved):
nice one

PR #1353, opened 2021-06-24
LIVE-2653: Follow icon

faresite:
## Why are you doing this?

This icon should go into source however as a quick solution the svg is in our codebase.

https://theguardian.atlassian.net/browse/LIVE-2653

@benwuersching @prisalcalde 

## Changes

- Add Follow icon svg
- icon styling

## Screenshots

Screenshot 2021-06-24 at 11 57 05
Screenshot 2021-06-24 at 11 57 00
Screenshot 2021-06-24 at 11 56 46
Screenshot 2021-06-24 at 11 56 41
Screenshot 2021-06-24 at 11 56 29
Screenshot 2021-06-24 at 11 56 22
Screenshot 2021-06-24 at 11 56 07
Screenshot 2021-06-24 at 11 55 57


src/components/follow.tsx
```diff
@@ -25,6 +25,28 @@ const styles = (kicker: string, inverted: string): SerializedStyles => css`
 	border: none;
 	background: none;
 	margin-left: 0;
+	margin-top: 4px;
+
+	svg {
```
JamieB-gu, line 30: Perhaps worth considering a `css` attribute on the `<svg>` over type selectors? Not a blocker.

src/components/follow.tsx
```diff
@@ -25,6 +25,28 @@ const styles = (kicker: string, inverted: string): SerializedStyles => css`
...
+		width: 24px;
+		height: 24px;
+		margin-bottom: -6px;
```
JamieB-gu, line 33: Worth converting these to `rem` for scaling reasons (the scale slider)?

src/components/follow.tsx
```diff
@@ -25,6 +25,28 @@ const styles = (kicker: string, inverted: string): SerializedStyles => css`
...
+		path {
+			fill: #fff;
+		}
+
+		${darkModeCss`
+		circle {
+			fill: ${inverted};
+		}
+		path {
+			fill: #121212;
```
JamieB-gu, line 46: This is `neutral.7` in Source. I'm not sure there's a semantic colour name (`background.inverse` doesn't really work for SVG paths 😬).

src/components/follow.tsx
```diff
@@ -25,6 +25,28 @@ const styles = (kicker: string, inverted: string): SerializedStyles => css`
 	border: none;
 	background: none;
 	margin-left: 0;
+	margin-top: 4px;
```
JamieB-gu, line 28: `remSpace[1]`?

JamieB-gu (commented):
👍 Few small comments.

src/components/follow.tsx
faresite: I checked with the templates (that we're trying to mimic) and they don't scale with the slider so decided to go with px. What do you think?

JamieB-gu: We're not trying to mimic the scaling behaviour of the templates. They _only_ scale body text. We scale everything (with the exception of the headline).

src/components/follow.tsx
```diff
@@ -59,16 +44,42 @@ function getStyles({ theme }: Format): SerializedStyles {
...
-			d="M11.167 9.833c1.216 0 2.65-1.366 2.65-3.033 0-1.667-.984-2.633-2.65-2.633-1.667 0-2.634.966-2.634 2.633s1.55 3.033 2.634 3.033zm6.5 3.667h-1v-2.167H14.5v-1h2.167V8.167h1v2.166h2.166v1h-2.166V13.5zm-6.5-2.333c.716 0 1.4.05 2.016.166.184 1.734 1.55 3.134 3.25 3.434l.734 2.733-.684.667H5.817l-.65-.667 1.316-5 .684-.683c1.333-.45 2.533-.65 4-.65z"
-		/>
-	</svg>
-);
+const FollowIcon: FC<IconProps> = ({ format }) => {
+	const iconStyles = ({ theme }: Format): SerializedStyles => {
+		const { kicker, inverted } = getThemeStyles(theme);
+		return css`
+			width: ${remSpace[6]};
+			height: ${remSpace[6]};
```
JamieB-gu, line 52: Technically these aren't spaces but it probably doesn't matter too much 🙂

JamieB-gu (commented):
There's one `margin-top` that isn't yet resolved, but apart from that looks good.

benwuersching:
@faresite @JamieB-gu Great to see we're fixing this! I just wanted to make sure we're aware that this component comes with two states/icons, i.e. the "not following" state and the "following" state. The two icons are subtly different.
follow_unfollow

faresite:
Thanks @benwuersching for pointing this out. I've now added the following state, here's a gif in light and dark mode.

prisalcalde:
Great @faresite ! Looking good to me.

benwuersching:
Thank you @faresite !

webpack.config.ts
```diff
@@ -64,6 +64,7 @@ function resolve(
 	// We rely on Buffer for our bridget thrift client
 	if (isClient) {
 		resolveOptions.alias = { ...resolveOptions.alias, Buffer: 'buffer' };
+		resolveOptions.fallback = { crypto: false };
```
faresite, line 67: @JamieB-gu webpack stopped compiling client side bundles out of the blue, don't know why .. it was something to do with `image-rendering` and `crypto`.. but I don't understand it as nothing has changed 🤷‍♂️

JamieB-gu: Something has changed, we've added a new component to the `import`s in the client-side bundle. And the import tree results in `image-rendering` being included client-side. I'd suggest breaking `FollowStatus` out into its own file rather than exporting it from `Follow`. That should narrow the import tree enough to exclude `image-rendering`.

src/components/follow.tsx
```diff
@@ -37,6 +44,79 @@ function getStyles({ theme }: Format): SerializedStyles {
 	return styles(kicker, inverted);
 }
 
+const FollowIcon: FC<IconProps> = ({ isFollowing }) => {
+	return (
+		<>
+			{isFollowing && (
```
JamieB-gu, line 50: This might be simpler as:

```tsx
if (isFollowing) {
  return <svg className="following-icon"...>...</svg>
} else {
  return <svg className="follow-icon"...>...</svg>
}
```

But perhaps best not to touch it until merging #1355 to prevent conflicts.

JamieB-gu (approved):

PR #1354, opened 2021-06-24
LIVE-2641: rich links

faresite:
## Why are you doing this?

https://theguardian.atlassian.net/browse/LIVE-2641

## Screenshots

Screenshot 2021-06-24 at 16 06 18
Screenshot 2021-06-24 at 16 06 14
Screenshot 2021-06-24 at 15 59 06
Screenshot 2021-06-24 at 15 59 01

### DIFF
--- a/src/components/richLink.tsx
+++ b/src/components/richLink.tsx
@@ -14,7 +14,7 @@ import { getThemeStyles } from 'themeStyles';
 
 export const richLinkWidth = '8.75rem';
 
-const richLinkPillarStyles = (kicker: string): string => {
+const richLinkPillarStyles = (kicker: string, inverted: string): string => {
 	return `
 		border-top: solid 1px ${kicker};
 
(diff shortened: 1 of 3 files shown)


src/components/richLink.tsx
```diff
@@ -46,40 +46,52 @@ const richLinkStyles = (format: Format): SerializedStyles => {
...
 
 		&.js-sport {
-			${richLinkPillarStyles(sportKicker)}
+			${richLinkPillarStyles(sportKicker, sportInverted)}
 		}
 
-		&.js-culture {
-			${richLinkPillarStyles(cultureKicker)}
+		&.js-culture,
+		&.js-arts {
```
JamieB-gu, line 85: Hold on, how does this class get applied to the element? There is no "arts", it's the old name for "culture".

JamieB-gu (commented):
Looks good, one question about "arts".

src/components/richLink.tsx
faresite: Screenshot 2021-06-24 at 17 09 58

Very good question ... I was puzzled by this too hence my commit message. I believe the class is added via this https://github.com/guardian/apps-rendering/blob/79841c991c6089a781ffcf970476f6abae17a182/src/client/article.ts#L456 
Which makes me think we're getting the data wrong from MAPI/CAPI?

JamieB-gu: If that's just proxying the `pillar` from CAPI then it will give you arts. It's still called that in the CAPI response for legacy reasons. We'll want to map it across to `js-culture` if it discovers `arts` (could do that here or at the MAPI level).

faresite: I'd vote to do that in MAPI, because if we did in here then we have a situation where MAPI is serving legacy stuff, and code here that is trying to fix that. If we fix it in MAPI, we have a more "sound" content flow.

prisalcalde:
Looking good @faresite ! Thank you for fixing that.

JamieB-gu (approved):

PR #1358, opened 2021-06-28
LIVE-2639: Newsletter sign up 

faresite:
## Why are you doing this?

https://theguardian.atlassian.net/browse/LIVE-2639

@prisalcalde @benwuersching I'm aware this is not the desired design, however we don't have much control over it as it's coming from inside an embed .. it's an iframe inside an iframe which is not flexible at all when it comes styling it from our side. The reason it doesn't split on bigger viewports is because our body text doesn't get wide enough for the inner iframe. 
We can definitely create a separate ticket to think about what can be done, potentially from the source. But it's a bigger task than we can do right now, apologies as I understand it seems like it should be straight forward, but it's not. For now, we're just fixing it so it's usable without any overlapping text issues, like in the screenshot.

## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |

### DIFF
--- a/src/client/article.ts
+++ b/src/client/article.ts
@@ -498,15 +498,17 @@ function hydrateClickToView(): void {
 function resizeEmailSignups(): void {
 	const isIframe = (elem: Element): elem is HTMLIFrameElement =>
 		elem.tagName === 'IFRAME';
-	const emailSignupIframe = document.querySelector('.js-email-signup');
-	if (emailSignupIframe && isIframe(emailSignupIframe)) {
-		const innerIframe = emailSignupIframe.contentDocument?.querySelector(
-			'iframe',
-		);
-		if (innerIframe) {
-			innerIframe.style.width = '100%';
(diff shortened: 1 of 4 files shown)


prisalcalde:
That's okay @faresite . We can release it like this. Out of curiosity, which team looks after this component?

src/client/article.ts
```diff
@@ -495,10 +495,25 @@ function hydrateClickToView(): void {
 		);
 }
 
+function resizeEmailSignups(): void {
+	const isIframe = (elem: Element): elem is HTMLIFrameElement =>
+		elem.tagName === 'IFRAME';
+	const emailSignupIframe = document.querySelector('.js-email-signup');
```
JamieB-gu, line 501: Maybe `querySelectorAll`?

JamieB-gu (approved):
One comment about `querySelectorAll`, otherwise looks good.

PR #1360, opened 2021-06-28
LIVE-2666: add dropcaps to editions body

joecowton1:
## Why are you doing this?

We were missing dropcaps on Editions.

## Changes

- add styles

## Screenshots

| Before | After |
| --- | --- |
|  |  |


abeddow91 (approved):
Nice one!

PR #1364, opened 2021-06-30
Live 2632 offline header videos

joecowton1:
## Why are you doing this?

If a user is offline (and the header video is not already cached), the user will see an error message.

This PR creates a hook to check the online status based on this approach: https://levelup.gitconnected.com/useonlinestatus-a-react-hook-to-know-when-your-app-is-offline-2d06e4e536a

It also creates a placeholder component. It currently uses an inline svg, there is an open PR to add the correct icon to Source after which we should import the SVG from there: https://github.com/guardian/source/pull/840

## Changes

- create `useOnlineStatus` hook
- create `Placeholder` component
- render placeholder in `video` component if offline

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/__snapshots__/storyshots.test.ts.snap
+++ b/src/__snapshots__/storyshots.test.ts.snap
@@ -15732,6 +15732,84 @@ exports[`Storyshots Editions/Headline Showcase 1`] = `
 </div>
 `;
 
+exports[`Storyshots Editions/Placeholder Default 1`] = `
+.emotion-0 {
+  width: 100%;
+  position: relative;
+  padding-bottom: 56.25%;
+  background-color: #333333;
+  color: #F6F6F6;
+}
+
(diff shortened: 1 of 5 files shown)


prisalcalde:
Thanks, @joecowton1 . Looking good to me. Let's ask Ben to take a look as well when he's back!

joecowton1:
> Thanks, @joecowton1 . Looking good to me. Let's ask Ben to take a look as well when he's back!

Ok sure! I'm thinking we maybe we merge it - review permitting - just to check it works properly (its hard to test without it being on a device) but v happy to make any changes :)

abeddow91 (approved):
Nice work, looks great!

mohammad-haque (approved):
Nice work! Looks good to me 👍

PR #1368, opened 2021-07-05
LIVE-2679: Fix follow icon cropping

faresite:
## Why are you doing this?

https://theguardian.atlassian.net/browse/LIVE-2679

There were two reasons for this to happen, a) the viewbox was missing so the svg elements didn't know what to scale towards. b) the button min-height didn't accommodate the height of the svg itself.

## Changes

- Add min height to the follow button (Same height as the svg)
- add the svg viewbox

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/__snapshots__/storyshots.test.ts.snap
+++ b/src/__snapshots__/storyshots.test.ts.snap
@@ -17335,6 +17335,7 @@ exports[`Storyshots Follow Default 1`] = `
   background: none;
   margin-left: 0;
   margin-top: 0.25rem;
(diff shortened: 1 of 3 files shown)


JamieB-gu (approved):

prisalcalde:
@faresite from the screenshot on the right, it looks like the bottom is still a little bit cut.

prisalcalde:
Okay, the bottom is another issue. Thanks for letting me know, @faresite .In this case, this one is good to merge.

PR #1373, opened 2021-07-06
LIVE-2566: Prepare for square ads

faresite:
## Why are you doing this?

This PR prepares for square ads format once enabled. Only for first ad in an article.

## Changes

- prepare a boolean flag to turn on square ads
- `.ad-slot-square` styles.

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/client/nativeCommunication.ts
+++ b/src/client/nativeCommunication.ts
@@ -15,6 +15,8 @@ import {
 	videoClient,
 } from '../native/nativeApi';
 
+const TEADS_ENABLED = true;
+
 type Slot = AdSlot | VideoSlot;
 
 function areRectsEqual(rectA: IRect, rectB: IRect): boolean {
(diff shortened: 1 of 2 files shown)


abeddow91 (approved):
LGTM!

src/client/nativeCommunication.ts
```diff
@@ -15,6 +15,8 @@ import {
 	videoClient,
 } from '../native/nativeApi';
 
+const TEADS_ENABLED = true;
```
JamieB-gu, line 18: Could you elaborate on what this flag is for?

JamieB-gu (approved):

src/client/nativeCommunication.ts
faresite: Actually it should be `false` for now. Glad you mentioned it. It's more for other developers in the future if they need to turn on or off the square slots for teads, they just do it using this flag as opposed to by changing the code logic itself.

JamieB-gu: Do you think we could pull that upstream into `mobile-static`/Firebase so that it can be configured remotely?

faresite: I thought about it and still not sure what's the best to do. The reason I went with this approach after all:
a) less complicated
b) it doesn't rely on making a call to an external service to decide if it should render the square slot or not. Which could be quite important when we're serving ads.

What are your thoughts @JamieB-gu ?

JamieB-gu: I think this is reasonable while we're rolling out Teads. But I don't think we necessarily need this flag here after we've turned it on and Teads are in production.

I think if we reckon there's a reasonable chance we may want to turn these off sometimes then we should have a switch in remote config. If we _don't_ think we'll turn them off sometimes then I don't think we need a flag/switch in general.

faresite: In my tenure here we've had to turn on/off different 3rd party providers (including Teads) based on product requirements etc.. The problem with remote switch is we are depending on an external service, but maybe that's not too bad when we're having caching. 
I'm for introducing this as it is to unblock native devs and work on the remote switch at a later date?

PR #1384, opened 2021-07-09
Unknown Parser

JamieB-gu:
## Why are you doing this?

A module for parsing `unknown` values into other TypeScript types. It's generic enough to work for any `unknown` value, but is probably most useful for parsing JSON. It attempts to simplify the parsing process by abstracting the repetitive manual checks behind a declarative-style API, based on parser combinators.

For more details see the included [`parser.md` file](https://github.com/guardian/apps-rendering/blob/12c9ba1baeeb2541c37a318c945c98e70fe5493b/src/parser.md) (I think this module is self-contained enough to justify its own README).

I'd like to try incubating this in AR for a while to see if we find it useful for managing some of our parsing needs (e.g. JSON). If it does prove useful we can lift it up into a library for use in other projects (e.g. `mobile-purchases`).

## Changes

- Added the `parser` module
- Added tests for the `parser` module
- Added README for the `parser` module

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1,0 +1,759 @@
+// ----- Imports ----- //
+
+import type { Option, Result } from '@guardian/types';
+import {
+	err,
+	ok,
+	resultAndThen,
+	ResultKind,
+	resultMap,
+	toOption,
+} from '@guardian/types';
+
+// ----- Types ----- //
+
+interface Parser<A> {
+	run: (a: unknown) => Result<string, A>;
+}
+
+// ----- Core Functions ----- //
+
+const parser = <A>(f: (a: unknown) => Result<string, A>): Parser<A> => ({
+	run: f,
+});
+
+/**
+ * Ignores whatever the input is and instead provides a successful parsing
+ * result containing `a`.
+ *
+ * @param a The value to be contained in the successful parsing result
+ * @returns {Parser<A>} A `Parser` of `A`
+ * @example
+ * const fooParser = succeed('foo'); // Parser<string>
+ *
+ * const result = parse(fooParser)(42); // Ok<string>, with value 'foo'
+ */
+const succeed = <A>(a: A): Parser<A> => parser((_) => ok(a));
+
+/**
+ * Ignores whatever the input is and instead provides a failed parsing
+ * result containing the error string `e`.
+ *
+ * @param e The error string to be provided as the parsing error
+ * @returns {Parser<A>} A `Parser` of `A`
+ * @example
+ * const fooParser = fail('Uh oh!'); // Parser<A>
+ *
+ * const result = parse(fooParser)(42); // Err<string>, with value 'Uh oh!'
+ */
+const fail = <A>(e: string): Parser<A> => parser((_) => err(e));
+
+const isObject = (a: unknown): a is Record<string, unknown> =>
+	typeof a === 'object' && a !== null;
+
+/**
+ * Run a parser over a given input that's currently `unknown`, and attempt
+ * to parse that input into a value of type `A`. If the parsing was successful
+ * the result will be an `Ok<A>`. If it failed the result will be an
+ * `Err<string>`.
+ *
+ * @param pa The parser to run over the input
+ * @param a The `unknown` input to be parsed into type-safe value of type `A`
+ * @returns {Result<string, A>} A parsing result, wrapped in a `Result` type.
+ * @example
+ * const json: unknown = JSON.parse('{ "foo": 42 }');
+ *
+ * const parserA = fieldParser('foo', numberParser); // Parser<number>
+ * const resultA = parse(parserA)(json); // Ok<number>, with value 42
+ *
+ * const parserB = fieldParser('bar', numberParser); // Parser<number>
+ * const resultB = parse(parserB)(json); // Err<string>, with 'missing field' err
+ */
+const parse = <A>(pa: Parser<A>) => (a: unknown): Result<string, A> =>
+	pa.run(a);
+
+// ----- Basic Parsers ----- //
+
+/**
+ * Parses a value into a `string`.
+ */
+const stringParser: Parser<string> = parser((a) =>
+	typeof a === 'string'
+		? ok(a)
+		: err(`Unable to parse ${String(a)} as a string`),
+);
+
+/**
+ * Parses a value into a `number`.
+ */
+const numberParser: Parser<number> = parser((a) =>
+	typeof a === 'number' && !isNaN(a)
+		? ok(a)
+		: err(`Unable to parse ${String(a)} as a number`),
+);
+
+/**
+ * Parses a value into a `Date`.
+ */
+const dateParser: Parser<Date> = parser((a) => {
+	if (typeof a === 'string' || typeof a === 'number' || a instanceof Date) {
+		const d = new Date(a);
+
+		if (d.toString() === 'Invalid Date') {
+			return err(`${String(a)} isn't a valid Date`);
+		}
+
+		return ok(d);
+	}
+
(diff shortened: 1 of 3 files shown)


src/parser.ts
```diff
@@ -0,0 +1,759 @@
...
+ * @returns {Parser<A>} A parser for the value at the given location
+ * @example
+ * const json: unknown = JSON.parse('{ "foo": { "bar": 42 } }');
+ *
+ * const parser = locationParser(['foo', 'bar'], numberParser); // Parser<number>
+ * const result = parse(parser)(json); // Ok<number>, with value 42
+ */
+const locationParser = <A>(location: string[], pa: Parser<A>): Parser<A> =>
+	parser((a) => {
+		if (location.length === 1) {
```
davidfurey, line 212: Sorry, it is Friday afternoon and my brain is struggling slightly to work this out myself.

In this function, it seems like you are recursively building a bunch of functions and executing them at the point when a `parse` is triggered.  Is it possible to the recursive build the bunch of functions when `locationParser` is first called, rather than every time a `parse` is clear.

Apologies if this makes no sense

src/parser.ts
```diff
@@ -0,0 +1,759 @@
...
+	parser((a) => resultAndThen<string, A, B>((x) => f(x).run(a))(pa.run(a)));
+
+const oneOf = <A>(pas: Array<Parser<A>>): Parser<A> =>
+	parser((a) => {
+		const f = (
+			remPas: Array<Parser<A>>,
+			errs: string[],
+		): Result<string, A> => {
+			if (pas.length === 0) {
+				return err(errs.join(' '));
```
davidfurey, line 713: Could display a more helpful message if `pas` was initially empty array (ie `errs` is empty at this stage)

davidfurey (commented):
I haven't reviewed the tests, but have asked a few questions about the implementation.

I think this would be much more useful if you could automatically derive parsers for basic objects.

src/parser.ts
```diff
@@ -0,0 +1,759 @@
...
+	typeof a === 'string'
+		? ok(a)
+		: err(`Unable to parse ${String(a)} as a string`),
+);
+
+/**
+ * Parses a value into a `number`.
+ */
+const numberParser: Parser<number> = parser((a) =>
+	typeof a === 'number' && !isNaN(a)
```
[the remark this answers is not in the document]
JamieB-gu: Yeah, I guess this could go either way. On balance, I think in most cases I've come across it's more useful to treat something that's not a number as a parse failure than a `NaN`. For example, in:

```ts
const a = parseInt('foo');
```

`a` is a `NaN`. It will trickle through your code as such, converting the result of all mathematical operations like `a * 5` into `NaN`s, until it eventually makes its way into a piece of UI: `"NaN comments on this article"`. I'd prefer to catch that early.

src/parser.ts
JamieB-gu: Good point. I've actually got a note to look at this function because I think I've made a bit of a mess of the base cases and error reporting.

As it happens, I don't think the `length === 0` case will _ever_ be hit unless the original array is empty, in which case it won't be hit when recursing through a larger array, in which case it probably doesn't need to live in the recursive function. Therefore, it probably makes sense to pull it into the enclosing scope and report an empty array error as you suggested.

The `errs.join(' ')` should then probably live in the `length === 1` case, because at the moment that's swallowing all prior errors and just reporting the last one.

In summary, the function should work fine, but it could report errors better.

src/parser.ts
JamieB-gu: > it seems like you are recursively building a bunch of functions and executing them at the point when a `parse` is triggered

Yes I think so. In general this library builds a parser from a series of functions. They're only executed when `parse` is given said parser and a value to parse. This allows the parser to be re-used multiple times once it's been built.

> Is it possible to the recursive build the bunch of functions when `locationParser` is first called, rather than every time a `parse` is clear.

Sorry, I'm struggling to understand this sentence 😬🙃

src/parser.ts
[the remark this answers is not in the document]
davidfurey: 👍

src/parser.ts
davidfurey: It should say "Is it possible to recursively build the bunch of functions when locationParser is first called, rather than every time a parse is called?"

src/parser.ts
davidfurey: Indeed, I've no concerns about how the function works, just the error reporting.

JamieB-gu: I've refactored it. `f` now only has the one base case (`length === 0`), and concats the error messages properly.

Regarding an initial empty list, I've opted to include a check at the top-level. Alternatively it could be rolled into the `length === 0` check inside `f`, but I thought it might be clearer to read if separated. What do you think?

src/parser.ts
JamieB-gu: Very good spot! I've refactored it to do just that and added some tests.

It also occurred to me that the case of an empty list of parsers isn't an immediate error. It's a parser for a value that's not nested. See the tests for an example 🙂.

src/parser.ts
davidfurey: I agree the approach you've taken is cleaner

src/parser.ts
davidfurey: That turned out to be a smaller change than I expected.

I don't know if this makes it more/less clear what is happening:
```
const locationParser = <A>(location: string[], pa: Parser<A>): Parser<A> => {
        if (location.length === 0) {
                return pa;
        }
        return fieldParser(
                location[0],
                location.length === 1 ? pa : locationParser(location.slice(1), pa),
        );
}
```

JamieB-gu: Yeah tbf that does fit a more standard pattern for a recursive function.

Hmm, do you actually need that check for length 1 at all?

```ts
const locationParser = <A>(location: string[], pa: Parser<A>): Parser<A> => {
        if (location.length === 0) {
                return pa;
        }

        return fieldParser(location[0], locationParser(location.slice(1), pa));
}
```

which could also be expressed using [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) as:

```ts
const locationParser = <A>(location: string[], pa: Parser<A>): Parser<A> => {
	if (location.length === 0) {
		return pa;
	}

	const [head, ...tail] = location;

	return fieldParser(head, locationParser(tail, pa));
};
```

Is the `slice` version or the destructured version easier to read?

davidfurey: Good point on removing the `=== 1`. I think the destructuring is slightly nicer, but there isn't much in it from my point of view.

src/parser.ts
```diff
@@ -0,0 +1,749 @@
...
+ * const jsonA: unknown = JSON.parse('{ "answer": "a" }');
+ * const resultA = parse(answerParser)(jsonA); // Ok<MultipleChoiceAnswer>, 'a'
+ *
+ * const jsonB: unknown = JSON.parse('{ "answer": "d" }');
+ * const resultB = parse(answerParser)(jsonB); // Err<string>, "Needed 'a', 'b' or 'c'."
+ */
+const andThen = <A, B>(f: (a: A) => Parser<B>) => (pa: Parser<A>): Parser<B> =>
+	parser((a) => resultAndThen<string, A, B>((x) => f(x).run(a))(pa.run(a)));
+
+const oneOf = <A>(parsers: Array<Parser<A>>): Parser<A> =>
```
frankie297, line 696: I know we spoke about in IRL but just wanted to add here, it would be good to add a description/examples like you have done with the other functions.

frankie297 (approved):

PR #1386, opened 2021-07-13
LIVE-2703 : Add Design.Obituary

abeddow91:
## Why are you doing this?

Add the new Obituary design type to the parser and handle type in article rendering for Editions and Apps.


joecowton1 (commented):
Will this cause any complications if when components (Headline, Standfirst etc) go to style an article by `Design.Obituary`?

joecowton1:
> Will this cause any complications if when components (Headline, Standfirst etc) go to style an article by `Design.Obituary`?

Ah i see now it just defaults to standard design.

joecowton1 (approved):

PR #1387, opened 2021-07-13
Rich Link For Narrow Breakpoints

JamieB-gu:
## Why are you doing this?

Currently using the scaling slider to scale up articles results in the text around rich links becoming large and hard to read. We'd like to push rich links to full width beyond a certain scaling threshold to prevent this. One way to achieve this is to use a `rem`-based breakpoint, below which they become full-width.

## Changes

- Added a `rem`-based breakpoint for `mobileMedium`, below which rich links are full width

## Screenshots

| light mode | dark mode |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/richLink.tsx
+++ b/src/components/richLink.tsx
@@ -139,6 +139,10 @@ const richLinkStyles = (format: Format): SerializedStyles => {
 		@media (max-width: 23.4rem) {
 			width: 100%;
 			box-sizing: border-box;
+
+			.js-image {
+				display: none;
+			}
 		}
 		${from.wide} {
 			margin-left: calc(
--- a/src/components/richLink.tsx
+++ b/src/components/richLink.tsx
@@ -1,7 +1,7 @@
 import type { SerializedStyles } from '@emotion/react';
 import { css, jsx as styledH } from '@emotion/react';
 import { remSpace } from '@guardian/src-foundations';
-import { from } from '@guardian/src-foundations/mq';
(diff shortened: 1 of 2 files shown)


src/components/richLink.tsx
```diff
@@ -136,6 +136,10 @@ const richLinkStyles = (format: Format): SerializedStyles => {
 		margin: 0.375rem ${remSpace[4]} ${remSpace[3]} 0;
 
 		width: ${richLinkWidth};
+		@media (max-width: 23.4rem) {
+			width: 100%;
+			box-sizing: border-box;
```
faresite, line 141: We will also need to hide the rich link image here, so it doesn't look like it's part of the story.

faresite (approved):

faresite:
@JamieB-gu this should be ok now

src/components/richLink.tsx
```diff
@@ -139,6 +139,10 @@ const richLinkStyles = (format: Format): SerializedStyles => {
 		@media (max-width: 23.4rem) {
 			width: 100%;
 			box-sizing: border-box;
+
+			.js-image {
+				display: none;
+			}
```
JamieB-gu, line 145: I'd suggest applying a `css` attribute to the div rather than styling on the `.js-image` class. The purpose of `.js-*` classes is to say "hey, this is a hook for some JS code, it's safe to manipulate (add/remove etc.) because there are no styles hanging off it".

PR #1388, opened 2021-07-13
LIVE-2708: hydrate editions video component

joecowton1:
## Why are you doing this?

We we're seeing the desired behaviour from the `useOnlineStatus` hook when templates were rendered.

## Changes

- Hydrate `Video` component

## Screenshots

| Before | After |
| --- | --- |
|  |  |


joecowton1:
> Nice one! We'll need to make sure our static js file is updated in S3 to reflect these changes.

What's the best way to do this?

PR #1389, opened 2021-07-13
LIVE-2788: use params to override theme

joecowton1:
## Why are you doing this?

We need to override article theme styles based on which fronts container they live in. From the perspective of Editions Rendering this means we need to be able to pass a parameter containing the `theme` value from Editions backend, and use that to override the existing article theme value.

## Changes

- get `themeOverride` from query params
- pass value through to article `item` level

### DIFF
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -6,7 +6,18 @@ import type { RelatedContent } from '@guardian/apps-rendering-api-models/related
 import type { RenderingRequest } from '@guardian/apps-rendering-api-models/renderingRequest';
 import type { Content } from '@guardian/content-api-models/v1/content';
 import type { Option, Result, Theme } from '@guardian/types';
-import { either, err, map, ok, OptionKind, withDefault } from '@guardian/types';
+import {
+	either,
+	err,
+	map,
+	none,
+	ok,
+	OptionKind,
+	Pillar,
+	some,
+	Special,
+	withDefault,
+} from '@guardian/types';
 import bodyParser from 'body-parser';
 import { capiEndpoint } from 'capi';
 import compression from 'compression';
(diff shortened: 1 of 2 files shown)


src/server/server.ts
```diff
@@ -205,11 +214,12 @@ async function serveEditionsArticlePost(
 	try {
 		// The "req.body" should contain a 'Content' object which fetched by the
 		// Edition backend from the capi
+		const themeOverride = req.query.theme as Theme | undefined;
```
joecowton1, line 217: @JamieB-gu how do I parse this so i'm not type casting please?

src/server/server.ts
```diff
@@ -185,11 +193,12 @@ async function serveArticlePost(
 		const renderingRequest = await mapiDecoder(req.body);
 		const richLinkDetails = req.query.richlink === '';
 		const isEditions = req.query.editions === '';
+		const themeOverride = req.query.theme as Theme | undefined;
```
joecowton1, line 196: Same here.

src/server/server.ts
```diff
@@ -244,11 +254,17 @@ async function serveArticleGet(
 				};
 
 				const richLinkDetails = req.query.richlink === '';
+				const themeOverride = req.query.theme as Theme | undefined;
```
joecowton1, line 257: And here..

src/server/server.ts
```diff
@@ -185,11 +245,17 @@ async function serveArticlePost(
...
 		if (richLinkDetails) {
 			void serveRichLinkDetails(renderingRequest, res);
 		} else {
-			void serveArticle(renderingRequest, res, isEditions);
+			void serveArticleSwitch(
+				renderingRequest,
+				res,
+				isEditions,
+				themeOverride,
+			);
```
JamieB-gu, line 258: Is this needed in `serveArticlePost`? This function is never triggered for Editions is it? Editions uses `serveEditionsArticlePost`?

joecowton1: Good point, fixed!

JamieB-gu (approved):

PR #1391, opened 2021-07-15
LIVE-2788: fix  parser for theme params

joecowton1:
## Why are you doing this?

We were passing the `theme` query param as an enum but being a query param it became a string enroute to AR. This PR fixes the parser to acomodate that.

## Changes

- parse strings not numbers


joecowton1:
Duplicate, closing.

PR #1392, opened 2021-07-15
LIVE-2708: simplify onlineStatusHook

joecowton1:
## Why are you doing this?

The previous incarnation of the `useOnlineStatus` hook was proving slow and ineffective. This version simplifies it.

## Changes

- simplify `useOnlineStatus` hook

### DIFF
--- a/src/components/editions/utils/useOnlineStatus.tsx
+++ b/src/components/editions/utils/useOnlineStatus.tsx
@@ -1,66 +1,22 @@
 import { useEffect, useState } from 'react';
 
-const PING_RESOURCE = '/ping.txt';
-const TIMEOUT_TIME_MS = 3000;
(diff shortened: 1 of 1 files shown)


faresite:
@joecowton1 Was there a reason you were polling to check that is not needed anymore? or was that not necessary in the first place?

joecowton1:
> @joecowton1 Was there a reason you were polling to check that is not needed anymore? or was that not necessary in the first place?

It was to check for changes in online status, but... I think now we're checking on each new article that's enough.

joecowton1:
@faresite we're seeing a couple of issues with the old implementation so I think if we merge this for now we can address any concerns as we go :)

abeddow91 (approved):
Nice one!

PR #1394, opened 2021-07-15
LIVE-2708: rollback offline video screen

joecowton1:
## Why are you doing this?

We don't want to include the unpredictable video placeholder in our public beta.

## Changes

- remove unpredictable hook
- roll back placeholder ternary


abeddow91 (approved):
LGTM!

PR #1400, opened 2021-07-19
LIVE-2708: hydrate video component for offline placeholder

joecowton1:
## Why are you doing this?

We need to show a placeholder instead of the Video component when Editions users are offline.

This PR fixes the hydration process and simplifies the `useOnlineState` hook. Once merged we will need to update the Editions.js file in S3.

## Changes

- update `@guardian/src-icons` to `3.8.0`
- use classnames for video component styles and pull above client side hydration level so styles exists in SSR template
- pass `atomId` and `title` props to hydration function as values on dom
- simplify `useOnlineStatus` hook
- update `clientScript` path so `editions.js` file lookup  works in dev mode

## Screenshots

 https://user-images.githubusercontent.com/77005274/126204268-3bfb8b63-4b9a-47ae-9570-b1ccf0b0e536.mov


abeddow91 (approved):
Nice one! This looks great

PR #1407, opened 2021-07-22
LIVE-2836: revert changes which break asset path

joecowton1:
## Why are you doing this?

The previous commit https://github.com/guardian/apps-rendering/pull/1404 broke the asset paths for Editions. This commit reverts those changes.


groakland (approved):
revert of asset paths

PR #1409, opened 2021-07-27
LIVE-2848: update letters styling

joecowton1:
## Why are you doing this?

Headline styling was broken on the Letters template, standfirst styling was broken for both letters and analysis.

Notes: 
- there is no 400 weight headline style in Source for the Letters headline, so have gone with 300.
- think the CI errors will probably be fixed by: https://github.com/guardian/apps-rendering/pull/1408#pullrequestreview-715711646 - does this sound correct @faresite ?

## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |


abeddow91 (commented):
It looks like all of the snapshots have been deleted (or a good amount at least). Any reason for this? PR doesn't look like it should be doing that..

src/components/editions/headline/index.tsx
```diff
@@ -42,6 +42,11 @@ const commentStyles = css`
 	padding-right: 6rem;
 `;
 
+const lStyles = css`
+	padding-bottom: 0;
+	padding-right: 6rem;
+`;
+
```
abeddow91, line 49: Doesn't look like this is being used, can we remove it?

joecowton1: Good spot 👍

abeddow91 (approved):
Looks good to go!

PR #1410, opened 2021-07-27
Chromatic tweaks

faresite:
## Why are you doing this?

Chromatic kept on asking to approve false positive changes that were a result of chromatic's quality threshold, this PR changes the tolerance to 0.4 as that was the closest tolerance I found that wouldn't give false positives.

This PR also addresses the fact the instagram embed diff kept on appearing because the post was quite popular and the likes amount kept on changing. I've chosen a post that is a lot less popular, so hopefully this won't happen as often, however it will happen from time to time.

## Changes

- Change chromatic diff tolerance to 0.4
- Change the instagram post id

## Tolerance

| **0 tolerance** |
| --- |
|  |

| **0.3 tolerance** |
| --- |
|  |

| **0.35 tolerance** |
| --- |
|  |

### DIFF
--- a/src/__snapshots__/storyshots.test.ts.snap
+++ b/src/__snapshots__/storyshots.test.ts.snap
@@ -17545,7 +17545,7 @@ html:not(.src-focus-disabled) .emotion-4:focus {
     className="js-click-to-view-container"
     data-caption="a caption"
     data-editions="false"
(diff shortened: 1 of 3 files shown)


joecowton1 (approved):
Good idea!

PR #1416, opened 2021-07-29
LIVE 2857 & 2864:  Editorial template

abeddow91:
## Why are you doing this?
Editorial articles were rendering incorrectly because they were being parsed as a Comment. Whilst this is ok for AR, as they share designs, ER visually distinguishes between Editorial and Comment.  

In investigating this, a bug was discovered whereby Letter and Editorial were not always styled the same as Comment (which they should be). This PR fixes this too. 

## Changes

- Move Editorial above Comment in parser to style Editorial differently in ER.
- Add Editorial and Letter where Comment styling is being applied in AR only. 

## ER

| Before | After |
| --- | --- |
|  |  |

### Editorial

| Before | After |
| --- | --- |
|  |  |

### Letter

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/body.tsx
+++ b/src/components/body.tsx
@@ -60,7 +60,11 @@ const Body: FC<Props> = ({ item, shouldHideAds }) => {
 		return <Interactive>{renderAllWithoutStyles(item, body)}</Interactive>;
 	}
 
-	if (item.design === Design.Comment || item.design === Design.Letter) {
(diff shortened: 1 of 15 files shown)


joecowton1 (approved):
Looks good!

JamieB-gu (approved):
Nice!

PR #1417, opened 2021-07-29
LIVE-2859 : Add Corrections and Clarifications Header

abeddow91:
## Why are you doing this?
Corrections and Clarifications need a different header without a byline on Editions. This PR checks for the Corrections and Clarifications tag and renders a specific header if true. 

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/editions/header/index.tsx
+++ b/src/components/editions/header/index.tsx
@@ -12,7 +12,7 @@ import Lines from 'components/editions/lines';
 import Series from 'components/editions/series';
 import Standfirst from 'components/editions/standfirst';
 import type { Item } from 'item';
-import { isPicture } from 'item';
+import { isCorrection, isPicture } from 'item';
 import type { FC, ReactElement } from 'react';
 import {
 	articleMarginStyles,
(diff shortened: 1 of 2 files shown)


src/components/editions/header/index.tsx
```diff
@@ -252,7 +252,18 @@ const LetterHeader: FC<HeaderProps> = ({ item }) => (
...
+		<HeaderMedia item={item} />
+		<Series item={item} />
+		<Headline item={item} />
+		<Standfirst item={item} />
+	</header>
+);
+
 const renderArticleHeader = (item: Item): ReactElement<HeaderProps> => {
+	console.log(item);
+	console.log();
```
JamieB-gu, line 266: ?

JamieB-gu (commented):
Whilst this change seems innocuous, I'm cautious about us returning to the pattern of making styling and layout changes in the UI code based on arbitrary tags. This is essentially what `frontend` and the old templates were doing, and unfortunately it becomes difficult to predict quite quickly. The reason `Format` was created was to avoid the issues that stem from this.

I notice that this is also happening on line 272 with the `isPicture` function. Can we have a chat about what the design requirements are here?

src/components/editions/header/index.tsx
abeddow91: Yep - removed!

abeddow91:
Totally get that. For this is was that comments and clarifications don't have a byline on Editions. Of course, within Format they parse as a `article` so have a byline. 

For cartoons/pictures, I'd need to refresh myself with the code but I think it's based around it needing to be full width if it's a cartoon. I'm not working tomorrow but maybe we can chat through both on Monday?

abeddow91:
closing as replaced by https://github.com/guardian/apps-rendering/pull/1433

PR #1421, opened 2021-08-01
LIVE-2875: hide share when no web url

joecowton1:
## Why are you doing this?

We have repeated user feedback from users trying to share articles and the button doing nothing. This PR copies the logic from Mallard so a share icon will not be rendered when there is not weburl present.

Needs testing.

## Changes

- add `shareIcon` checkbox to relevant stories
- add `showShareIcon` logic to `standfirst` and `byeline`
- add min-height to byeline to fix height when icon is not there

## Screenshots

| Before | After |
| --- | --- |
|  |  |


joecowton1:
Closing as no longer needed.

PR #1425, opened 2021-08-03
LIVE-2860 :  Pinch and Zoom

abeddow91:
## Why are you doing this?
Users on Editions want to be able to pinch and zoom on articles. This was previously disabled for all articles. This PR adds a metatag to allow zooming on editions articles only.


joecowton1 (approved):
LGTM!

PR #1426, opened 2021-08-03
LIVE-2836: add console logs

joecowton1:
## Why are you doing this?

We need visibility on why fonts  / js aren't loading in the Editions app.

## Changes

- add console.logs


joecowton1:
Closing as no longer required

PR #1429, opened 2021-08-04
LIVE-2875: hide shareIcon if no webUrl

joecowton1:
## Why are you doing this?

We have repeated user feedback from users trying to share articles and the button doing nothing. This PR copies the logic from Mallard so a share icon will not be rendered when there is not weburl present.

## Changes

- add `shareIcon` checkbox to relevant stories
- add `showShareIcon` logic to `standfirst` and `byeline`
- add min-height to byeline to fix height when icon is not there

## Screenshots

| Before | After |
| --- | --- |
|  |  |


abeddow91 (approved):
Nice one!

PR #1430, opened 2021-08-04
LIVE-2892: remove headline padding when no avatar

joecowton1:
## Why are you doing this?

Styles were wrong on comment articles.

## Changes

- add check for avatar image to headline component

## Screenshots

| Before | After |
| --- | --- |
|  |  |


abeddow91 (approved):
Nice one!

PR #1437, opened 2021-08-06
Correctly `await` Promises In Express

JamieB-gu:
### TL;DR

We weren't handling `Promise` rejection correctly, which caused requests to hang on error.

### In Full

We had a recent bug in Editions, which we traced back to a hanging request in the group of requests the Editions backend makes to AR. I believe this issue would also have an impact on the live apps, although perhaps with less severe consequences.

As seen in the temporary fix in #1436, it was triggered by a crash in the rendering code. However, that should have resulted in a rapid 500 response rather than a hanging request.

I believe what happened was that the crash caused a `Promise` to be rejected, and we weren't handling that rejection. As a result, the code path that catches errors and sends an Express response was not hit, and so the request stayed open indefinitely.

I suspect we didn't realise this because our `server` file contains some `async`/`await` code mixed in with some `Promise` code, and this caused some confusion about how exactly errors were being caught. When using `async`/`await`, any `Promise`s have to be explicitly `await`ed for them to trigger an exception that can be understood by the `try...catch` block. Allowing them to run normally doesn't work, and results in `Promise` rejections going unhandled. At the moment, in Express 4, you have to explicitly handle all `Promise` rejections, either by catching and writing a response or by passing errors to the `next` function. It looks like [in Express 5](https://expressjs.com/en/guide/error-handling.html) returning rejected `Promise`s from the handler function will be handled automatically 🤞.

**Note:** I think a few other improvements could be made to mitigate future problems like this:
1. We should tidy up our `server` code, it's not easy to understand.
2. The Editions backend could handle request timeouts more gracefully, so they don't take down the entire build.
3. We should set a timeout on the AR Express server, so it doesn't tie up failed connections indefinitely.

## Changes

- Explicitly `await` `Promise`s in our `server` file so that responses are sent correctly on error

### DIFF
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -295,8 +295,8 @@ async function serveArticleGet(
 		const isEditions = req.query.editions === '';
 		const capiContent = await askCapiFor(articleId);
 
-		either(
-			(errorStatus: number) => {
+		await either(
+			async (errorStatus: number) => {
 				res.sendStatus(errorStatus);
 			},
 			async ([content, relatedContent]: [Content, RelatedContent]) => {
@@ -317,9 +317,9 @@ async function serveArticleGet(
 				const themeOverride = themeFromUnknown(req.query.theme);
 
 				if (richLinkDetails) {
-					void serveRichLinkDetails(mockedRenderingRequest, res);
+					await serveRichLinkDetails(mockedRenderingRequest, res);
 				} else {
-					void serveArticleSwitch(
+					await serveArticleSwitch(
 						mockedRenderingRequest,
 						res,
 						isEditions,
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -250,9 +250,9 @@ async function serveArticlePost(
 		const themeOverride = themeFromUnknown(req.query.theme);
 
 		if (richLinkDetails) {
-			void serveRichLinkDetails(renderingRequest, res);
+			await serveRichLinkDetails(renderingRequest, res);
 		} else {
-			void serveArticleSwitch(
+			await serveArticleSwitch(
 				renderingRequest,
 				res,
 				isEditions,
@@ -279,7 +279,7 @@ async function serveEditionsArticlePost(
 		};
 		const themeOverride = themeFromUnknown(req.query.theme);
 
-		void serveEditionsArticle(renderingRequest, res, themeOverride);
+		await serveEditionsArticle(renderingRequest, res, themeOverride);
 	} catch (e) {
 		logger.error('This error occurred', e);
 		next(e);
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -296,8 +296,9 @@ async function serveArticleGet(
 		const capiContent = await askCapiFor(articleId);
 
 		await either(
-			async (errorStatus: number) => {
+			(errorStatus: number) => {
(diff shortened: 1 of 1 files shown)


jonathonherbert:
This makes lots of sense. Sadly, it looks like this will be difficult to test for without testing all the possible asynchronous paths via some involved mocking or integration.

Re: express 5, it looks like the handler will have a more elegant way of dealing with errors via the middleware, but I don't think this would defend against the bug in question here. IIUC a Promise that's left dangling won't be (could _not_ be?) automagically picked up by a caller unless it's explicitly added to the promise chain, either via an `await`, by chaining on the promise with `.then` or similar, or by returning  it.

jonathonherbert:
Re: 2., it's my feeling that it's probably better to fail fast (or in this case, fast-ish!) at proof or publish time rather than gracefully ignore unrenderable content.

If we cannot publish the entirety of the content in a front, for example, we should probably know about it as soon as possible – or we risk silently publishing something that could be incomplete (perhaps, empty) without our editorial staff knowing about it.  

The alternative might be some sort of alarm, but IIUC because proofing is a manual process, errors will always result in an issue being raised to e.g. 24/7, which is as good as an alarm as I can imagine for this sort of problem.

I suppose I can't think of any sort of mitigation right now that would mean we could safely ignore these sorts of issues rather than failing quickly, because their scope is so broad: 

- a missing article – bad, but maybe we can get away with it and republish?
- many missing articles – oh boy
- a missing crossword – cut to scenes from Armaggedon

... that an immediate intervention is pretty much always warranted.

jonathonherbert (approved):
Looks good, and I can't see any more `void`s in the server part of the codebase. In the absence of a test for these sorts of problems, I wonder if a linting rule might be useful to prevent dangling promises.

(Probably not – I can't think of an example in server code where you'd want to ignore the result of asynchronous work, but let's face it, that example is probably out there! I just can't think of a decent automated way to prevent this from happening again 😢 )

joecowton1 (approved):
Thanks for the long explanation Jamie, makes sense.

JamieB-gu:
Thanks for the comments Jon!

> Re: 2., it's my feeling that it's probably better to fail fast (or in this case, fast-ish!) at proof or publish time rather than gracefully ignore unrenderable content.

Yeah that seems sensible. Perhaps that's actually happening already given that the build did fail. Possibly we just need to work on the error reporting so we're more clear on _why_ the build failed 🤔.

> Re: express 5, it looks like the handler will have a more elegant way of dealing with errors via the middleware, but I don't think this would defend against the bug in question here. IIUC a Promise that's left dangling won't be (could _not_ be?) automagically picked up by a caller unless it's explicitly added to the promise chain, either via an `await`, by chaining on the promise with `.then` or similar, or by returning it.

I think ideally we'd want to go with this final option, i.e. return `Promise`s from the callbacks and have any rejections caught by the middleware as mentioned.

> I wonder if a linting rule might be useful to prevent dangling promises.

I have a feeling it was a lint rule that flagged these originally and recommended prefixing with `void`, possibly it was [this one](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md). Anyway, as you say in some cases prefixing with `void` may be what you want, but not here 😬.

PR #1438, opened 2021-08-06
Docs For `oneOf`

JamieB-gu:
## Why are you doing this?

There were some docs missing for the `oneOf` function introduced in #1384. This adds those docs.

## Changes

- Added docs for the oneOf parser function


frankie297 (approved):
lovely

PR #1440, opened 2021-08-09
LIVE-2906: hide share icons without url

joecowton1:
## Why are you doing this?

Currently on Editions we're rendering shareIcons on every article without checking if that article exists. This PR adds event listeners to listen to a ping from Mallard containing the webUrl of an article on a message in the form of:

```
{kind: "ShareIcon", value: "[weburl]"}
```
if that `value` is empty we then hide the shareIcon.

## Changes

- upgrade Renditions to give access to new message types and functions
- add shareIcon listeners
- remove outdated stories

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6052,10 +6052,11 @@
       "dev": true
     },
     "@guardian/renditions": {
-      "version": "0.1.4",
-      "resolved": "https://registry.npmjs.org/@guardian/renditions/-/renditions-0.1.4.tgz",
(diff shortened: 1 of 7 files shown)


JamieB-gu (approved):
Looks good. One question.

Why is there such a large `package-lock.json` change from bumping `renditions`? Are these all the transitive deps from `np` @abeddow91? Should it be a dev dependency of `renditions` instead?

abeddow91:
Ah good spot, yes it should be. I'll amend that and bump here.

PR #1443, opened 2021-08-11
update node version to 14.16

marjisound:
## Why are you doing this?
This PR updates node version from **`12.13`** to **`14.16`** and also updates github actions **`setup-node`** step to use **nvmrc** 

## Changes

- updates node version from `12.13` to `14.16`
- use `guardian/actions-setup-node` instead of `setup-node` in github actions in order to install node based on nvmrc 
- Update TeamCity steps to use nvmrc as the node version (https://github.com/guardian/apps-rendering/pull/1448)

## TODO: 
- Update AMI node version


JamieB-gu:
> * updates github actions `setup-node` step to use nvmrc

How does `setup-node` know to use `.nvmrc`?

marjisound:
> > * updates github actions `setup-node` step to use nvmrc
> 
> How does `setup-node` know to use `.nvmrc`?

Thanks @JamieB-gu for spotting this. I updated the action library to `guardian/actions-setup-node`. The original action `actions/setup-node` won't use nvmrc for node version and it recommends that we give node version to it via its `node-version` parameter while `guardian/actions-setup-node` would use nvmrc node version if it exists.

JamieB-gu (approved):

PR #1446, opened 2021-08-12
Add ingress security group rule for loadbalancer

marjisound:
## Why are you doing this?
This PRs updates AR loadbalancer security group to allow inbound traffic from the lambda within the same VPC.

## Changes

- update cloudformation loadbalancer security rule

### DIFF
--- a/config/cloudformation.yaml
+++ b/config/cloudformation.yaml
@@ -65,6 +65,12 @@ Parameters:
   DomainName:
     Type: String
     Description: The DomainName of the ALB, should contain the trailing dot stuff.zone.example.com.
+  EditionsPreviewLambdaSecurityGroupId:
+    Type: String
+    Description: A security group id for editions preview lambda to access AR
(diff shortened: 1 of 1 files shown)


config/cloudformation.yaml
```diff
@@ -65,6 +65,12 @@ Parameters:
   DomainName:
     Type: String
     Description: The DomainName of the ALB, should contain the trailing dot stuff.zone.example.com.
+  EditionsPreviewLambdaSecurityGroupId:
+    Type: String
+    Description: A security group id for editions preview lambda to access AR
+  EditionsPublishedLambdaSecurityGroupId:
+    Type: String
+    Description: A security group id for editions published lambda to access AR
```
philmcmahon, line 73: My only suggestion would be to include the word 'backend' in the name or description as there are a lot of different editions lambdas - 'editions backend published lambda'

philmcmahon (approved):
Nice, I forgot that there are 2 lambdas!

PR #1449, opened 2021-08-12
Remove Paragraph From `Bullet` Component

JamieB-gu:
## Why are you doing this?

The `Bullet` component assumed too much about its context. It included a `p` tag, which caused problems when the component was nested inside _another_ paragraph. Paragraphs are [not allowed](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p) inside other paragraphs, so the browser was not able to render this combination correctly. See screenshots below for the result.

## Changes

- Removed paragraph from `Bullet` component

## Screenshots

| Article | Before | After |
| --- | --- | --- |
| [Article](https://www.theguardian.com/news/2021/jul/28/corrections-and-clarifications) (Editions-Rendering) | bullet-before | bullet-after |
| [Article](https://www.theguardian.com/us-news/2020/jan/21/why-havent-you-shut-down-the-border-inside-trumps-white-house) (Apps-Rendering) | bullet-alt-before | bullet-alt-after |

**Note:** The second example is more subtle, but there _is_ a difference in font size if you look closely.

### DIFF
--- a/src/components/bullet.tsx
+++ b/src/components/bullet.tsx
@@ -2,8 +2,6 @@
 
 import type { SerializedStyles } from '@emotion/react';
 import { css } from '@emotion/react';
-import { remSpace } from '@guardian/src-foundations';
-import { body } from '@guardian/src-foundations/typography';
 import type { Format } from '@guardian/types';
 import type { FC } from 'react';
 import { darkModeCss } from 'styles';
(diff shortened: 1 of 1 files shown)


joecowton1 (approved):
Great work!!

abeddow91 (approved):

joecowton1 (approved):

PR #1451, opened 2021-08-12
LIVE-2906: hide byline

joecowton1:
## Why are you doing this?

Now we're hiding the shareIcon when article share is not available we were getting UI bugs as if there was no byline copy. In the old Editions app we simply hid the byline component, this PR implements both the non-rendering of byline components and the hiding of the header border bottom if there is no byline.

It also fixes a minor bug where use of `h2` text was breaking the border-right on the article body.

## Changes

- hide byline component if byline is not present 
- hide header border bottom if byline  is not present
- swap margin for padding on editions body `h2` tag (production use this in draft articles sometimes :( ).

## Screenshots

| Before | After |
| --- | --- |
|  |  |
|  |  |

### DIFF
--- a/src/components/editions/article/index.tsx
+++ b/src/components/editions/article/index.tsx
@@ -5,8 +5,8 @@ import { css } from '@emotion/react';
 import { remSpace } from '@guardian/src-foundations';
 import { from } from '@guardian/src-foundations/mq';
 import { background, border, neutral } from '@guardian/src-foundations/palette';
-import type { Format } from '@guardian/types';
-import { Design, Display, partition } from '@guardian/types';
+import type { Format, Option } from '@guardian/types';
+import { Design, Display, OptionKind, partition } from '@guardian/types';
 import type { Item } from 'item';
 import { isPicture } from 'item';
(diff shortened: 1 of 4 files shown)


abeddow91 (approved):
Great work with this!

src/components/editions/article/index.tsx
```diff
@@ -44,15 +44,23 @@ const articleStyles = css`
 	${articleMarginStyles}
 `;
 
-const headerStyles = css`
-	${articleWidthStyles}
-	border-bottom: 1px solid ${border.secondary};
+const hasChildNodes = (field: Option<DocumentFragment>): boolean => {
+	return field.kind === OptionKind.Some && field.value.hasChildNodes();
+};
```
JamieB-gu, line 49: Possible abbreviation:

```suggestion
const hasChildNodes = (field: Option<DocumentFragment>): boolean =>
	field.kind === OptionKind.Some && field.value.hasChildNodes();
```

JamieB-gu (approved):
Couple minor suggestions, not blockers 👍

PR #1453, opened 2021-08-16
Tidy Directories - Part One

JamieB-gu:
## Why are you doing this?

Tidying up some of our directory structure before we migrate into DCR.

## Changes

- Moved `parserContext` to top level
- Moved `interactiveAtom` to `components` directory


joecowton1 (approved):
Few linting errors but otherwise looks good..

PR #1459, opened 2021-08-20
LIVE-2961: hide quote marks on editions

joecowton1:
## Why are you doing this?

Sometimes Editorial use block quotes for poetry.  We should double check we don't need these anywhere else in the body in Editions.

## Changes

- Don't render `Blockquote` component on Editions

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -196,13 +196,13 @@ const plainTextElement = (node: Node, key: number): ReactNode => {
 	}
 };
 
-const textElement = (format: Format, supportsDarkMode = true) => (
+const textElement = (format: Format, isEditions = false) => (
 	node: Node,
 	key: number,
 ): ReactNode => {
 	const text = node.textContent ?? '';
 	const children = Array.from(node.childNodes).map(
-		textElement(format, supportsDarkMode),
+		textElement(format, isEditions),
 	);
 	switch (node.nodeName) {
 		case 'P':
@@ -218,7 +218,7 @@ const textElement = (format: Format, supportsDarkMode = true) => (
 					href: withDefault('')(getHref(node)),
 					format,
 					key,
-					supportsDarkMode,
+					isEditions,
 				},
 				transform(text, format),
 			);
@@ -231,7 +231,7 @@ const textElement = (format: Format, supportsDarkMode = true) => (
 						children,
 				  );
 		case 'BLOCKQUOTE':
-			return supportsDarkMode
+			return !isEditions
 				? h(Blockquote, { key, format }, children)
 				: h('blockquote', { key }, children);
 		case 'STRONG':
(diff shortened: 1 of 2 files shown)


src/renderer.ts
```diff
@@ -196,10 +196,11 @@ const plainTextElement = (node: Node, key: number): ReactNode => {
...
 };
 
-const textElement = (format: Format, supportsDarkMode = true) => (
-	node: Node,
-	key: number,
-): ReactNode => {
+const textElement = (
+	format: Format,
+	supportsDarkMode = true,
+	isEditions = false,
```
JamieB-gu, line 202: Given that `supportsDarkMode` was introduced to handle the fact that Editions doesn't, should we combine both these flags into the `isEditions` flag and just pass that to `Anchor`?

joecowton1: Given that `supportsDarkMode` is used everywhere, I've changed the blockquote renderer code just to use this rather than replace it throughout with `isEditions`

JamieB-gu: Wait, so we use `supportsDarkMode` to decide whether to show quote marks in light mode?

joecowton1: When you put it like that it sounds kind of silly :) Fixed now so it's `isEditions` throughout.

src/renderer.ts
```diff
@@ -231,7 +231,9 @@ const textElement = (format: Format, supportsDarkMode = true) => (
 						children,
 				  );
 		case 'BLOCKQUOTE':
-			return h(Blockquote, { key, format }, children);
+			return !isEditions
+				? h(Blockquote, { key, format }, children)
+				: h('blockquote', { key }, children);
```
JamieB-gu, line 236: Nitpick. Might be preferable to swap the order to avoid the `!`, i.e.:

```ts
isEditions ? h('blockquote', ...) : h(Blockquote, ...);
```

Just slightly less cognitive load for the reader.

JamieB-gu (approved):

src/renderer.ts
JamieB-gu: Nice. You know, given the migration I think it might be that we end up with a `Platform` type to handle these minor UI changes on different platforms. Something like:

```ts
enum Platform {
  Apps,
  Editions,
  Dotcom,
  Amp,
}
```

Then you can do stuff like:

```ts
const styles = (platform: Platform): SerializedStyles => {
  switch (platform) {
    case Platform.Apps:
      return someDarkModeStyles;
    case Platform.Editions:
      return someSpecialEditionsStyles;
    default:
      return someWebStyles;
  }
}
```

In a given component you could also swap in entirely different sub-components depending on the platform you're currently rendering.

src/renderer.ts
joecowton1: Good shout, fixed.

PR #1462, opened 2021-08-23
LIVE-2976: fix styling

joecowton1:
## Why are you doing this?

We've updated analysis header styling.

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/editions/headline/index.tsx
+++ b/src/components/editions/headline/index.tsx
@@ -44,6 +44,7 @@ const analysisStyles = css`
 	background-clip: content-box;
 	background-origin: content-box;
 	padding-bottom: ${remSpace[2]};
+	padding-right: 0;
 
 	${from.mobile} {
 		line-height: 48px;
--- a/src/components/editions/headline/index.tsx
+++ b/src/components/editions/headline/index.tsx
@@ -38,7 +38,7 @@ const analysisStyles = css`
 		transparent 46px,
 		${neutral[86]}
 	);
-	line-height: 48px;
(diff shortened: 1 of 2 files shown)


abeddow91:
It looks like there is too much padding-right. Could we remove it or override with padding-right 0 so the lines are aligned?

abeddow91 (commented):
Slight tweak needed to the right padding. It looks like there is too much padding-right. Could we remove it or override with padding-right 0 so the lines are aligned? Otherwise looks good to me!

joecowton1:
> Slight tweak needed to the right padding. It looks like there is too much padding-right. Could we remove it or override with padding-right 0 so the lines are aligned? Otherwise looks good to me!

Good shout - should be fixed now :)

abeddow91 (approved):

src/components/editions/headline/index.tsx
```diff
@@ -31,11 +31,28 @@ const tablet = tabletContentWidth + 12;
...
-	text-decoration-color: ${kickerColor};
-	padding-bottom: 0;
+const analysisStyles = css`
+	background-image: repeating-linear-gradient(
+		to bottom,
+		transparent,
+		transparent 46px,
+		${neutral[86]}
+	);
+	line-height: 48px;
```
JamieB-gu, line 41: I'd suggest `rem`s for this to avoid scaling issues (assuming Editions uses the same scaling mechanism as AR).

joecowton1: Updated.

PR #1467, opened 2021-08-25
Live 2970 : Header image caption

abeddow91:
## Why are you doing this?
Sometimes the header image caption has `em` tags nested within it. At the moment, ER renders this as a string so the tags are visible and not taken into account during rendering. This PR shifts to rendering the caption from document fragments instead so we render the html tags. 

## Screenshots

| Before | After |
| --- | --- |
|  |  |


joecowton1 (commented):
Nice work! I think this might have broken all of the storybook stories though - do we need to change the prop names or something?

src/components/editions/headerImageCaption.tsx
```diff
@@ -79,14 +80,26 @@ const svgStyle = (iconColor?: string): SerializedStyles => css`
...
 	styles?: SerializedStyles;
 	iconColor?: string;
 	iconBackgroundColor?: string;
 	isFullWidthImage: boolean;
 }
 
+const toReact = (node: Node, key: number): ReactNode => {
+	switch (node.nodeName) {
+		case 'EM':
+			return h('em', { key }, node.textContent);
```
JamieB-gu, line 94: You're in a `tsx` file so you could do this if you prefer?

```tsx
<em key={key}>{node.textContent}</em>
```

src/components/editions/headerImageCaption.tsx
```diff
@@ -79,14 +80,26 @@ const svgStyle = (iconColor?: string): SerializedStyles => css`
...
+		case '#text':
+			return node.textContent;
+	}
+};
+
+const renderText = (byline: DocumentFragment): ReactNode =>
+	Array.from(byline.childNodes).map((node, i) => toReact(node, i));
```
JamieB-gu, line 101: Byline?

JamieB-gu (approved):

src/components/editions/headerImageCaption.tsx
abeddow91: right you are, will switch that out

src/components/editions/headerImageCaption.tsx
abeddow91: Updated!

